Issue #107, opened 2014-06-17
Topics for the "comment" chapter


japaric:
- [ ] types of comment: {,in}line `//`, block `/* */`
- [ ] comments can be nested
- [ ] briefly mention that special `///` comments are used to build documentation, and link to the `rustdoc` chapter

mdinger:
This seems out of date and invalid now

PR #232, opened 2014-09-02
hello: usage go into the text from the comments.

nodakai:
This may be a matter of preference, but I think the usage of the playground codes should be in the main text because it's too important to overlook.

Also attempts to resolve the issue #107


steveklabnik:
Okay, it's been a while since we've heard from you. If you're interested in updating this pull request, just let me know!

PR #348, opened 2014-12-21
Fixes bounds/bounds.rs

### LINKED ISSUE #347: Update the Bounds chapter to support changes in std::ops
The example on chapter 32 (Bounds) fails to compile due to changes in std::ops traits (operators now take parameters by value instead of by reference, if I'm not mistaken).
The same changes make the Operator Overloading chapter description and comments misleading.

error: method `add` has an incompatible type for trait: expected struct Vec2, found &-ptr [E0053]
...
error: mismatched types: expected `T`, found `&T` (expected type parameter, found &-ptr)
error: method `sub` has an incompatible type for trait: expected struct Vec2, found &-ptr [E0053]
error: method `mul` has an incompatible type for trait: expected struct Vec2, found &-ptr [E0053]
error: type `Vec2<f32>` does not implement any method in scope named `dot`
error: binary operation `+` cannot be applied to type `Vec2<char>`
note: in expansion of format_args!
<std macros>:2:23: 2:77 note: expansion site
<std macros>:1:1: 3:2 note: in expansion of println!
error: aborting due to 6 previous errors

GBGamer:
First of all, change Add and Sub to take by value, not reference. Second,
reverts part of PR #260, because that person made the dot product into a
multiplication operation, which shouldn't have ever happened.

Also fixes #347

### DIFF
--- a/examples/bounds/bounds.rs
+++ b/examples/bounds/bounds.rs
@@ -4,11 +4,10 @@ struct Vec2<T> {
     y: T,
 }
 
-impl<
-    // Bound: `T` must implement the `Add` trait
-    T: Add<T, T>
-> Add<Vec2<T>, Vec2<T>>
-for Vec2<T> {
+// Apply bound to `T` at first instance of `T`. `T`
+// must implement the `Add` trait.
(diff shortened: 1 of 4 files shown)


brson:
Thanks!

Issue #356, opened 2014-12-25
deriving(Decodable) is deprecated in favor of deriving(RustcDecodable)


GBGamer:
We need to add an external crate, rustc_serialize, and switch over to this.

Also, deriving(Encodable) -> deriving(RustcEncodable)

steveklabnik:
I have a patch that fixes this, but it's not going to be good to go until tonight.

mdinger:
This is fixed I think...

steveklabnik:
Yes, I ended up merging it in. Thanks for the triage!

PR #357, opened 2014-12-25
Stop using #[deriving(de/encodable)]

GBGamer:
Fix-ishes #356 for the update part of it; I haven't fixed examples.


GBGamer:
@steveklabnik This needs to be merged before we can update to the latest rustc, I'm fairly certain. We need to get off of #[derive(Serialize)] and (Deserialize).

steveklabnik:
Yeah, the error is about prelude stuff, not the extra crate. I haven't had a chance to check out all this machinery.

GBGamer:
Please note: There are three files with Associated types screwery. I don't understand it at all, so it's currently whitelisted. If someone does have an idea, pull request or comment.

steveklabnik:
This is outdated, and I'm not sure I want to deal with the infrastructure changes right now. Sorry it took me so long to look at this :/

GBGamer:
It no longer uses cargo. Would you mind looking at this again?

mdinger:
You have to deal with the infrastructure change...rbe uses the regex crate so you have to get it externally or rewrite the functionallity internally. rbe won't build without it. What's the alternative?

GBGamer:
@mdinger `Regex::new()`. `regex` is still an `extern crate` that you don't have to use Cargo for, unlike `regex_macro`.

mdinger:
Ah. My mistake. I knew about that but didn't think of it.

GBGamer:
I initially attempted to implement regex_macros; not the easiest thing... :D

mdinger:
I agree. Not easy.

mdinger:
Also, rbe is structured in json. Doesn't it need serialize for it? Is there a workaround?

GBGamer:
@mdinger How do you mean? We have serialize with `mod rustc_serialize`.

mdinger:
Doesn't work for me. rust can't find serialize for me. Serialize has been moved to crates.io. See commit note below:

```plain
commit b04bc5cc49a398df712092a68ab9ad83019498ad
Merge: 9b99436 a76a802
Author: Alex Crichton <alex@alexcrichton.com>
Date:   Mon Dec 22 12:46:31 2014 -0800

    rollup merge of #20033: alexcrichton/deprecate-serialise
    
    This commit completes the deprecation story for the in-tree serialization
    library. The compiler will now emit a warning whenever it encounters
    `deriving(Encodable)` or `deriving(Decodable)`, and the library itself is now
    marked `#[unstable]` for when feature staging is enabled.
    
    All users of serialization can migrate to the `rustc-serialize` crate on
    crates.io which provides the exact same interface as the libserialize library
    in-tree. The new deriving modes are named `RustcEncodable` and `RustcDecodable`
    and require `extern crate "rustc-serialize" as rustc_serialize` at the crate
    root in order to expand correctly.
    
    To migrate all crates, add the following to your `Cargo.toml`:
    
        [dependencies]
        rustc-serialize = "0.1.1"
    
    And then add the following to your crate root:
    
        extern crate "rustc-serialize" as rustc_serialize;
    
    Finally, rename `Encodable` and `Decodable` deriving modes to `RustcEncodable`
    and `RustcDecodable`.
    
    [breaking-change]
```

GBGamer:
@mdinger that's why I have added the `rustc_serialize` directory :)

mdinger:
Did you forget to push?

mdinger:
This is still using cargo

GBGamer:
@mdinger it still says it's using Cargo in the PR because it doesn't update.
[Here][rbe] is where it actually is.

[rbe]: github.com/GBGamer/rust-by-example

mdinger:
Hmm...that seems the worse of two options.

@steveklabnik You really want to internalize rustc_serialize?

GBGamer:
I agree, I don't like this option, but I only see the two, and it seems like we
don't want to have a dependency on Cargo.

GBGamer:
I guess we also have a third option, which is relying on
`extern crate serialize`, but that'll only last for a little while longer 
before we have to decide the lesser of two evils.

mdinger:
Well, all json examples have been excised already so that's probably not the option he's considering.

GBGamer:
@mdinger unlikely :D

mdinger:
If you have to download it, you could do it like [here](https://github.com/rust-lang/rust-by-example/blob/master/setup-stage.sh#L8). That would be something like this:
```cmd
serialize_repository=https://github.com/rust-lang/rustc-serialize/trunk/src'
serialize_local_folder='src/rustc-serialize'

svn checkout ${serialize_repository} ${serialize_local_folder}
```

Then it will stay up to date everytime `setup-stage.sh` was run.

GBGamer:
@mdinger I have now tried it that way, and it actually works pretty well. Definitely check out my solution.

steveklabnik:
I have a local build from last night that Cargo-ifies our build process,
but it needs the new nightlies to update, and all of the upstream
dependencies I had to fix in order to work.

PR #361, opened 2014-12-29
First attempt at switching to RustcEncodable and RustcDecodable

GBGamer:
I need help on making this actually work; it `make && make test`s well;
however, after serving, the module for rustc_serialize doesn't exist. I do not
know how to get around it, or even if we want to; we need to do something,
though, because `Decodable` and `Encodable` are now deprecated, and
rustc-serialize is an external crate.


steveklabnik:
Yeah, the external crate part here is bad. I think I'm going to go with removing this for now, as it is external.

steveklabnik:
Thank you though!

PR #363, opened 2015-01-02
attribute: fixing Markdown link

Quackmatic:
The Markdown link is named `[crate]` but the link referred to it as `[config]`, causing a formatting error. This fixes the link reference. This is essentially a typo fix so I'm just doing a PR directly.


Quackmatic:
The failed build is unrelated to this commit.

GBGamer:
Works correctly on my machine (although we need to update `update`)

brson:
Thanks!

PR #336, opened 2014-12-18
Manually grab `ace` to use latest gitbook

mdinger:
This makes gitbook 1.5.0 (what NPM installed) work for me. I had an syntax highlighting issues with 1.4.2 which is what is installed by Ubuntu but I think it's related to `gitbook serve` not working properly. Otherwise, it seems to work fine.

I wasn't sure if you'd want *ace* in tree so I downloaded it in the script.

PR #369, opened 2015-01-07
make `make clean` clean gitbook

GBGamer:


brson:
On first glance it seems a bit odd for clean to delete gitbook to me, because it's not something that we control. Is installing gitbook also part of the build process?

mdinger:
@brson 
In [Makefile#L32](https://github.com/rust-lang/rust-by-example/blob/master/Makefile#L32), `make book` calls `npm` to install gitbook to `node_modules/gitbook`. You manually delete them to try different gitbook versions.

Previously, ace only worked with gitbook version 0.7.1 because they bundled ace with gitbook then. Rustbyexample has been installing gitbook 0.7.1 locally since then because of it, ignoring any local distribution copy. Even at https://github.com/rust-lang/rust-by-example/pull/336 it's still using npm to install locally because the distribution copy I was using didn't work (I think `make serve` was broke somehow but was fixed in a newer version).

```Makefile
book: node_modules/gitbook
	...
node_modules/gitbook:
	npm install gitbook@0.7.1

**Comment by GBGamer** [comment]
So does that mean that the latest gitbook works?

**Comment by GBGamer** [comment]
Sorry, I think we should still do this. A `make clean` should make the directory as close as possible to the default source tree.

**Comment by mdinger** [comment]
Yeah, latest gitbook worked for me. I'd still install directly from npm though because then you aren't distribution bound. @brson already merged it.

**Comment by mdinger** [comment]
I looked into this and it's probably not the best way to do this. While `npm install gitbook` is considered local, it does place files in `~/.npm/` and I reached a situation where it (`npm`) by trying to modify those files wanted `sudo` priviledges, though I'm not sure why. There is a corresponding command [local uninstall](https://docs.npmjs.com/getting-started/uninstalling-local-packages) command: `npm uninstall gitbook` which is probably the better way.

More like this I suppose:
```Makefile
clean:
	npm uninstall gitbook
	rm -rf bin stage
	rm -rf node_modules/gitbook
	rm -rf node_modules/.bin
```

PR #370, opened 2015-01-07
Allow '$' in rust files

### LINKED ISSUE #334: Escaped Characters on fs.html
There are some characters that have been escaped in the [fs.html example](http://rustbyexample.com/fs.html). 

``` rust
use std::io::fs;
use std::io::fs::PathExtensions;
use std::io::{File, IoResult, USER_RWX};

// A simple implementation of `cat path&#96;
fn cat(path: &Path) -&gt; IoResult&lt;String&gt; {
    File::open(path).and&#95;then(|mut f| f.read&#95;to&#95;string())
}
// A simple implementation of &#96; echo s > path`
```

GBGamer:
Actually fixes #334, without workarounds.

### DIFF
--- a/src/playpen.rs
+++ b/src/playpen.rs
@@ -12,6 +12,7 @@ fn escape(source: &str) -> String {
 
     for chr in source.trim().chars() {
         match chr {
+            '$' => s.push_str("&#36;"),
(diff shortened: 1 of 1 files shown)


brson:
Danke!

PR #372, opened 2015-01-08
Make deploy.sh script executable

frewsxcv:


steveklabnik:
Is there a specific reason to do this?

frewsxcv:
All their other shell scripts that are in root are executable. Doesn't make a difference to me whether this gets merged or not, I just wanted to point out the inconsistency

steveklabnik:
Ahh, that's just a difference between me and the previous maintainer, I guess. Might as well make it consistent.

steveklabnik:
Thanks!

PR #375, opened 2015-01-08
Upgrade to HEAD Rust.

steveklabnik:
It's unfortunate to lose all of the JSON stuff, but we can't use external
crates with the current build system.


steveklabnik:
:confetti_ball:

PR #376, opened 2015-01-09
array: Update documentation to new fixed-length array syntax.

stevencrockett:
The code for the example was updated previously, but the accompanying explanation was not.


steveklabnik:
Thank you!

PR #377, opened 2015-01-09
Port src/update to cargo

steveklabnik:
And update the build to `rustc 1.0.0-dev (6354d60ed 2015-01-08 20:59:48 +0000)`

/cc @brson

### DIFF
--- a/src/markdown.rs
+++ b/src/markdown.rs
@@ -1,6 +1,7 @@
 use file;
 use playpen;
 use std::iter::repeat;
+use regex::Regex;
 
 pub struct Markdown<'a, 'b> {
     content: String,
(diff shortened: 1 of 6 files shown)


src/markdown.rs
```diff
@@ -117,7 +118,7 @@ impl<'a, 'b> Markdown<'a, 'b> {
     fn insert_playpen_links(&mut self) -> Result<(), String> {
         let id = self.id;
         let prefix = self.prefix;
-        let re = regex!(r"\{(.*)\.play\}");
+        let re = Regex::new(r"\{(.*)\.play\}").unwrap();
```
mdinger: If you do this you don't need the macros right?

steveklabnik: I don't need `regex_macros`, it's true

PR #381, opened 2015-01-10
Update to latest rustc (int -> i32)

### LINKED ISSUE #380: Several examples give warnings about deprecated 'i' and 'u' suffix
The following pages need to have updated suffixes:

* http://rustbyexample.com/print.html
* http://rustbyexample.com/literals.html
* http://rustbyexample.com/variables.html
* http://rustbyexample.com/variables/mut.html
* http://rustbyexample.com/variables/scope.html
* http://rustbyexample.com/variables/declare.html
* http://rustbyexample.com/type.html
* http://rustbyexample.com/type/literals.html

And possibly more. Would fix myself but don't know what the desired fix is. (`is` and `us`?)

The last page I checked was http://rustbyexample.com/type/inference.html, which is broken but not because of suffixes. Starting to think maybe there's an issue with the Javascript Rust... thing.

Additionally, http://rustbyexample.com/type/literals.html should be updated to explain the new suffixes.

UPDATE YET AGAIN: actually pretty much all the examples I've checked so far are broken. And some of them for other reasons (e.g. http://rustbyexample.com/variables.html breaks because apparently `()` doesn't play well with `println!`..?) If this is already known sorry for creating more noise in your issue tracker!!

GBGamer:
Also add #[feature(box_syntax)] for places where it is necessary.

Rename `-A experimental` to `-A unstable`

Change `{}` to `{:?}` where necessary.

Fix #380


GBGamer:
The failure is due to `make`; this PR is specifically for `make test`

steveklabnik:
This PR has conflicts, can you rebase please

GBGamer:
@steveklabnik It's is unconflicted now :)

steveklabnik:
Thanks!

PR #382, opened 2015-01-10
Change int and uint to new types

### LINKED ISSUE #380: Several examples give warnings about deprecated 'i' and 'u' suffix

gingermusketeer:
Makes a few smaller changes so that `make test` passes

Closes #380

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 GITBOOK = node_modules/.bin/gitbook
 RUSTC = rustc
 STRICT = -D deprecated
-QUIET = -A unused-variables -A dead-code -A unused-assignments -A experimental
+QUIET = -A unused-variables -A dead-code -A unused-assignments -A unstable
 RUSTC_NT = $(RUSTC) -Z no-trans --test $(QUIET) ${STRICT}
 WHITELIST = examples/attribute/cfg/custom/custom.rs \
 						examples/borrow/borrow.rs \
--- a/examples/array/array.rs
+++ b/examples/array/array.rs
@@ -1,17 +1,17 @@
 use std::mem;
 
 // This function borrows a slice
-fn analyze_slice(slice: &[int]) {
+fn analyze_slice(slice: &[isize]) {
     println!("first element of the slice: {}", slice[0]);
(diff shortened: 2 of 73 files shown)


GBGamer:
The issue with this PR is that we don't want `isize` as the default (hence the renaming). We want, for most things, to use i32/u32, and, in cases where it might be necessary to use 64 bits, i64/u64. We only want to use isize in places where we create vectors, and such like that.

gingermusketeer:
@GBGamer I didn't see it mentioned that the renaming was to discourage its usage. My understanding is that it was changed to better reflect the consequences of using it.

I am happy to change it though. I do think that `i32`/`u32` etc would be a better choice

steveklabnik:
We want:

1. To not use suffixes where they're not needed.
2. To prefer i32/u32 for most things
3. to use isize/usize when dealing with lengths.

gingermusketeer:
@steveklabnik Just spotted @GBGamer's PR with the same fixes. Happy for you to close this and use #381 instead

steveklabnik:
Okay! That's what I've done. Thank you!

PR #385, opened 2015-01-11
Fix #[derive(Copy)] on borrow/mut/mut.rs

### LINKED ISSUE #384: borrow/mut: calls move a copy
The example:

```rust
    // `mutable_geb` is a mutable copy of `geb`
    let mut mutable_geb = geb;

    // Borrow a mutable object as mutable
    new_edition(&mut mutable_geb);

    // Mutable objects can be immutably borrowed
    borrow_book(&mutable_geb);
```

says "`mutable_geb` is a mutable *copy* of `geb`". To me *copy* means creation of another independent object (a `memcpy()` of the data), but this doesn't behave like a copy. If I add at the end:

```rust
    borrow_book(&geb);
```

I get:

> error: use of moved value: `geb`
> error: aborting due to previous error

Which indicates that it wasn't a copy — there's still only one shared object.

GBGamer:
Fixes #384

### DIFF
--- a/examples/borrow/mut/mut.rs
+++ b/examples/borrow/mut/mut.rs
@@ -1,4 +1,5 @@
 #[allow(dead_code)]
+#[derive(Copy)]
 struct Book {
     // `&'static str` is a reference to a string allocated in read only memory
     author: &'static str,


steveklabnik:
Thanks.

PR #388, opened 2015-01-11
Add missing attribute example

gingermusketeer:


steveklabnik:
I'm sorry, what is this fixing? I think I've missed something.

gingermusketeer:
`custom.out` did not exist so it ended up looking like

screen shot 2015-01-12 at 1 14 15 pm

steveklabnik:
Gotcha. The right way to make this change would be to create a file for `custom.out`. This system is used so that it gets updated automatically... would you like to take a crack at that?

gingermusketeer:
Ah right. I changed it to match the section below "With the custom `cfg` flag:". Should i change that to be separate file as well?

gingermusketeer:
@steveklabnik I got to the bottom of the issue. It was a wider problem with all `{example.out}` replacements. When the show formatter is used on a string it surrounds it in quotes. This meant it was trying to replace `{"example".out}` instead of `{example.out}`

steveklabnik:
Ahh awesome. Thanks!

PR #395, opened 2015-01-11
Fix generics/phantom type comments

### LINKED ISSUE #392: phantom/units: The example is very dense
```rust
impl<Unit, T: Add<T, Output=T> + Copy> Add<Length<Unit, T>>
        for Length<Unit, T> {
    type Output = Length<Unit, T>;
```

There's *a lot* going on in that line, and it requires significantly more effort to deconstruct than previous examples.

Could it be broken down? Maybe the large comment that is above that line could be mixed with the code like this:

```rust
impl<Unit, T: Add<T, Output=T> + Copy> // this is impl<T> blah blah
Add<Length<Unit, T>> // of this and that
for Length<Unit, T> // for this yada yada
```

Also it's not explicitly mentioned what `Output=T` is for.

GBGamer:
Fixes #392, hopefully.

### DIFF
--- a/examples/generics/phantom/units/units.rs
+++ b/examples/generics/phantom/units/units.rs
@@ -1,24 +1,34 @@
 use std::ops::Add;
 
-// Null enumerations to define unit types
+/// Null enumerations to define unit types
 #[derive(Show, Copy)]
-enum Inch {}
+struct Inch;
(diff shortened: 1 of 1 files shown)


steveklabnik:
Thanks!

PR #396, opened 2015-01-11
Update threading example

sinistersnare:
Now the example uses range notation (`..`) and `i32` instead of `int`


sinistersnare:
I should have looked at the other PRs when i did this :) ill close this in favor of one of the other.

PR #407, opened 2015-01-13
Make macros work inside playpen

mdinger:
This makes the macros use playpen instead of just being text. I just named the titles whatever the `*.rs` file was at the time. It can be changed later if need be. I tried to make each patch minimal so hopefully it'll be pretty straightforward to review.

All the examples work also.


GBGamer:
+1 from me, this seems like a good change.

steveklabnik:
Thank you!

PR #408, opened 2015-01-13
Updating borrow/borrow: adding box_syntax feature, changing int to isize

jamiltron:


steveklabnik:
I like this patch, but two things:

1. We shouldn't use `box_syntax`, we should stick to no feature gates where possible.
2. This should be `i32`, not `isize`.

Would you mind making those changes? And thank you!

jamiltron:
Will do!

jamiltron:
Updated. Should the other box_syntax features be removed? It appears that the only place where its really useful is in the Box, stack and heap example.

steveklabnik:
Thank you!

PR #409, opened 2015-01-13
Stop attempting to be clever in iter

### LINKED ISSUE #398: iter: uses clever code without a comment
The example uses `mem::replace` without explanation why. 

```rust
Some(mem::replace(&mut self.curr, new_curr))
```

Assigning to `self.curr` instead seems to work, so is it just to make the code a line shorter? Or is there another reason, e.g iterators have to be thread-safe and `mem::replace` is doing atomic swap?

GBGamer:
fixes #398


alexcrichton:
Thanks!

steveklabnik:
:confetti_ball:

PR #412, opened 2015-01-14
Update print.rs

naveensrinivasan:
Fixed the warning "`i` suffix on integers is deprecated; use `is` or one of the fixed-sized suffixes" by using `is` suffix.


steveklabnik:
Thank you! Did you try removing the suffix entirely? That'd be even better.

naveensrinivasan:
I was thinking about it but didn't know if example was trying convey that. I have updated it. Thanks

steveklabnik:
Cool thanks!

PR #414, opened 2015-01-15
generics/impl: clarify that generics don't depend on a null struct

aidanhs:


steveklabnik:
Good call! Thank you!

PR #415, opened 2015-01-15
Add margin to bottom of interactive code blocks

iKevinY:
On pages like [this one](http://rustbyexample.com/type.html) where interactive code blocks are followed by more content, there is no spacing between the bottom of the code block and the top of the following paragraph. By adding a 16px margin to the bottom of `#active-code`, the margin below the code block matches the margin above.


steveklabnik:
Thanks!

PR #417, opened 2015-01-16
Rename int/uint to isize/usize in type/input.md

GBGamer:
Also, isize and usize are technically the size of a pointer, not a machine
word


steveklabnik:
:+1:

PR #418, opened 2015-01-16
Remove use of the term "prelude"

### LINKED ISSUE #404: path: prelude?
> The prelude exports the appropriate platform-specific Path variant.

Sounds like compiler jargon. Does it mean "global `Path`"?

pornel:
Fixes #404


steveklabnik:
Hey thanks! This is strictly less information than before, so I don't want to fix that issue this way.

PR #420, opened 2015-01-18
Remove unused use

mdinger:
[Iter](http://rustbyexample.com/iter.html) is throwing an `unused import` warning for `std::mem` because https://github.com/rust-lang/rust-by-example/pull/409 removed `mem::replace()` usage. This fixes that warning.


steveklabnik:
Thanks!

Issue #390, opened 2015-01-11
borrow/ref: Not clear what's the difference between ref and &


pornel:
In other languages `&` is called a reference and borrowing looks very much like references, so to me `&` and `ref` seem like the same thing.

I'd appreciate an example that contrasts the two — why `let Point { x: ref ref_to_x, y: _ }` is OK, but `let Point { x: & ref_to_x, y: _ }` isn't?

GBGamer:
`ref` is used in pattern matching, so its used when you want to borrow
something in, for example, a match statement.

    let a = Some("Hello!".to_string());
    match a {
        Some(ref s) => { //s is an &String here
        }
        ...
    }

    match a {
        Some(s) => {
            // s is a String here, and therefore
            // is owned by the match
        }
        ...
    }

pornel:
Why can't you use `&` to borrow in the match?

```rust
match a {
Some(&s) => {}
}
```
?

GBGamer:
That, I don't know. It's not really a decision I agree with. I like that
style better, personally.

On Sun, Jan 11, 2015 at 11:27 AM, Kornel <notifications@github.com> wrote:

> Why can't you use & to borrow in the match?
>
> match a {Some(&s) => {}
> }
>
> ?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/rust-lang/rust-by-example/issues/390#issuecomment-69507268>
> .
>



-- 
"Calling all. This is our last cry before our eternal silence."
"The lone and level sands stretch far and away"

pornel:
I'm not trying to question Rust design decision here, just understand difference between `ref` and `&` — and I'm reporting that the example about references doesn't elaborate on that.

Whether that's just arbitrary, or quirky syntax to avoid ambiguity, or it has slightly different meaning, I'd like to know that, because I'm not sure when to use which and why rustc complains about `&_-ptr` or such when I use `&` instead of `ref`.

GBGamer:
I think the idea is that you destruct as you construct, like in a

   let op = Some(a);

    match op {
        Some(x) => x,
        None => ...
    }

So if you type

    match op {
        Some(&x) => // here you are dereferencing a pointer
            ...
        None => ...
    }

pornel:
Sorry, I don't understand your last answer. `&x` is dereferencing a pointer? I thought `*x` is dereferencing.

Perhaps I'm asking stupid question, but the mental model I have is from C:

[...]

PR #421, opened 2015-01-18
Move all flow control items into `flow control` group. Add `if let` explanations.

mdinger:
This isn't done but I'm working on moving all flow control structures together into the same folder and adding `if let` and `while let` which I don't think are explained anywhere except for their RFC.

Previously the flow controllers have been together mostly except for `match` which seems to be alone so some data structures can be introduced. Currently, I think it would be best to move the flow control together, explain them, and link to necessary external sources as necessary. This is not ideal but not bad either. I don't think theres a good reason the controllers should be separated.

In the future, it may be a good idea for the datatypes to all be moved to the beginning before the flow control. They shouldn't be as problematic to explain (except for `String`/`&str`).

I'm not sure moving the controllers around will be well received so I'm trying to explain a little beforehand. I'll try to get `while let` explanations up maybe tomorrow...we'll see.

I didn't spend a ton of time on the `if let` example. Mainly tried to get it to work. It may need tweaking.

cc @GBGamer You may be interested in this. Comments welcome, particularly with `if let`.


examples/flow_control/if-else/input.md
```diff
@@ -4,3 +4,6 @@ block. If-else conditionals are expressions too; and, because of Rust type
 safety, all the branches must return the same type.
 
 {if-else.play}
+
+###See also:
```
steveklabnik: I think there should be a space between the `#` and the `S`

examples/flow_control/match/guard/guard.rs
```diff
@@ -3,13 +3,10 @@ fn main() {
     // TODO ^ Try different values for `pair`
 
     println!("Tell me about {:?}", pair);
-    // Match can be used to destructure a tuple
     match pair {
-        // Destructure the tuple
         (x, y) if x == y => println!("These are twins"),
         // The ^ `if condition` part is a guard
         (x, y) if x + y == 0 => println!("Antimatter, kaboom!"),
-        // `_` means don't bind the value to a variable
```
steveklabnik: Why remove this stuff?

mdinger: I split this into 2 sections because the way it was previously named "Destructuring and guards". The problem was when in a deeply nested folder (`Flow_control -> Pattern matching -> Destructuring and guards`), `Destructuring and guards` didn't fit in the sidebar. So I split it into 2 sections: `Destructuring` and `Guards`.

`Destructuring` (directly before `Guards`) covers those items so I dropped them. I'm not sure I split it in the most ideal way but I'll look at it again later. Opinions welcome.

examples/flow_control/match/destructure/input.md
```diff
@@ -0,0 +1,6 @@
+The arms of a `match` block can be destructured.
+
+{destructure.play}
+
+###See also:
+[Tuples](/tuples.html)
```
mdinger: I didn't know links could use the local (`/tuples.html`) syntax until recently. I thought they needed to be absolute. This is nice.

examples/flow_control/if-else/input.md
mdinger: Fixed all these.

mdinger:
Added `while let` and reworked the `if let` example a bit. I think it's better. I've gotta let it sit for a bit so I can try and come look the whole thing over fresh later.

examples/flow_control/if_let/input.md
```diff
@@ -0,0 +1,30 @@
...
+`if let` is cleaner for this use case and in addition allows various
+failure options to be specified:
+
+{if_let.play}
+
+### See also:
+[`enum`][enum], [`Option`][option], and the [RFC][if_let_rfc]
+
+[enum]: /enum.html
+[if_let_rfc]: https://github.com/rust-lang/rfcs/pull/160
```
mdinger: Added RFC links for  `if let` and `while let`. May drop it. Kinda useful for more info if someones interested. Uncertain.

examples/flow_control/while_let/input.md
```diff
@@ -0,0 +1,39 @@
...
+            // ^ Requires 3 indentations!
+        },
+        // Quit when the destructure fails, meaning `break`.
+        _ => { break; }
+        // ^ Why should this be required? Seems superfluous.
+    }
+}
+```
+
+`while let` makes this sequence much nicer:
```
mdinger: Note for me to remember: `while let` does *not* have the corresponding `else`/`else if` extensions that `if let` has and I probably should comment about it.

mdinger: Fixed

mdinger:
Added more examples covering `ref`, pointer, enum destructuring, and `match` binding. Gotta look over it again but it seems much more comprehensive than it was before.

@steveklabnik I don't know if you're gonna want to review this in one patch...or what. Definitely easier to peruse if you make the examples locally though.

examples/flow_control/match/destructuring/destructure_enum/input.md
```diff
@@ -0,0 +1,10 @@
+An `enum` is destructured similarly:
```
mdinger: Filenames are really long because some of the `id`s (like `struct`) that were being used in JSON were clashing with other locations (like `struct`). This fixed it and will help with future clashing but it's not really pretty.

mdinger:
r? @steveklabnik Made a few more minor changes. I definitely think this is an improvement over before (that is before the PR).

Trying to add cross referencing to important concepts as I go. Avoided linking to `println!` because that would add links to every page. Though, I don't mind doing so (later PR though if desired).

mdinger:
Made changes to elaborate on destructuring for #390

steveklabnik:
I've looked at this patch once or twice already and am just not 100% sure. I'm trying to get to rust by example later this wekk.

mdinger:
Okay. Just let me know what some of the reservations are when you get to it. Theoretically, it's possible they could be addressed but I'm also hesitant to make the patch bigger because review time gets bigger. Easier to fix later.

GBGamer:
@mdinger If you need any help with the patch still, tell me. Sorry I've been out of contact, I just got a new guitar and I've been playing it... a lot.

mdinger:
@GBGamer Thanks. I'm not anticipating many issues here except that in moving all the flow controls together, some things are now introduced too early. Pointers, structs, enums, and options are now used before they are introduced. So I tried to be liberal with cross referencing.

I don't think that something that can be avoided long term though. When looking at a language for the first time, people want to know what the structures, types, flow controls, and operators look like and don't want to have to hunt around for them. Not being explained (`option` for example) at first sight isn't a huge problem as long as you point to where they need to look though.

This depends on the review though. @steveklabnik may reject it without shuffling of structures also. Then I'd need an additional patch for that. You could help with that. But it entirely depends on what he says. He may also have additional concerns I haven't thought of.

My preference would be to merge and have the rest (possibly) restructured later. But I gotta wait I guess.

mdinger:
@steveklabnik This needs a rebase and updating but I'm hoping you'll give comments for and against this before I do that. I doubt I can update this until late next week.

steveklabnik:
@mdinger well Monday -> Weds I am flying from Europe to Australia, soooooo

mdinger:
Okay. I guess it can't be helped.

steveklabnik:
Okay! So. I think that this is good to merge, other than needing a rebase. @mdinger , please rebase it, i'll merge, and we can improve from there. Sorry about the continued delays :/

mdinger:
Okay cool!!

mdinger:
This threw errors with the new link checker but the links do work. I didn't track it down. It might be that the new folders on rustbyexample.com don't exist yet or that the link checker doesn't deal with a root directory properly (`/new_example`). I'm not sure.

mdinger:
@dbbrian See previous comment.

examples/flow_control/match/destructuring/destructure_structures/input.md
```diff
@@ -0,0 +1,6 @@
+Similarly, a `struct` can be destructured as shown:
+
+{struct.play}
+
+### See also:
+[Structs](/structs.html)
```
mdinger: The link checker is failing on links like this though they do work. They will only work when it's on rustbyexample.com or `make serve` is currently running though.

dbbrian:
@mdinger Will fix ASAP.

steveklabnik:
Awesome, glad to finally land this. thanks for your patience, @mdinger , and thanks for the quick link checker fix @dbbrian !

mdinger:
I'm soooooooo glad it finally landed. I was beginning to think the only non-trivial things that will ever land are the ones you write. Makes it hard to want to contribute ...

I really tried not to nag but seems like I had too...don't like doing it. I don't know. The docs still need so much improvement. Best to keep working to make them better.

Really glad you finally merged!!!

steveklabnik:
Such is open source. Sometimes you have to nag. It happens.

PR #422, opened 2015-01-18
`int` -> `isize`, `uint` -> `usize`

mdinger:


mdinger:
The type (`type u64_t`) change doesn't necessarily matter but I figure it's reminiscent of when `int` was actually a thing and that should be avoided.

steveklabnik:
Thanks!

PR #424, opened 2015-01-20
Remove json chapters because the examples are missing

mdinger:
The JSON examples were deleted. Remove the chapter titles so they don't show up on the webpage.


steveklabnik:
Ah thank you!

PR #430, opened 2015-01-24
Updated print.rs to remove the "i" suffix.

naveensrinivasan:
The "i" suffix has been replaced by "is" and is not required now for this change. This removes the warning.


steveklabnik:
Thank you!

PR #440, opened 2015-01-30
The fizzbuzz_to function is corrected now to use Rust's new a..b range syntax

michaelmcgirr:
It seems like there was a small typo in the fn.rs example.  It was using the old range(a, b) instead of the a..b syntax for range.

### DIFF
--- a/examples/fn/fn.rs
+++ b/examples/fn/fn.rs
@@ -31,7 +31,7 @@ fn fizzbuzz(n: u32) -> () {
 // When a function returns `()`, the return type can be omitted from the
 // signature
 fn fizzbuzz_to(n: u32) {
-    for n in range(1, n + 1) {
+    for n in 1..(n + 1) {
         fizzbuzz(n);
     }
 }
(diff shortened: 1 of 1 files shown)


examples/fn/fn.rs
```diff
@@ -31,7 +31,7 @@ fn fizzbuzz(n: u32) -> () {
 // When a function returns `()`, the return type can be omitted from the
 // signature
 fn fizzbuzz_to(n: u32) {
-    for n in range(1, n + 1) {
+    for n in 1..(n + 1) {
```
mdinger: Shouldn't need parenthesis: `1..n + 1`.
http://is.gd/GjEamj

michaelmcgirr: You're right @mdinger - I pushed some code now to remove the parenthesis.

mdinger:
@steveklabnik Can you merge this? Or ask for squash and merge. It's trivial.

steveklabnik:
Yeah sorry.

mdinger:
Thanks

PR #444, opened 2015-02-04
lifetime/borrow: fix word order

ijks:
'without knowing what is a lifetime' --> 'without knowing what a lifetime is'


steveklabnik:
Thank you! This is much better.

PR #445, opened 2015-02-06
Update rust

mdinger:
* `rand` is no longer in rustc so Cargo is needed to support it so I removed it.
* `std::os::args` got moved to `std::env::args` and from a `Vec` to an `iter`. I played around with some other options (besides `collect`ing into a vector) like `for` loops and such but they drastically altered the examples. It's easier to just make it work again for now. Someone will have to give the examples attention again in the future.


mdinger:
`Show` to `Debug` was automatic and I didn't scan the trillion warnings for new ones. All errors should be good.

mdinger:
Sorry, force pushed. Wanted `rand` to have it's own commit so it can be reverted easily in the future.

steveklabnik:
Looks good, thank you!

PR #452, opened 2015-02-14
Update to head Rust

steveklabnik:


examples/sockets/client.rs
```diff
@@ -6,7 +6,7 @@ mod common;
 
 fn main() {
     // `args` returns the arguments passed to the program
-    let args: Vec<String> = env::args().map(|x| x.into_string().unwrap())
+    let args: Vec<String> = env::args().map(|x| x.to_string())
                                        .collect();
```
mdinger: For whatever reason, these can now be written as below. I'm not exactly sure why but it works now. Pretty nice to not need map anymore...
`let args: Vec<String> = env::args().collect();`

PR #460, opened 2015-02-16
Remove redundant into_iter()

voidxnull:
Is this an appropriate fix?


steveklabnik:
It's certainly not needed here, yes. Thanks!

PR #461, opened 2015-02-19
Fix the build

steveklabnik:


steveklabnik:
Travis is being stupid, but this does merge now.

Issue #465, opened 2015-02-21
Incorrect search engine summary of rustbyexample.com


sjef:
Searching for "rust by example" in Google returns this "**[Rust][rust]** is a programming language with a focus..." as rustbyexample.com's summary(or whatever it is called). This issue is also true for Baidu.

Bing and Duckduckgo returns this "**Rust by Example. Rust** is a programming language with a focus on type..."

Below are images showing the search result in Google, Baidu, Bing and Duckduckgo.

Google:
![rbegoogle](https://cloud.githubusercontent.com/assets/8290704/6316077/3e8ed428-ba18-11e4-87fe-49e32eafb07a.PNG)

Baidu:
![rbebaidu](https://cloud.githubusercontent.com/assets/8290704/6316079/485e5546-ba18-11e4-9518-6e01e951b5a5.PNG)

Bing:
![rbebing](https://cloud.githubusercontent.com/assets/8290704/6316078/432fd4aa-ba18-11e4-8ef5-4dbf5760dc8d.PNG)

Duckduckgo:
![rbeduckduckgo](https://cloud.githubusercontent.com/assets/8290704/6316076/3d5faf82-ba18-11e4-8703-ce9bf9ad0ade.PNG)

mdinger:
A possible fix is sorta detailed at https://github.com/rust-lang/rust-by-example/pull/466#issuecomment-75396099 but it's very preliminary.

sjef:
No longer appears to be an issue with any of the search engines.
Probably should have written this comment before I closed it.

PR #466, opened 2015-02-21
Introduction: remove duplicate [rust] call

sjef:
Possible fix for issue #465

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,6 +1,6 @@
 # Rust by Example
 
-[Rust][rust] is a programming language with a focus on
+[Rust] is a programming language with a focus on
 type safety, memory safety, concurrency and performance.
 
 This book is a collection of bitesize examples that illustrate various rustic


mdinger:
That won't work. `[a][b]` is deferring a link to later in the file. Like here, the link would use the text `Rust` and use the `rust` link at the bottom of the page. This just means that google isn't parsing the markdown. This markdown:

```markdown
[Rust][rust] is a programming language with a focus on

[rust]: http://www.rust-lang.org/
```

Results in this:

[Rust][rust] is a programming language with a focus on

[rust]: http://www.rust-lang.org/

mdinger:
Markdown referencing conventions are listed [here](https://github.com/steveklabnik/rfcs/blob/api_comment_conventions/text/0000-api-comment-conventions.md#using-markdown)

sjef:
I see. 

The issue is then completely in the hands of the search engine provider?

mdinger:
Yeah. That's why Bing and Duckduckgo look correct; they are doing more comprehensive formatting of search results. Google has typically been pretty minimalistic about search results and may be part of it's popularity.

A more correct approach would be to request that Google parse the markdown in the search results though I don't know if Google has that option or where it would be. I just looked and couldn't find it.

[Yahoo](http://search.yahoo.com/search?p=rust%20by%20example) also get's it correct but Yahoo has typically been pretty heavy-handed with search styles so that's not surprising.

mdinger:
If Google has an option though or something to configure search results, it would be good to fix this. This isn't particularly pretty.

sjef:
I don't know how to get in touch with Google regarding issues like this. 

One fix from our side could be to refer/link to rust-lang.org at the second mention of "Rust". This is of course suboptimal.

mdinger:
I think it's probably buried somewhere in [here](https://support.google.com/webmasters/topic/4598337?hl=en&ref_topic=3309300) though I don't have time to find it now. You've probably got to annotate the webpage so it's labeled as Markdown somehow.

I doubt they'll go for changing the formatting. I'd explore annotation. It's probably possible.

steveklabnik:
Since this isn't going to be an actual fix, I'm going to give it a close.

PR #472, opened 2015-02-24
clone: Use newer Box syntax and Debug trait for printing

badboy:
This runs for me on latest rustc.

### DIFF
--- a/examples/clone/clone.rs
+++ b/examples/clone/clone.rs
@@ -1,5 +1,3 @@
-#![feature(box_syntax)]
-
 // A unit struct without resources
 #[derive(Debug, Copy)]
 struct Nil;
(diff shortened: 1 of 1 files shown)


steveklabnik:
Thank you! We should be able to remove the feature gate on line 1 then, too. Mind doing that?

badboy:
aaand done.

examples/clone/clone.rs
```diff
@@ -27,7 +25,7 @@ fn main() {
     println!("copy: {:?}", moved_pair);
 
     // Error! `pair` has lost it resources
-    //println!("original: {}", pair);
+    println!("original: {:?}", pair);
```
steveklabnik: when you lost the `//`, it broke the tests!

badboy: D'oh!

steveklabnik:
Thanks so much!

PR #477, opened 2015-02-24
Makeover

steveklabnik:
I've never actually gone over Rust By Example. I'm doing to do it now, starting here, and just edit as I go.


steveklabnik:
Merging what we've got for now, will pick up more later.

examples/print/print.rs
```diff
@@ -1,25 +1,23 @@
 fn main() {
-    // `print!` is like `println!` but it doesn't add a newline at the end
-    print!("January has ");
-
-    // `{}` are placeholders for arguments that will be stringified
+    // `{}` are placeholders for arguments that will be stringified.
```
mdinger: Probably should be `{:?}` in general because otherwise things get more complicated.

steveklabnik: Sure thing, feel free to open a PR :)

examples/structure.json
```diff
@@ -2,7 +2,7 @@
   { "id": "hello", "title": "Hello World", "children": null },
   { "id": "print", "title": "Formatted print", "children": null },
   { "id": "literals", "title": "Literals and operators", "children": null },
-  { "id": "variables", "title": "Variables", "children": [
+  { "id": "variable_bindings", "title": "Variable Bindings", "children": [
```
mdinger: Just watch that the `id` stays unique. I think it's a conflict if it doesn't. Might become less of an issue when rbe transfers to `cargo test`. Then the path will probably become an examples identification.

steveklabnik: /me nods

examples/expression/statements.rs
```diff
@@ -0,0 +1,6 @@
+fn main() {
+    // statement
+    // statement
+    // statement
+    // statement
+}
```
mdinger: You mean to merge `examples/expression/statements.rs` and `examples/expression/more_statements.rs`? Kinda looks like a mistake because you inlined the examples in the file.

examples/print/print.rs
mdinger: If I get to it...

examples/expression/statements.rs
steveklabnik: oh ugh, yes, i left these in by accident because i can only have one editor per page

PR #480, opened 2015-02-25
{} to {:?}

mdinger:
Switch `{}` to `{:?}` in the `println!` examples.

### DIFF
--- a/examples/print/print.rs
+++ b/examples/print/print.rs
@@ -1,23 +1,23 @@
 fn main() {
-    // `{}` are placeholders for arguments that will be stringified.
-    println!("{} days", 31);
+    // `{:?}` are placeholders for arguments that will be stringified.
+    println!("{:?} days", 31);
 
     // Without a suffix, 31 becomes an i32. You can change what type 31 is,
     // with a suffix, and that is covered in the next chapter.
 
     // The positional arguments can be reused along the template.
-    println!("{0}, this is {1}. {1}, this is {0}", "Alice", "Bob");
+    println!("{0:?}, this is {1:?}. {1:?}, this is {0:?}", "Alice", "Bob");
 
     // Named arguments can also be used.
-    println!("{subject} {verb} {predicate}",
+    println!("{subject:?} {verb:?} {predicate:?}",
              predicate="over the lazy dog",
              subject="the quick brown fox",
              verb="jumps");
 
-    // Special formatting can be specified in the placeholder after a `:`.
-    println!("{} of {:b} people know binary, the other half don't", 1, 2);
+    // Special formatting can be specified in the place of the `?`.
+    println!("{:?} of {:b} people know binary, the other half don't", 1, 2);
 
     // Error! You are missing an argument.
-    println!("My name is {0}, {1} {0}", "Bond");
+    println!("My name is {0:?}, {1:?} {0:?}", "Bond");
     // FIXME ^ Add the missing argument: "James"
 }


steveklabnik:
Thanks!

crumblingstatue:
Why were these changed to `{:?}`?
These can all be formatted with `{}`.
`{:?}` is for debugging.

GBGamer:
I agree with @crumblingstatue. Why were these changed? Especially since we don't see the suffixes until the next chapter, I think this'll just be confusing to new rust programmers.

mdinger:
If you just want something to print regardless, you use `{:?}`. All you possibly have to do is `derive(Debug)`. `{}` isn't reliable like that. It works for some types automatically and others require you implementing `Show` which is more complicated. From an instructing standpoint, it's simpler to use `Debug` and explain `Show` later.

Using `{}` will leave people trying to use it for other types which may or may not work. This is bad because examples need to be modifiable without breaking. `{}` can be examined as a separate construct later.
In my opinion, most examples should be switched from `{}` to `{:?}` though I haven't gone through trying to change everything in every example.

GBGamer:
Yeah, okay, I understand @mdinger's point. Never mind, I withdraw my
complaint. We should introduce `{}` soon, though, because I imagine people
getting annoyed when they automatically get `0i32` and they want `0`.

mdinger:
Yeah, that's reasonable.

crumblingstatue:
The examples should show idiomatic use of Rust.
Using fmt::Debug for user-facing output is non-idiomatic use of Rust.

Note that fmt::Display is (or should be) implemented for every standard library type that can be semantically expressed as a string.

If the user wanted to print his own type with {:?}, he wouldn't be able to, unless he derives Debug, which isn't explained here.

This example might also be an ugly first impression to the reader, seeing all the :? everywhere, strings being quoted in the output, etc.

It has already caused confusion: http://www.reddit.com/r/rust/comments/2ycl15/very_newb_printing_strings_without_quotes/

mdinger:
Idiomatic is nice but not required for good instruction. The example being clear is of primary importance (perhaps this isn't but that's beside the point). Idiomatic comes after understanding.

That thread isn't a problem for this example. It's that there isn't a follow up example for teaching `{}`. In the absence of `{}` being taught ,`{:?}` is a better default because it's simpler to setup for *all* types. Once there are both being taught then the distinction would be helpful.

crumblingstatue:
> In the absence of {} being taught ,{:?} is a better default because it's simpler to setup for all types.

The problem with `{:?}` is that its primary use is for debugging. If you want to print something to the user, use `{}`. If something is not meant to be printed, then there is no reason to print it. `{}` is harder to setup for all types because not all types are meant to be printed. If the user wants to print a custom type, he should define how it should be printed.

Normally, the reader wouldn't even be concerned with trying to print something that is not meant to be printed.

mdinger:
There is no rust IDE so debuggers are non-trivial to setup. Examining that something is working properly is basically easiest to deduce via `println!` variants. Stating this is uncommon seems unusual. Most example do either this or use `assert!` anyway.

You're arguing for more extensive examples though. The example isn't ideal and it's not going to be if it's changed or left the same. I'm setting up an issue to fix this. I'll cc you.

crumblingstatue:
> There is no rust IDE so debuggers are non-trivial to setup. Examining that something is working properly is basically easiest to deduce via println! variants. Stating this is uncommon seems unusual. Most example do either this or use assert! anyway.

I haven't stated that it is uncommon to debug by examining values with `{:?}`.
That is exactly the use for `{:?}`.
However, this example is concerned with printing things to the user, not debugging.
It produces incorrect user-facing output.

The output of this example is supposed to look like this:
```
31 days
Alice, this is Bob. Bob, this is Alice
the quick brown fox jumps over the lazy dog
1 of 2 people know binary, the other half don't
My name is Bond, James Bond
```

Instead, it looks like this:
```
31 days
"Alice", this is "Bob". "Bob", this is "Alice"
"the quick brown fox" "jumps" "over the lazy dog"
1 of 10 people know binary, the other half don't
My name is "Bond", "James" "Bond"
```

mdinger:
That's ugly but it works. You're asking for explaining `impl` and functions and other issues even before `let x = 3;` is introduced. Someone sees `struct A(i32); println!("{}", A(3));` but it won't print. Now they've got to add 5 lines of code so they can affirm it is indeed a 3. They don't necessarily know what the `!` means in `println!` or the difference between `struct A { a: i32}` and `struct A(i32)` but they're going to be expected to understand what `impl ...` means and why it's necessary.

People following examples always ***always*** modify them. You can't assume just because I used std lib defined `i32` that they won't stick a complicated user defined `enum` into it. Then they'll go to the IRC/Reddit/wherever trying to figure out why `i32` (which is automatic so they might not have noticed anyway) works but `struct Whatever` doesn't.

I don't think this discussion is going anywhere. More extensive examples are definitely needed though. And yeah, this example isn't ideal.

crumblingstatue:
The use of `#deriving(Debug)` isn't explained either.

So even if they use {:?}, it still won't work for their custom types.

mdinger:
It's easier to gloss over (punt) on `derive` than an `impl`

crumblingstatue:
It is indeed easier to `derive`. However, `derive` should still be explained on the same page, otherwise, the user would still be confused why printing his custom type doesn't work, and go to reddit/etc, like you said.

If we would explain `derive` on the same page, then we could easily say something to the effect of "If you want to print your custom type, use #derive(Debug) and {:?}", while retaining the {} format specifier for all the other examples.

Codenator81:
Last comment from  crumblingstatue look realy good for me

PR #481, opened 2015-02-25
flow_control/match/destructuring/destructure_structures: Link to the related "the ref pattern"

nickolay:
When looking up syntax of the ref pattern in destructuring let, I searched for 'destructuring' in the TOC and landed on this page. Couldn't find the 'ref pattern' page without looking through all the TOC.


mdinger:
Good idea

steveklabnik:
Thanks!

PR #485, opened 2015-02-25
rephrase tuple indexing

### LINKED ISSUE #333: Nits about tuples
From [tuples](http://rustbyexample.com/tuples.html):
```rust
// Values can be extracted from the tuple using the `valN` methods
    println!("long tuple first value: {}", long_tuple.0);
    println!("long tuple second value: {}", long_tuple.1);
```

The comment should be updated to something like:
```rust
// Values can be extracted from the tuple via tuple indexing:
```

May want a comment about generalizing it though:
```rust
fn main() {
    let tuple = (0i, 7i32);
    let N = 0i;
    
    // This does not work because tuple does not have a field named `N`
    tuple.N;
}
```

I think you can probably generalize it with a macro but when I tried, it split around the period: `tuple . N`. Probably a way around it but...

steveklabnik:
Fixes #333

### DIFF
--- a/examples/tuples/tuples.rs
+++ b/examples/tuples/tuples.rs
@@ -13,7 +13,7 @@ fn main() {
                       0.1f32, 0.2f64,
                       'a', true);
 
-    // Values can be extracted from the tuple using the `.N` syntax
+    // Values can be extracted from the tuple using tuple indexing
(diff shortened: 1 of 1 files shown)


steveklabnik:
Travis is laggy and this doesn't change anything, so I'm just gonna merge.

PR #487, opened 2015-02-26
Removed the extra s from bindings

chelseawinfree:
Just an extra s was tacked onto bindings. quick PR to remove that.


steveklabnik:
Thank you!

PR #488, opened 2015-02-26
Results: Fixing text wrapping bug

WrekitRalf:
Updated editor.css
Changed white-space on results to normal to stop text overflowing.


steveklabnik:
Thanks!

crumblingstatue:
In my opinion, the output shouldn't be wrapped, as that wouldn't give a faithful representation of the output, introducing line breaks where there shouldn't be.

It should, however be scrollable. That's how the Rust playpen works.

PR #491, opened 2015-03-04
Closures: Fix obsolete syntax: `&:`.

zachsitler:
Updated closures.rs.
Fixed the obsolete syntax so the example would run.


steveklabnik:
Thanks!

Issue #495, opened 2015-03-08
Expand println section


mdinger:
The [`println!`](http://rustbyexample.com/print.html) section should be expanded to include:
- [x] `print!`
- [x] `println!`
- [x] `format!`
- [x] `{}` vs `{:?}`
- [x] `{:b}` and other formatting parameters
- [x] Why doesn't `{}` work for vectors and generic containers?
- [ ] Explain `{:#?}` [RFC 640](https://github.com/rust-lang/rfcs/blob/master/text/0640-debug-improvements.md)

@crumblingstatue Note anything else that should be added. I'll add it to the list.

mdinger:
@crumblingstatue I don't think that last cc worked.

crumblingstatue:
I agree that more chapters could be added in the "Formatted print" section.
However, the first introduction to `println!` should use `fmt::Display` for formatting, as that is the natural way to write formatted output. `fmt::Debug` shouldn't be taught as the "default" way of writing formatted output, but it should certainly be mentioned, as it is a useful tool for debugging.

mdinger:
[reddit thread](http://www.reddit.com/r/rust/comments/2ycl15/very_newb_printing_strings_without_quotes/) highlighting confusion usage of `{:?}` created.

mdinger:
Added `{:#?}` to header. This has landed: http://is.gd/hdfaHi

mdinger:
https://github.com/rust-lang/rust-by-example/issues/196 is very similar to this issue but has different suggestions

coffeejunk:
It might be helpful to add a small paragraph about `print!` needing a `flush()` to immediately display text on `stdout`

(ref rust-lang/rust#23818)

steveklabnik:
Looks like there's only one checkbox left; I'd be more than happy to mentor anyone who wants to work on this!

maccoda:
@steveklabnik Looks like this one is also complete thanks to @vishalsodani :smile:

steveklabnik:
🎊

PR #496, opened 2015-03-08
Better print

mdinger:
Aims at fixing #495 by adding more detail about printing. Don't merge.

I gotta go do other things now but this is what I've got so far. May not get back to it for a few days.

First commit was a driveby fix.

### DIFF
--- a/examples/print/print.rs
+++ b/examples/print/print.rs
@@ -1,23 +1,34 @@
 fn main() {
-    // `{:?}` are placeholders for arguments that will be stringified.
-    println!("{:?} days", 31);
+    // In general, the `{}` will be automatically replaced with any
+    // arguments. These will be stringified.
+    println!("{} days", 31);
 
     // Without a suffix, 31 becomes an i32. You can change what type 31 is,
-    // with a suffix, and that is covered in the next chapter.
+    // with a suffix.
 
-    // The positional arguments can be reused along the template.
-    println!("{0:?}, this is {1:?}. {1:?}, this is {0:?}", "Alice", "Bob");
+    // There are various optional patterns this works with. Positional
+    // arguments can be used.
+    println!("{0}, this is {1}. {1}, this is {0}", "Alice", "Bob");
 
-    // Named arguments can also be used.
-    println!("{subject:?} {verb:?} {predicate:?}",
+    // As can named arguments.
+    println!("{subject} {verb} {predicate}",
              predicate="over the lazy dog",
              subject="the quick brown fox",
              verb="jumps");
 
-    // Special formatting can be specified in the place of the `?`.
-    println!("{:?} of {:b} people know binary, the other half don't", 1, 2);
+    // Special formatting can be specified after a `:`.
+    println!("{} of {:b} people know binary, the other half don't", 1, 2);
 
-    // Error! You are missing an argument.
-    println!("My name is {0:?}, {1:?} {0:?}", "Bond");
+    // It will even check to make sure the correct number of arguments are
+    // used.
+    println!("My name is {0}, {1} {0}", "Bond");
     // FIXME ^ Add the missing argument: "James"
+    
+    // Create a structure which contains an `i32`. Name it `Structure`.
+    struct Structure(i32);
+
+    // However, custom types such as this structure require more complicated
+    // handling. This will not work.
+    println!("This struct `{}` won't print...", Structure(3));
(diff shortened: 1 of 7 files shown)


mdinger:
cc @crumblingstatue You may be interested in this.

GBGamer:
Wow, this is awesome. :+1:

examples/print/print.rs
```diff
@@ -1,23 +1,34 @@
...
              verb="jumps");
 
-    // Special formatting can be specified in the place of the `?`.
-    println!("{:?} of {:b} people know binary, the other half don't", 1, 2);
+    // Special formatting can be specified after a `:`.
+    println!("{} of {:b} people know binary, the other half don't", 1, 2);
 
-    // Error! You are missing an argument.
-    println!("My name is {0:?}, {1:?} {0:?}", "Bond");
-    // FIXME ^ Add the missing argument: "James"
```
crumblingstatue: I don't think you should remove the FIXME here. It's helpful to tell the user exactly what he needs to do to fix this problem.

examples/print/print.rs
```diff
@@ -1,23 +1,34 @@
+// Create a structure which contains an `i32`. The structures
+// name is `FailedPrinter`.
+struct FailedPrinter(i32);
```
crumblingstatue: This feels confusing and out of place to me here at the top.
It would be better to declare this at the point we are trying to print it.
I also think `FailedPrinter` is a confusing name for it. It's not a printer. It's a custom type we're trying to print.

How about something like this instead?
```
// However, custom types such as this structure containing an
// `i32` require more complicated handling. This will
// not work.
struct MyCustomStruct(i32);
println!("This struct `{}` won't print...", MyCustomStruct(3));
// ^ FIXME: Comment this line out
```

examples/print/print.rs
steveklabnik: agreed

mdinger: Oh, I botched the build somewhere and I think I deleted this while tracking down the error. Forgot to add it back.

examples/print/print.rs
mdinger: I think your fix is sensible.

mdinger:
Pushed updates. Was working on this offline so I'm not sure how it gels with the previous comments. I'll have to compare again later.

I tried adding `print!` and `format!` but they're so similar that it didn't seem worth the extra space. In general, long examples aren't good so I dropped them. I've still gotta look it over again though.

steveklabnik:
Cool, when would you like me to review this?

mdinger:
I wanna sleep on it and look at it again tomorrow. I'll try to ping you tomorrow or the next day for review. I think it'll be ready then. I don't think there will be many big changes though. I'll update title from WIP too (gotta remember to do it...)

mdinger:
Something came up. Haven't gotten to it yet. I'll try to get to it in a day or so.

mdinger:
@steveklabnik r?
I rebased onto that other pull so I could build it. I don't know what it'll do if you push the other first. I assume it'll work though. I don't have any particularly strong criticism remaining about it other than `Display` is kinda long and I don't see much sense in splitting it into more sections.

@crumblingstatue I think I moved that specific `struct` you commented about and I tried to improve the naming in general though they follow naming schemes that makes sense to me. In general, I try to keep `struct` definitions outside of `main()` and rustbyexample historically does that too. I don't have any issues moving them inside where it makes sense though.

As before, comments welcome.

steveklabnik:
> I rebased onto that other pull so I could build it. I don't know what it'll do if you push the other first. I assume it'll work though

Oh, I didnt see this till now...

mdinger:
You want me to fix it? I'm kinda curious what it'll do...

steveklabnik:
It looks like everything is fine, right?

mdinger:
I think so. Since the first 3 patches already merged, it should just ignore them and add only the last. That's speculation though. If you don't want to risk it, I can rebase again so there's only a single patch.

Also, rather than closing #495 if you merge, can you check all but the last box. Or leave it to me. I'll do it.

mdinger:
@steveklabnik I went ahead and rebased to tip because it makes the PR cleaner though the commit/merge history shouldn't be different either way.

steveklabnik:
Okay! Let's land this. Thanks.

mdinger:
Cool

PR #498, opened 2015-03-11
Update use of std::env::args in arg chapter

caipre:
`std::os::args` is [deprecated](http://doc.rust-lang.org/std/os/fn.args.html).

`make tests` passed successfully.


steveklabnik:
Thank you so much!

PR #499, opened 2015-03-11
Playpen frontend: Will now properly display newlines in successful output

### LINKED ISSUE #494: Lines are not separated in the output
Take this example: http://rustbyexample.com/literals.html

The correct output (locally or playpen) is
```
1 + 2 = 3
1 - 2 = -1
true AND false is false
true OR false is true
NOT true is false
0011 AND 0101 is 0001
0011 OR 0101 is 0111
0011 XOR 0101 is 0110
1 << 5 is 32
0x80 >> 2 is 0x20
One million is written as 1000000
```

On rustbyexample, this is what I get with both Chromium 41 and Firefox 36:
```
1 + 2 = 3 1 - 2 = -1 true AND false is false true OR false is true NOT true is false 0011 AND 0101 is 0001 0011 OR 0101 is 0111 0011 XOR 0101 is 0110 1 << 5 is 32 0x80 >> 2 is 0x20 One million is written as 1000000
```

WaHe:
This should fix #494. The `handleProblem` function already does this properly when outputting errors. One thing to note is that lines wrap because of the fix for #488, which may cause some confusion. It may be worth it to have the result box scroll horizontally.

### DIFF
--- a/node_modules/gitbook-plugin-rust-playpen/book/editor.js
+++ b/node_modules/gitbook-plugin-rust-playpen/book/editor.js
@@ -219,7 +219,11 @@ function handleResult(statusCode, message) {
 // Called on successful program run
 function handleSuccess(message) {
   resultDiv.style.backgroundColor = successColor;
-  resultDiv.innerHTML = escapeHTML(message);
+  var lines = message.split(newLineRegex);
+  message = lines.map(function(line) {
+    return escapeHTML(line);
+  }).join('<br />');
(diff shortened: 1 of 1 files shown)


WaHe:
Example of the text wrapping on [type/literals](http://rustbyexample.com/type/literals.html): 
wrapping-exaple

steveklabnik:
Thank you so much!

PR #504, opened 2015-03-20
Fixed rustc-serialize issue

narutosanjiv:
+Compiling Error(make all) coming due to rust-serializer has been locked to 0.3.3 which does not have std::os::prelude::unix support, so it giving missing errors. 
error: unresolved import `std::os::unix::OsStringExt`. There is no `OsStringExt` in `std::os::unix`
error: unresolved import `std::os::unix::OsStrExt`. There is no `OsStrExt` in `std::os::unix`. 
Latest version of rust-serializer(0.3.6) has fixed this issue, so update Cargo.lock to fetch 0.3.6 version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,7 +3,7 @@ name = "update"
 version = "0.0.1"
 dependencies = [
  "regex 0.1.19 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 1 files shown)


steveklabnik:
Thanks so much!

Issue #505, opened 2015-03-22
process/pipe example panics because process.stdin is None


thepatrick:
```thread '<main>' panicked at 'called `Option::unwrap()` on a `None` value', /home/rustbuild/src/rust-buildbot/slave/nightly-dist-rustc-linux/build/src/libcore/option.rs:362```

I see the same output both when I run the example on rustbyexample.com and when I execute it locally (seen in both the nightly from a bout a week ago and today)

**Comment by mdinger** [comment]
For [process/pipe](http://rustbyexample.com/process/pipe.html), try replacing:
```rust
let process = match Command::new("wc").spawn() {
    Err(why) => panic!("couldn't spawn wc: {}", why.description()),
    Ok(process) => process,
};
```
With
```rust
let process = match Command::new("wc")
                            .stdin(std::process::Stdio::piped())
                            .stdout(std::process::Stdio::piped())
                            .spawn() {
    Err(why) => panic!("couldn't spawn wc: {}", why.description()),
    Ok(process) => process,
};
```

Seems to work properly. The stdio doesn't seem configured by default anymore.

mdinger:
This is fixed. Should be closed.

thepatrick:
Thanks!

PR #507, opened 2015-03-22
Fix warnings/errors in process/pipe

mdinger:
I'm not sure if that's the proper way to do the io but it seems to work and doesn't error. I figure more comprehensive examples will eventually arrive on the std library which will help.


mdinger:
Fixes https://github.com/rust-lang/rust-by-example/issues/505

steveklabnik:
Thanks! Yeah, IO doesn't have a lot of examples yet, working on it...

PR #508, opened 2015-03-22
Fix build

mdinger:
* Updates cargo lock
* Changes to `std::path::Path` in a few places
* Switches to `std::error::Error` in a few places


steveklabnik:
Thank you!

Issue #490, opened 2015-03-02
explain difference between `const` and `static`


vks:
Currently only `static` is introduced, `const` is never mentioned. (This is because historically `const` was introduced later.) However, `const` is the most common case, so the example needs to be updated.

steveklabnik:
This was fixed by https://github.com/rust-lang/rust-by-example/pull/512

PR #512, opened 2015-03-27
add info to constants

mdinger:
I've updated the bottom and I think it's better now than in my original change.

---

~~Here's a WIP for detailing the differences between `const` and `static`. I noticed it from [this](http://stackoverflow.com/questions/29225108/returning-slices-to-a-constant-array) stackoverflow question. I've still got to rewrite the `'static` lifetime section I added.~~

There are a bunch of other minor things that could be noted about constants on the RFC but probably shouldn't here.

Note: as written I'm not sure I like what I've written. There are differences between the two constants and what I've written seems accurate enough but I'm not sure it's worth pointing out here or not. Or that I've noted it in a good enough way.

Contants seem really simple. It doesn't seem right that they should need two pages...

### DIFF
--- a/examples/constants/constants.rs
+++ b/examples/constants/constants.rs
@@ -1,5 +1,6 @@
+// Globals are declared outside all other scopes.
 static LANGUAGE: &'static str = "Rust";
-static THRESHOLD: i32 = 10;
+const  THRESHOLD: i32 = 10;
 
 fn is_big(n: i32) -> bool {
     // Access constant in some function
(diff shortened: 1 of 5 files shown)


mdinger:
I'm not sure if I'll get back to this before next week or not so it may be a few days. Not sure.

vks:
If you want to simplify, only give examples for `const` and just mention that `static` exists for cases where initialization is expensive. Most of the time you just want use `const`.

mdinger:
That seems reasonable. At the time, I had found some of the minor differences between the types as surprising so when I tried to document, I tried to include too much info.

I think time since should give me a better perspective when I get back to it.

mdinger:
r? @steveklabnik 

The change isn't that drastic anymore. Basically I'm just trying to add in `const` and split off `'static` into it's own section of lifetimes. Global might even be able to be it's own section of a scope section but I'm not sure that section exists or that it should.

I have some bigger ideas I might run by you or I might just try them and see but every time I get close to trying them rbe is broken and fixing it pushes me back another week (I try not to spend too much time on this stuff).

mdinger:
Fixes https://github.com/rust-lang/rust-by-example/issues/490

steveklabnik:
Seems good, thank you! And sorry it took me a while to review.

Issue #429, opened 2015-01-24
Online editor highlight issue


caizixian:
It only highlights `i32` `i8` etc,but `is` `us`

mdinger:
https://github.com/ajaxorg/ace/pull/2325 merged to upstream and should fix this though it's untested. Currently though, ace is downloaded directly from [ajaxorg/ace-builds](https://github.com/ajaxorg/ace-builds) which releases periodically; so we're waiting for that to update.

[play.rust.org](https://play.rust-lang.org/) has the same issue but I'm not sure how they get new ace builds and when they get updates.

caizixian:
@mdinger LGTM.

mdinger:
This works now

PR #515, opened 2015-03-29
Use downloaded rust rules, not old ones

mdinger:
Syntax highlighting of `isize` and `usize` doesn't work yet in playpen because the `mode-rust.js` is out of date. This changes it to use the downloaded one instead which now includes syntax highlighting for those terms.

There are other ways to deal with this but since ace is being downloaded, it should be used.


mdinger:
https://github.com/rust-lang/rust-by-example/issues/429 was where this was originally noted though it was closed because upstream merged already.

cc @caizixian

caizixian:
@mdinger Copy that. Reopened.

Issue #516, opened 2015-03-30
cfg example perhaps refer to cfg! macro


mikedilger:
I was unaware of the `cfg!` macro and for a long time used the `cfg` attribute exclusively with much duplicated code.  Perhaps the http://rustbyexample.com/attribute/cfg.html page could include a small note about the macro

mdinger:
Good idea! I wasn't aware of this either. For reference, this refers to [`std::cfg!`](http://doc.rust-lang.org/std/macro.cfg!.html)

PR #517, opened 2015-03-31
Add the cfg! macro to cfg

mdinger:
It seems like it'd be better if macros and attributes each had their own section but there aren't two sections yet and this seems useful. Can be reorganized better later if that ever happens.

Fixes https://github.com/rust-lang/rust-by-example/issues/516

cc @mikedilger

### DIFF
--- a/examples/attribute/cfg/input.md
+++ b/examples/attribute/cfg/input.md
@@ -1,7 +1,16 @@
-The `cfg` attribute can be use to achieve conditional compilation.
+Conditional compilation is possible through two different operators:
+
+* the `cfg` attribute: `#[cfg(...)]` in attribute position
+* the `cfg!` macro: `cfg!(...)` in boolean expressions
+
+Both utilize identical syntax.
 
 {cfg.play}
 
-See [the
-Reference](http://doc.rust-lang.org/reference.html#conditional-compilation) for
-more details.
+### See also:
+
+[the reference][ref], [`cfg!`][cfg], and [macros][macros].
+
+[cfg]: http://doc.rust-lang.org/std/macro.cfg!.html
+[macros]: /macros.html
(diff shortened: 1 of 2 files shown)


examples/attribute/cfg/input.md
```diff
@@ -1,7 +1,15 @@
...
+* the `cfg!` macro: `cfg!(...)` in boolean expressions
+
+Both utilize identical syntax.
 
 {cfg.play}
 
-See [the
-Reference](http://doc.rust-lang.org/reference.html#conditional-compilation) for
-more details.
+### See also:
```
steveklabnik: there should be a space after this heading

examples/attribute/cfg/input.md
```diff
@@ -1,7 +1,15 @@
...
+[the reference][ref], [`cfg!`][cfg], and [macros][macros]
```
steveklabnik: There should be a period at the end here

examples/attribute/cfg/input.md
mdinger: @steveklabnik 
You mean like a newline? Like this:
```text
### See also:
// This line should be an empty blank line without comments
[the reference][ref], [`cfg!`][cfg], and [macros][macros]
```

I'm guessing this is what you mean but am surprised because these have all been like this. I usually cut and paste from somewhere else to keep formatting identical.

steveklabnik: yes, there should be that line between headings and the following paragraph.

Weird. I'll be changing them as I notice it, I guess.

mdinger: Fixed the newline (github mail really needs to do better context. it's not very helpful).

mdinger:
Added fixes.

Issue #528, opened 2015-04-04
move/mut example misleading 


brunobeltran:
The page on Mutability of references is misleading, since it only covers the semantics of Box.

Should we modify the code to explain the semantics of e.g. refs?

```
fn main() {
    let immutable_box = Box::new(5u32);

    println!("immutable_box contains {}", immutable_box);

    // Mutability error
    //*immutable_box = 4;

    // Hand over the box, changing the mutability
    let mut mutable_box = immutable_box;

    println!("mutable_box contained {}", mutable_box);

    // Modify the contents of the box
    *mutable_box = 4;

    println!("mutable_box now contains {}", mutable_box);

    let immutable_ref = &5u32;

    println!("immutable_ref contains {}", immutable_ref);

    // Mutability error
    //*immutable_ref = 4;

    // Handing over a ref does not allow mutability to change.
    // This "const"-correctness is one of the reasons refs are to
    // be preferred over Box<T>'s whenever possible.

    // This is allowed, since the compiler can detect that "variable
    // does not need to be mutable", although it causes a warning
    // and should be fixed
    let mut mutable_ref = immutable_ref;

    println!("mutable_ref to immutable value must still contain {}", mutable_ref);

    // Mutability error
    //*mutable_ref += 1;

    // the underlying value must be mutable for a mutable reference
    // to it to exist
    let mut x = 5u32;
    let ref mut mutable_ref = x;
    println!("mutable_ref to mutable value contains {}", mutable_ref);

    // so this works
    *mutable_ref = 4;
    println!("mutable_ref to mutable value contains {}", mutable_ref);
}
```

Or maybe we should have separate sub-sections for the semantics of Box<T>, Arc<T>, and &?

brunobeltran:
These issues are addressed partially in the next section, however the distinction between "pass by reference" and passing a Box<T> by value is not made explicit enough I think. I'll try to think of a better way to word.

mdinger:
See this thread also: https://github.com/rust-lang/rust-by-example/issues/390

marioidival:
Hi, I'm closing this issue just to clean up the items that already exist. If you think this issue makes sense to stay open, please create a new issue with updated information and mention this one.

thanks!

PR #529, opened 2015-04-04
move/mut: clarified semantics of immutable references

brunobeltran:
Related to issue #528.

### DIFF
--- a/examples/move/mut/mut.rs
+++ b/examples/move/mut/mut.rs
@@ -16,19 +16,15 @@ fn main() {
 
     println!("mutable_box now contains {}", mutable_box);
 
-    // on the other hand, suppose the variable itself is immutable
-    let x = 5u32;
+    // on the other hand, once a reference is made immutable
+    // it cannot be borrowed mutably.
+    let mut x = 5u32;
     let immutable_ref = &x;
+    // Compiler error!
+    let mutable_ref = &mut *immutable_ref;
 
-    println!("immutable_ref contains {}", immutable_ref);
-
-    // Mutability error
-    //*immutable_ref = 4;
-
-    // Compiler error! Can't force mutability by pointer indirection.
-    //let mutable_ref = &mut *immutable_ref;
-
-    // the underlying value must be mutable for a mutable reference to exist
+    // the underlying mutable value must be borrowed directly for a mutable
+    // reference to exist
     let mut x = 5u32;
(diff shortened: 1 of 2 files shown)


examples/move/mut/mut.rs
```diff
@@ -15,4 +15,32 @@ fn main() {
     *mutable_box = 4;
 
     println!("mutable_box now contains {}", mutable_box);
+
+    // on the other hand, suppose the variable itself is immutable
+    let x = &5u32;
+    let immutable_ref = &x;
```
mdinger: `let immutable_ref = x;` otherwise you end up with a double reference `&&5u32`.

This [page](http://doc.rust-lang.org/book/pointers.html#references) is more precise. You'll note he doesn't really mix and match left side `mut` and right side `mut` much. It gets more confusing when you do.

brunobeltran:
Thanks for the bug fix mdinger, I fixed in the next push. Do you think it would be better so simply link to the rust book page instead of adding the extra example code?

examples/move/mut/mut.rs
```diff
@@ -15,4 +15,32 @@ fn main() {
...
+
+    println!("immutable_ref contains {}", immutable_ref);
+
+    // Mutability error
+    //*immutable_ref = 4;
+
+    // This is allowed, since the compiler can detect that "variable does not
+    // need to be mutable", but generates a warning that should be fixed.
+    let mut mutable_ref = immutable_ref;
```
mdinger: I would expect `let mutable_ref = &mut *immutable_ref` (dereference and then reborrow) is probably clearer. I think `mut` on the left side makes you able to mutate what the reference points to but not necessarily the value. I think this whole thing can get confusing though.

brunobeltran:
After reading through the examples available again, it seems my concerns are addressed already in the subsequent sections, borrow and borrow/mut. Due to this, I don't think the current pull requests actually contributes constructively, so I'm closing it.

mdinger:
In general I think it's good for them both to approach each problem from a different angle if possible. Rustbyexample's coverage of borrow/ref is a bit disjointed and is spread out over the whole site. With time/effort, hopefully the examples will be grouped together better and become more cohesive.

Currently, each is better at certain things than others. The book is often better at in depth descriptions. Rustbyexample at quickly finding working code.

Issue #526, opened 2015-04-04
Explain rules for casting to smaller types in "Casting" section


brunobeltran:
The current code in the [Casting](http://rustbyexample.com/type/cast.html) section only has examples where the target of a cast can hold the value.

I think an explanation should be inserted as to why this code prints "still compares equal!" (although that comparison causes a warning, "comparison is useless due to type limits, #[warn(unused_comparisons)] on by default") and "I understand unsigned casting!"

    let x = 1000u16;
    let max_u8 = (std::u8::MAX as u16) + 1u16;
    let y = x as u8;
    let z = (x % max_u8) as u8;

    if 1000 == y {
        println!("Still compares equal!")
    } 
    if z == y {
        println!("I understand unsigned casting!")
    }

I would add it myself but I started learning Rust last night so I'm not confident whether Rust is truly implicitly casting 1000 to u8 to get 232 == y which evaluates to true, or whether the comparison is literally ignored as suggested by the warning. 

I also don't know how signed wrap-around works at all. (Discovered the above via an educated guess).

mdinger:
This is a good question. You could ask on IRC. Someone probably knows. Then post the [botbot](https://botbot.me/mozilla/rust/) log with the answer so someone can write it up unless you want to. If you don't, that's the method anyone else will take unless they already know the answer.

brunobeltran:
No one seemed to be able to agree on the answer (Steve thought that e.g. i8->u8 was gated), so I simply tested my best guess with some huge for loops and wrote up the answer when it worked, in PR #530.

mdinger:
Your PR merged. You should close this or modify the OP so that it states precisely what the remaining problems are.

brunobeltran:
Whoops, thanks for the reminder. I think that fixes my concern.

PR #530, opened 2015-04-04
type/cast fix

brunobeltran:
Added some examples related to casting integral types when the value cannot be represented by the target type, as well as an exact explanation of the semantics thereof to type/cast.

Related to issue #526

### DIFF
--- a/examples/type/cast/input.md
+++ b/examples/type/cast/input.md
@@ -1,4 +1,16 @@
 Rust provides no implicit type conversion (coercion) between primitive types.
 But, explicit type conversion (casting) can be performed using the `as` keyword.
 
+Rules for converting between integral types follow C conventions, with added
+provisions for conversion to signed types.
+
+Let val be the value being assigned, and let uT be the unsigned type with the
+same number of bits as the target type. Then
+1. If the value can be represented in the target type, the value is unchanged.
+2. Otherwise, if the two's-complement of the value can be represented in the
+   target type, the two's-complement of the value is the result.
+3. Otherwise, if the target type has fewer bits than the source type, the
+   resulting value is as if ( val % std::uT::MAX + 1 ) was performed with
+   exact arithmetic and reinterpreted according to the previous rules.
+
 {cast.play}
--- a/examples/type/cast/input.md
+++ b/examples/type/cast/input.md
@@ -9,7 +9,7 @@ same number of bits as the target type. Then
 1. If the value can be represented in the target type, the value is unchanged.
 2. Otherwise, if the two's-complement of the value can be represented in the
    target type, the two's-complement of the value is the result.
-3. Otherwise, if the target type has fewer bits than the source type, the
+3. Otherwise, the target type has fewer bits than the source type, and the
    resulting value is as if ( val % std::uT::MAX + 1 ) was performed with
    exact arithmetic and reinterpreted according to the previous rules.
 
--- a/examples/move/mut/input.md
+++ b/examples/move/mut/input.md
@@ -1,6 +1,8 @@
-Mutability of data can be changed when ownership is transferred.
+Mutability of data in a Box<t> can be changed when ownership is transferred.
 
-However, immutable data is protected from being accessed through
-mutable references at compile time.
+However, once an immutable reference to an object is borrowed, the reference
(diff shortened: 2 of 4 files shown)


mdinger:
I think [this](https://github.com/rust-lang/rfcs/blob/master/text/0560-integer-overflow.md) is the RFC which specifies it's behavior.

brunobeltran:
I thought so too, but careful reading shows that it left the semantics of "as" unchanged. The only comment that particular RFC makes on "as" is that 
"as was changed to restore the behavior before the RFC (that is, it truncates, as a C cast would)."

However, C casts can produce undefined behavior in various cases (e.g. (char)1000, the C equivalent of 1000i32 as i8). On the other hand, it appears after testing this on both the nightly and beta branches of Rust that Rust's behavior is well defined in these cases.

I have yet to find a document guaranteeing the behavior described the in PR, however.

brunobeltran:
The RFC on integer overflow links to [this](http://internals.rust-lang.org/t/on-casts-and-checked-overflow/1710/48) discussion on how "as" should work, which didn't seem to have a conclusion...

mdinger:
It doesn't seem to be fully implemented yet so some things may still need to be decided. See https://github.com/rust-lang/rust/issues/22020

brunobeltran:
That only seems to allow for the possibility that there will be optional
error checking added to as.

As I mentioned before, the actual RFC explicitly states that the current
semantics of "as" will be left as-is. So I think it's safe to include them
in the examples now that the beta has dropped.

examples/type/cast/input.md
```diff
@@ -1,4 +1,16 @@
 Rust provides no implicit type conversion (coercion) between primitive types.
 But, explicit type conversion (casting) can be performed using the `as` keyword.
 
+Rules for converting between integral types follow C conventions, with added
+provisions for conversion to signed types.
+
+Let val be the value being assigned, and let uT be the unsigned type with the
+same number of bits as the target type. Then
```
mdinger: Reads like a mathematical proof which seems different than most of the example (I hope they don't read like proofs...)

mdinger: I think these would be easier to read if they were interspersed through the example and they were less abstract. In the examples, the nearness of the numbers would make it less dense.

mdinger:
It should also be noted overflow it is considered an error, but an allowable error.

brunobeltran:
From my understanding, arithmetic overflow is considered and error, but
casting (for bit magic and so forth, such as the standard -1i32 as u32 to
get !0) is explicitly allowed.

I purposefully erred on the side of precision, but I'll push up a more
readable version soon, since I see now how much the current version clashes
with the rest of the website.

brunobeltran:
For some reason my most [recent push](https://github.com/brunobeltran/rust-by-example/commit/b8a0ada0a7b61e6002e35e1bfde1dc51dc6c8731) isn't showing up in the PR. 

I put the detailed explanations into the actual example code, which seems to make for a more readable experience, and used concrete examples while maintaining the correctness of the statement of the rules.

mdinger:
Eh...I don't know. You could try amending it and then force pushing. See if anything changes. Before being considered for merge, you probably should squash the commits too.

It looks much more readable and much less abstract than before.

brunobeltran:
I can't seem to get the fixed version to show up, even after a git push --force. [Commits are squashed](https://github.com/rust-lang/rust-by-example/compare/master...brunobeltran:cast-fix?expand=1). I force pushed it up, Obi-wan style, so it should show up as a single commit if you decide to merge. Otherwise I can reopen a new pull request, no big deal.

mdinger:
I don't have push rights. @steveklabnik does. I'm just trying to get it into better shape. Yeah, open up a new PR then.

mdinger:
Also, builds are failing currently and will continue until some of the other merges land. The beta playpen is also causing breakage. He might be waiting for that to resolve first, or he might just be prioritizing other things currently. Idk.

brunobeltran:
Makes sense. Thanks for your patience in either case, the new pull request is #533

mdinger:
You're welcome. Hope he approves it or critiques it and approves it.

PR #533, opened 2015-04-05
type/cast: added detailed integral type casting rules

brunobeltran:
Redo of PR #530


mdinger:
Yeah, this is much clearer/easier than before. This definitely seems like a good idea in general.

Later, further casts will need to be added in the form of [dynamic dispatch](http://doc.rust-lang.org/book/static-and-dynamic-dispatch.html#dynamic-dispatch) and [pointer manipulations](http://doc.rust-lang.org/book/unsafe.html#references-and-raw-pointers) where casts do some other things. That's further and another topic though.

steveklabnik:
Thanks!

PR #536, opened 2015-04-07
Some grammar fixups

frozencemetery:
Hello, here are some grammar fixes.  They are, of course, not the only ways to fix the sentences in question, so I will not be offended in the slightest if you decide to resolve them differently.

Thanks so much for making this guide!


steveklabnik:
Thank you!

Issue #532, opened 2015-04-05
Example code can not running  48.1 Argument parsing


netroby:
this code:

http://rustbyexample.com/arg/matching.html

48.1 Argument parsing
```
error: unstable feature
note: this feature may not be used in the beta release channel
error: aborting due to previous error
$ ./match_args Rust
This is not the answer.
$ ./match_args 42
This is the answer!
$ ./match_args do something
error: second argument not an integer
usage:
match_args <string>
    Check whether given string is the answer.
match_args {increase|decrease} <integer>
    Increase or decrease given integer by one.
$ ./match_args do 42
error: invalid command
usage:
match_args <string>
    Check whether given string is the answer.
match_args {increase|decrease} <integer>
    Increase or decrease given integer by one.
$ ./match_args increase 42
43
```

mdinger:
This is because rustbyexample currently uses the nightly channel. Hopefully https://github.com/rust-lang/rust-playpen/issues/83 will get fixed soon.

netroby:
The problem fixed or not?

mdinger:
Nope. Waiting on the linked bug.

netroby:
Seems not working. the page hang up on Running .....

mdinger:
This works now and should be closed

PR #539, opened 2015-04-09
Use nightly playpen

mdinger:
Valid options seem to be [`"beta"` or `"nightly"`](https://github.com/rust-lang/rust-playpen/blob/master/web.py#L62)


mdinger:
This fixes the playpen running on most examples. [hof](http://rustbyexample.com/hof.html) and maybe a few others still don't run yet because the playpen hasn't updated yet to a newer nightly.

Currently the playpen on rbe doesn't work because `"master"` isn't valid anymore.

mdinger:
Probably fixes https://github.com/rust-lang/rust-by-example/issues/532 too

steveklabnik:
Thanks!

PR #543, opened 2015-04-13
Fix first example of flow_control/while_let

CaseyLeask:
The original example doesn't compile.

Just wrapped it in the mandatory `fn main()` to avoid confusion.

### DIFF
--- a/examples/flow_control/while_let/input.md
+++ b/examples/flow_control/while_let/input.md
@@ -2,26 +2,28 @@ Similar to `if let`, `while let` can make awkward `match` sequences
 more tolerable. Consider, for example the following sequence:
 
 ```rust
-// Make `optional` of type `Option<i32>`
-let mut optional = Some(0);
+fn main() {
+    // Make `optional` of type `Option<i32>`
+    let mut optional = Some(0);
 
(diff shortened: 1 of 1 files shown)


mdinger:
I don't particularly like this. Sometimes it's useful to explain something without all the filler to make it compile. [print/debug](http://rustbyexample.com/print/print_debug.html) has a similar one too. `fn main() {}` would essentially just be noise there. If there were links to a playpen, then they could include the full runnable code but not here.

An alternative would be to have a button to "make this runnable" on the page similar to "Reset" and "Run" on playpen code blocks. That would an extra feature which needs implementing though.

So, it'd be awesome if everything was runnable but until the implementation supports it it seems bad to me. Other places would also benefit from being runnable too such as [benchmarking](http://rustbyexample.com/bench.html) and [testing](http://rustbyexample.com/test.html) but these are mainly limitations of the playpen. It doesn't support testing/benchmarking.

CaseyLeask:
Okay. I was just going through the examples and running them locally when I noticed that the sample didn't work.

It might be noise, but it's mandatory noise to get something running.

PR #545, opened 2015-04-14
Fix build failure by adding prefix

mdinger:
A fix for the [travis build failure](https://travis-ci.org/rust-lang/rust-by-example#L91) was suggested by @brson at https://github.com/rust-lang/rustup/issues/3 . It worked with a multi-line rustup on [travis](https://travis-ci.org/mdinger/rust-by-example/builds/58376675).

I think I got the single line right but I gotta see if this succeeds. Should be safe to merge if travis succeeds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: node_js
 sudo: true
 
 install:
-  - curl -L https://static.rust-lang.org/rustup.sh | sudo sh -s -- --channel=nightly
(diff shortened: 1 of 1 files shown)


mdinger:
Repushed so that the prefix goes to rustup instead of the shell. This passes travis now. Kinda strange breakage though.

steveklabnik:
Yeah, this is very strange. Thanks!

PR #546, opened 2015-04-15
Fix rustup

mdinger:
I don't have a lot of time but this makes rbe build. Timer could just as easily be deleted because it isn't in std currently. I replaced timer with thread functions where I could and just left it at that. Kinda sparse though. I'm not really that familiar with timer though.


steveklabnik:
Yeah, we should just get rid of timer, but we should also have a green build. Thanks!

PR #548, opened 2015-04-17
lifetime/struct: Change int to i32

lxrite:
If uncomment the struct Singleton.
Before this change, the error message is
```
error: use of undeclared type name `int`
```

After this change, the error message is
```
error: missing lifetime specifier [E0106]
```

And the second one it is exactly what we wanted.


steveklabnik:
Thanks so much!

PR #549, opened 2015-04-17
Make macros/dry tests pass.

IslandUsurper:
As written, the tests were failing because `test!` was building vectors of different sizes instead of building vectors filled with different values. Swapping the variable names into the right methods and fixing their types allows the tests to pass.


steveklabnik:
Thanks!

PR #551, opened 2015-04-17
Delete timers because no longer in std

mdinger:
Make sure it passes travis because I did this from windows. I'm assuming I didn't miss anything. Didn't test.


steveklabnik:
Thanks!

PR #552, opened 2015-04-18
Don't explicitly prefer Courier New

ysangkok:
It looks worse than the default font on FreeType systems with low DPI.

See http://askubuntu.com/q/28419/19466 and http://ubuntuforums.org/showthread.php?t=1752937

I can provide screenshots if necessary.

### DIFF
--- a/node_modules/gitbook-plugin-rust-playpen/book/editor.css
+++ b/node_modules/gitbook-plugin-rust-playpen/book/editor.css
@@ -16,7 +16,7 @@
   width: 100%;
   min-height: 72px;
   font-size: 13px;
-  font-family: Menlo, Monaco, Consolas, "Courier New", monospace;
+  font-family: Menlo, Monaco, Consolas, monospace;
   white-space: pre-wrap;
 }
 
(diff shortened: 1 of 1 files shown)


mdinger:
Can you provide screenshot comparisons? That would be helpful.

ysangkok:
With Courier New: http://i.imgur.com/ax4aRsy.png

With the default font (I believe it is DejaVu Sans Mono): http://i.imgur.com/rJSR3Fo.png

mdinger:
Is that why the code always looks grey? I wondered about that.

By the way, github can host the pics inline in the thread. You just drag them onto the "Write" thing in the website and github will start uploading the pic.

ysangkok:
Well, it looks more grey to me when using Courier New. But on Windows I don't think it's as light as it is with FreeType. So I can't say what the reason is on your system.

mdinger:
It definitely looks light on mine. If it fixes that, it'd be a little nicer to me too.

steveklabnik:
Seems good. Thanks!

PR #557, opened 2015-04-23
Update rust

mdinger:
There's another warning but I can't fix it yet because the fix which is to use `std::os::windows` doesn't seem to exist yet.


mdinger:
Fixes the build

steveklabnik:
Thanks!

Issue #558, opened 2015-04-24
Modules section breaks on `std::old_io::stdio::println`


CaseyLeask:
![screen shot 2015-04-24 at 7 21 02 pm](https://cloud.githubusercontent.com/assets/430277/7316080/1608acac-eab7-11e4-8fb8-38a9244f3943.png)

mdinger:
[Modules section](http://rustbyexample.com/mod.html)

CaseyLeask:
Thanks @mdinger, it's all fixed :+1:

PR #559, opened 2015-04-24
Remove old_io example

mdinger:
I don't know of a good replacement for this. Most simple functions are defined as methods which can be written with UFCS but I don't see the point in making the distinction here. Best I could think of is `from()` but I'm not thrilled with it. I don't see much harm in just dropping it.

### DIFF
--- a/examples/mod/mod.rs
+++ b/examples/mod/mod.rs
@@ -22,12 +22,6 @@ mod my {
 fn main() {
     function();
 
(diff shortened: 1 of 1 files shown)


mdinger:
Fixes https://github.com/rust-lang/rust-by-example/issues/558

steveklabnik:
agreed

PR #561, opened 2015-04-30
Reorganize rbe

mdinger:
[EDIT]
5/1/2015

* Still is a drastic reorganization of rbe
* Still reset to 100 character width
* I've had to ask how to do `fmt::Display` for `Vec` so I added a testcase explaining it. Testcases aren't supposed to be required reading at that point but are supposed to teach something useful if the user is interested or needs it. I think examples like that should be very helpful.
* Should pass travis
* I didn't always have anything great to say when removing stubs so they're probably not that great. I don't think they're currently that big an issue though.
* Some categories are tricky. `const` and `static` aren't really custom types nor are they primitives. Sometimes I just wasn't sure where to put them.
* *Formatting* in the print section just got stuck on the end. I didn't try to make it meld with the others.

---

4/29/2015
This is very WIP to get the gist of a reorganization of rbe that I'm trying. It does 2 main things:
* Set the character limit to 100 because rust uses 100. I ran into this because getting the `structure.json` file to globally use 4 space tabs made it too wide. Seems reasonable.
* Attempt to organize rbe into something sensible. It only seems to be organized at the beginning. The gist of it's current organization seems to be:

1. Explain most concepts, then
2. Introduce all the extraneous types in a somewhat random fashion though with more difficult things generally at the end.

This is problematic because finding things later can only be done via searching.

@steveklabnik Do you have any issues with this in general or the structure I chose? I'm not looking for a detailed answer. I just don't want to pursue something you aren't ever going to sign off on.

It's definitely easiest to comprehend if you build it manually. I tried to keep it as a skeleton so it's easier to review. If you don't have any issues with it, I'll try to get it into a little nicer shape for a later actual merge. I may actually have some time this week to play with it.

When building, don't do `make test` because it will probably fail.


steveklabnik:
:+1: looks great to me. You're right that there hasn't been a lot of attention paid to this, generally.

mdinger:
@steveklabnik r?

I updated the OP. I'm currently not sure what else I should change here. Maybe with a few days perspective but I think it's decent for merge. Comments definitely appreciated though.

@GBGamer Critiques also welcome.

PR #562, opened 2015-04-30
threads/threads.rs: change NTHREADS type to usize

tairrzayev:
That is also consistent with the type used in
channels/channels.rs

### DIFF
--- a/examples/channels/channels.rs
+++ b/examples/channels/channels.rs
@@ -2,13 +2,13 @@ use std::sync::mpsc::{Sender, Receiver};
 use std::sync::mpsc;
 use std::thread;
 
-static NTHREADS: usize = 3;
+static NTHREADS: i32 = 3;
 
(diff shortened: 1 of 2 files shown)


steveklabnik:
Hey @tairrzayev , thanks for this!

We should actually change the `channels.rs` to use `i32` instead: isize is the wrong type here. Care to make that change instead?

tairrzayev:
Hello @steveklabnik ,
Thank you for prompt response!

In channels.rs NTHREADS is usize since it is used in Vec::with_capacity(). Therefore when changed to i32 we have to add a cast which is not very nice.

Also, please, let me know if you would like to have channels.rs changes submitted in a separate PR.

steveklabnik:
Ahhh interesting, I didn't re-read that section, but that makes sense. That said, this is considered the better way of doing it. Thanks!

PR #564, opened 2015-05-03
Fix format traits based on API docs

robmv:
See Formatting traits at https://doc.rust-lang.org/std/fmt/


mdinger:
Nice. It's funny how this has been near the back (about chapter 50) for months probably. Shifting it to the front gets it fixed the first day.

steveklabnik:
Thanks!

PR #565, opened 2015-05-03
RGB(i32, i32, i32) does not require name

jyrkiput:
The RGB(i32, i32, i32) does not require name, so fix the comment.

At least this is how I understood what the code does.

### DIFF
--- a/examples/flow_control/match/destructuring/destructure_enum/enum.rs
+++ b/examples/flow_control/match/destructuring/destructure_enum/enum.rs
@@ -8,7 +8,7 @@ enum Color {
     Red,
     Blue,
     Green,
-    // This requires 3 `i32`s and a name.
+    // This requires 3 `i32`s.
     RGB(i32, i32, i32),
 }
 


steveklabnik:
Thanks!

mdinger:
The intent was 3 `i32`s and the name `RGB` as opposed to `HSV`, `HSL`, or `CMYK`. Thus:
```rust
RGB(i32, i32, i32),
HSV(i32, i32, i32),
HSL(i32, i32, i32),
CMYK(i32, i32, i32, i32),
```

No name would be `(i32, i32, i32)`. If you find it cleared without the name though, leave it without.

steveklabnik:
Ahhhh, I see. Yeah, I misunderstood. Hmm

mdinger:
It's unclear anyway otherwise he wouldn't have made the suggestion. Maybe HSV/HSL or something should be added.

@jyrkiput Would adding those `HSV(...)` and `HSL(...)` have made it clearer?

PR #566, opened 2015-05-03
Be more clear about enum variants name

mdinger:
This is another take on https://github.com/rust-lang/rust-by-example/pull/565.

@jyrkiput This should be more proper than the previous change but is it actually clearer to you?

@steveklabnik Don't merge immediately. I'm hoping he replies and I'm curious if he will think it's clearer. I didn't actually need to list all the color spaces but I figured I might as well use the main ones I found on wikipedia.

### DIFF
--- a/examples/flow_control/match/destructuring/destructure_enum/enum.rs
+++ b/examples/flow_control/match/destructuring/destructure_enum/enum.rs
@@ -8,8 +8,12 @@ enum Color {
     Red,
     Blue,
     Green,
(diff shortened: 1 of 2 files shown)


jyrkiput:
@mdinger Yep, this makes things pretty obvious.

mdinger:
Excellent. Thanks for confirming and pointing out how this example was slightly vague.

steveklabnik:
:confetti_ball:

PR #567, opened 2015-05-03
Fix broken links

mdinger:


mdinger:
I'm gonna pull these into my other PR because they collide somewhat

PR #570, opened 2015-05-06
Fix warning soft_link warning. Remove unnecessary Paths

mdinger:
* Fixes the `soft_link()` warning.
* Replaces most `Path`s with slices because they're unnecessary.

The windows one doesn't seem to exist so I just dropped it after applying a `cfg!`.


mdinger:
Updated. This should work and is really minor.

steveklabnik:
Thanks!

PR #572, opened 2015-05-14
Expand generics to cover more topics

mdinger:
From the first commit:

Expand generics to include sections on:

* generic functions
* generic traits
* expressing bounds
  - testcase covering empty bounds
* expressing multiple bounds
* explain where clauses
* explaining associated types
  - how to constrain associated types
  - added a testcase implementing an `Add` trait

Most of the other changes were to try to make the sections go together. I've think most things are in place. I'll probably look it over some more but I'd be happy to merge.


mdinger:
Ug. Forgot to run `test`. Fixed now and updated.

mdinger:
Added Associated items in a separate commit. I could squash again but I don't really see any need to.

steveklabnik:
@mdinger sorry about the delay. I'm going to review this tonight / tomorrow, I took the weekend off, and am just catching up :heart:

mdinger:
It's fine. Just so long as I have an approximate ETA I'm happy. A couple days is fine. It's kinda big anyway.

mdinger:
@steveklabnik What's the ETA on the review? I'm just looking for an estimate.

steveklabnik:
literally just opened this PR up to look :)

examples/generics/assoc_items/constrain/constrain.rs
```diff
@@ -0,0 +1,37 @@
...
+    type A = T;
+
+    // Clone to prevent move.
+    fn inner(&self) -> Self::A { self.0.clone() }
+}
+
+// Bounds on associated types.
+fn printer<C>(c: C) where
+    C: Contains,
+    C::A: Display {
```
steveklabnik: This isn't something to be fixed, but many people are including the `{` on the next line in this kind of situation. It's not clear how the style will work out, just a thought for the future :)

examples/generics/assoc_items/constrain/input.md
```diff
@@ -0,0 +1,30 @@
...
+    type A;
+    ...
+}
+
+// Bounds are applied through the container: `C::A: Trait`.
+fn apply_bounds<C>(c: C) where
+    C: Contains,
+    C::A: Display { ... }
+
+// Caveat: This hasn't been implemented yet and doesn't work.
```
steveklabnik: Why include something if it doesn't work yet?

examples/generics/assoc_items/types/types.rs
```diff
@@ -0,0 +1,50 @@
+struct Container(i32, i32);
+
+// A trait which will check to see if 2 items are stored inside of container.
```
steveklabnik: numbers less than ten should be spelled out

examples/generics/gen_fn/fn.rs
```diff
@@ -0,0 +1,34 @@
+struct T;          // Concrete type.
```
steveklabnik: I guess I see what you're doing here, but `T` as both a generic type parameter and a concrete type is confusing!

steveklabnik:
Thanks so much @mdinger ! Just a few nits

examples/generics/assoc_items/constrain/input.md
mdinger: Because the shorthand doesn't make any sense in terms of something people already understand. It's a shorthand for something else which happens too look like defaults applied to some type of generics.
```rust
fn apply_eq_shorthand<C>(c: C) where
    C: Contains<A = i32> { ... }
```

Explain the full form, then the shorthand for it and people can translate. The fact that the full syntax doesn't work seems just incidental and a bug (a weird one at that).

examples/generics/gen_fn/fn.rs
mdinger: Generics *are* confusing *and* [under explained](http://stackoverflow.com/questions/30399953/what-is-the-proper-way-to-create-a-new-generic-struct). There's no getting around it. The places `<T>` goes in generics are not explained well and when questions come up like why 2 `T`s in `impl<T> Foo<T> {}` then this is the example. The reason is because there could be some other `T` beforehand.

I sympathize though and wish there was a way to make it nicer.

mdinger:
Fixed the number. I don't see the others as requiring change so I just replied explaining the reasoning.

mdinger:
Thanks for reviewing. I think this updated generics section should be much more helpful than it was before and is sufficiently different from the book to complement it.

mdinger:
:fireworks: :sparkles: (I think that's my first ever emoticon)

Now the only really major things lacking are closures and lifetimes I think. I'm not sure which I'll try to tackle next but I'm not looking forward to lifetimes...

PR #577, opened 2015-05-24
Some random fixes

spastorino:
The important ones are https://github.com/spastorino/rust-by-example/commit/1580f7f920a3fabef8513786395fff0c78e70f09 and https://github.com/spastorino/rust-by-example/commit/d85950646234e2decc9de6cd2e88af447953ffc9
The rest are some related to the way some stuff is rendered and one is about a phrase which my bad english found weird so my fix is probably not better :).

### DIFF
--- a/examples/hello/comment/comment.rs
+++ b/examples/hello/comment/comment.rs
@@ -1,7 +1,7 @@
 fn main() {
     // This is an example of a line comment
     // Notice how there are two slashes at the beginning of the line
-    // And that anything written inside these will not be read by the compiler
+    // And that nothing written inside these will be read by the compiler
 
     // println!("Hello, world!");
 
--- a/examples/hello/print/input.md
+++ b/examples/hello/print/input.md
@@ -1,4 +1,4 @@
-Printing is handled by a series of [macros][macros] defined in [`std::fmt`][fmt]
+Printing is handled by a series of [`macros`][macros] defined in [`std::fmt`][fmt]
 some of which include:
 
 * `format!`: write formatted text to [`String`][string]
@@ -22,7 +22,7 @@ for these types. To print text for custom types, more steps are required.
 
 ### See also:
 
-[`std::fmt`][fmt], [macros][macros], [`struct`][structs],
+[`std::fmt`][fmt], [`macros`][macros], [`struct`][structs],
 and [`traits`][traits]
 
 [fmt]: http://doc.rust-lang.org/std/fmt/
--- a/examples/hello/print/input.md
+++ b/examples/hello/print/input.md
@@ -10,7 +10,7 @@ be checked at compile time.
 
 {print.play}
 
-[`std::fmt`][fmt] contains many [`trait`s][traits] which govern the display
+[`std::fmt`][fmt] contains many [`traits`][traits] which govern the display
 of text. The base form of two important ones are listed below:
 
 * `fmt::Debug`: Uses the `{:?}` marker. Format text for debugging purposes.
@@ -23,7 +23,7 @@ for these types. To print text for custom types, more steps are required.
 ### See also:
 
 [`std::fmt`][fmt], [macros][macros], [`struct`][structs],
-and [`trait`s][traits]
+and [`traits`][traits]
 
 [fmt]: http://doc.rust-lang.org/std/fmt/
 [macros]: /macros.html
--- a/examples/hello/print/print_display/input.md
+++ b/examples/hello/print/print_display/input.md
@@ -44,7 +44,7 @@ This is not a problem though because for any new *container* type which is
 {display.play}
 
 So, `fmt::Display` has been implemented but `fmt::Binary` has not, and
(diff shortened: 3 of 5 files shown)


examples/hello/comment/input.md
```diff
@@ -6,8 +6,9 @@ a few different varieties:
  - `/* Block comments which go to the closing deliminator. */`
 * *Doc comments* which are parsed into HTML library
 [documentation][docs]:
- - `/// Generate library docs for the following item.`
- - `//! Generate library docs for the enclosing item.`
+ - `/// Generate library docs for the item that follows the comment.`
+ - `//! Generate library docs for the enclosing item that follows the
```
mdinger: `//` means basically the same as before and is wordier. `//!` is wrong. Enclosing would be the module or file or something the `//!` is sitting inside, not what follows.

examples/hello/print/input.md
```diff
@@ -10,7 +10,7 @@ be checked at compile time.
 
 {print.play}
 
-[`std::fmt`][fmt] contains many [`trait`s][traits] which govern the display
+[`std::fmt`][fmt] contains many [`traits`][traits] which govern the display
```
mdinger: Typically we emphasize the keyword being used over the plural: that is  `trait`s over `traits`. Same for `enum`s, `struct`s, `fn`s, etc: the exact keyword is emphasized. The plural part is separated. It just looks a little odd because of the styling. Particularly when it's a link.

Making the link styling more similar to the code formatting could possibly be a good idea. Not sure.

examples/hello/print/print_display/input.md
```diff
@@ -4,7 +4,7 @@ customize the output appearance. This is done by manually implementing
 looks like this:
 
 ```rust
-// Import the `fmt` module to make it available via `use`.
+// Import (via `use`) the `fmt` module to make it available.
```
mdinger: This is good.

examples/hello/comment/comment.rs
```diff
@@ -1,7 +1,7 @@
 fn main() {
     // This is an example of a line comment
     // Notice how there are two slashes at the beginning of the line
-    // And that anything written inside these will not be read by the compiler
+    // And that nothing written inside these will be read by the compiler
```
mdinger: Both mean the same. It should be okay either way.

examples/hello/print/input.md
steveklabnik: :+1:

examples/hello/comment/input.md
steveklabnik: Yeah, I wasn't sure about this myself, and told @spastorino to submit it anyway. If you think it's worse, mabe we should take other changes but not this one

steveklabnik:
Thanks so much @spastorino ! I'm generally in favor of this, modulo the nits @mdinger brought up

examples/hello/print/input.md
mdinger: Is this +1 `trait`s or +1 better css or +1 `traits` or what? I don't actually care which way it goes but thought I'd point out how it's typical to write it like that.

examples/hello/comment/input.md
mdinger: You might be able to say: `//! Generate library docs for this module` which I think is true but I'm not sure. I'm not sure if there are any weird cases where it would be wrong.

What he wrote kinda implies this (regardless of if it's called `EnclosingItem`):
```rust
//! Document this enclosing item which houses an `i32` ...
struct EnclosingItem(i32);
```

I'm not in favor of these doc changes though as they stand.

examples/hello/print/input.md
steveklabnik: plus one to `trait`s.

spastorino:
Thanks for the review.

So we'd stick with https://github.com/spastorino/rust-by-example/commit/1580f7f920a3fabef8513786395fff0c78e70f09 and https://github.com/spastorino/rust-by-example/commit/ec6a824a8cc4c63ae3c964472668196736834441 ?

Asking just in case I didn't get exactly what you meant.

mdinger:
Everything is okay except https://github.com/spastorino/rust-by-example/commit/d85950646234e2decc9de6cd2e88af447953ffc9

spastorino:
I had doubts with https://github.com/spastorino/rust-by-example/commit/709f5bc57755f903d26e820bed9e86ad84fa0874, https://github.com/spastorino/rust-by-example/commit/a781bc2df40fb3f5a3341a6c2df8cb0ff74895e2 and https://github.com/spastorino/rust-by-example/commit/41475fb6d003da2dcdb4fbae10c3c30cd0627380 I thought you guys preferred the way it was.

mdinger:
I tried to explain why they were like they were. Steve said he liked the fix so they're okay. They will look less sloppy after your fix. You should keep them.

spastorino:
Ok, removed that commit and pushed again.
Thanks for reviewing.

steveklabnik:
Thanks! :D

PR #583, opened 2015-05-31
generics/phantom/testcase_units: alleviate the syntactical heaviness of the types and bounds

### LINKED ISSUE #582: generics/phantom/testcase_units: example adds unnecessary complication
The example is about demonstrating the usefulness of phantom types by implementing the Add trait for a Length type.

I think it is unnecessary to do it over a generic <T> type:
`struct Length<Unit, T>(T,PhantomData<Unit>);` should just be
`struct Length<Unit>(f64,PhantomData<Unit>);` (or `f32` as in the example)

The implementation:
```
impl<Unit, T> Add<Length<Unit, T>> for Length<Unit, T> where
    T: Add<T, Output=T> + Clone + Copy {
    type Output = Length<Unit, T>;

    fn add(self, r: Length<Unit, T>) -> Length<Unit, T> {
        Length(self.0 + r.0, PhantomData)
    }
}
```
can then be written:
```
impl<Unit> Add for Length<Unit> {
    type Output = Length<Unit>;

    fn add(self, rhs: Length<Unit>) -> Length<Unit> {
        Length(self.0 + rhs.0, PhantomData)
    }
}
```
Unless being able to implement Length over several type (f32, f64, i32, ...) is needed, there is no point to it, and I think this is not the point of this example (but I might be wrong).

Letheed:
I tried to retain the information on the bounds in a few comments because it does matter.
Close #582

### DIFF
--- a/examples/generics/phantom/testcase_units/input.md
+++ b/examples/generics/phantom/testcase_units/input.md
@@ -3,20 +3,22 @@ for a phantom type. The `Add` `trait` is examined below:
 
 ```rust
 // This construction would impose: `Self + RHS = Output`.
+// where RHS and Self are of the same type.
 pub trait Add<RHS = Self> {
     type Output;
 
     fn add(self, rhs: RHS) -> Self::Output;
 }
 
-// So, `Output` must be `T` and therefore, `T + T = T`.
-impl<T> Add<T, Output = T> for T {}
+// `Output` must be `T` so that `T + T = T`.
+impl Add for T {
+    type Output = T;
+    ...
+}
 
-// Similarly, this imposes: `S<T> + S<T> = S<T>` can be
-// added only when `T + T = T`.
-impl<S<T>> Add<S<T> for S<T> where
-    T: Add<T, Output = T> {
-    type Output = S<T>;
+// Similarly, this imposes: `T<U> + T<U> = T<U>`
+impl<U> Add for T<U> {
+    type Output = T<U>;
     ...
 }
 ```
--- a/examples/generics/phantom/testcase_units/input.md
+++ b/examples/generics/phantom/testcase_units/input.md
@@ -2,15 +2,15 @@ A useful method of unit conversions can be examined by implementing `Add`
 for a phantom type. The `Add` `trait` is examined below:
 
 ```rust
-// This construction would impose: `Self + RHS = Output`.
-// where RHS defaults to Self if not specified in the implementation
+// This construction would impose: `Self + RHS = Output`
+// where RHS defaults to Self if not specified in the implementation.
 pub trait Add<RHS = Self> {
     type Output;
 
     fn add(self, rhs: RHS) -> Self::Output;
 }
 
-// `Output` must be `T<U>` so that `T<U> + T<U> = T<U>`
+// `Output` must be `T<U>` so that `T<U> + T<U> = T<U>`.
 impl<U> Add for T<U> {
     type Output = T<U>;
     ...
--- a/examples/generics/phantom/testcase_units/input.md
+++ b/examples/generics/phantom/testcase_units/input.md
@@ -3,7 +3,7 @@ for a phantom type. The `Add` `trait` is examined below:
 
 ```rust
 // This construction would impose: `Self + RHS = Output`.
-// where RHS and Self are of the same type.
+// where RHS defaults to Self if not specified in the implementation
 pub trait Add<RHS = Self> {
     type Output;
 
--- a/examples/generics/phantom/testcase_units/input.md
+++ b/examples/generics/phantom/testcase_units/input.md
@@ -10,13 +10,7 @@ pub trait Add<RHS = Self> {
     fn add(self, rhs: RHS) -> Self::Output;
 }
 
-// `Output` must be `T` so that `T + T = T`.
-impl Add for T {
-    type Output = T;
-    ...
-}
-
-// Similarly, this imposes: `T<U> + T<U> = T<U>`
+// `Output` must be `T<U>` so that `T<U> + T<U> = T<U>`
 impl<U> Add for T<U> {
     type Output = T<U>;
     ...
--- a/examples/generics/phantom/testcase_units/units.rs
+++ b/examples/generics/phantom/testcase_units/units.rs
@@ -8,26 +8,30 @@ struct Inch;
 struct Mm;
 
 /// Length is phantom type with hidden parameter `Unit`
+/// f64 already implements the Clone and Copy traits
 #[derive(Debug, Clone, Copy)]
-struct Length<Unit, T>(T,PhantomData<Unit>);
+struct Length<Unit>(f64,PhantomData<Unit>);
 
-// This is similar to the header except `T` must also
-// implement `Clone` and `Copy.
-impl<Unit, T> Add<Length<Unit, T>> for Length<Unit, T> where
-    T: Add<T, Output=T> + Clone + Copy {
-    type Output = Length<Unit, T>;
+/// The Add trait defines the behavior of the `+` operator
+impl<Unit> Add for Length<Unit> {
+     type Output = Length<Unit>;
 
-    fn add(self, r: Length<Unit, T>) -> Length<Unit, T> {
-        Length(self.0 + r.0, PhantomData)
+	// add() returns a new Length struct containing the sum
+    fn add(self, rhs: Length<Unit>) -> Length<Unit> {
+		// `+` calls the Add implementation for f64
+        Length(self.0 + rhs.0, PhantomData)
     }
 }
 
 fn main() {
     // Specialize one_foot to have hidden parameter `Inch`
-    let one_foot:  Length<Inch, f32> = Length(12.0, PhantomData);
+    let one_foot:  Length<Inch> = Length(12.0, PhantomData);
     // one_meter has hidden parameter `Mm`
-    let one_meter: Length<Mm, f32>   = Length(1000.0, PhantomData);
+    let one_meter: Length<Mm>   = Length(1000.0, PhantomData);
 
+	// `+` calls the add() method we implemented for Length<Unit>
+	// Since Length implements Clone + Copy, add() does not consume
+	// one_foot and one_meter but makes a copy of them in `self` and `rhs`
     let two_feet = one_foot + one_foot;
     let two_meters = one_meter + one_meter;
 
--- a/examples/generics/phantom/testcase_units/units.rs
+++ b/examples/generics/phantom/testcase_units/units.rs
@@ -1,15 +1,15 @@
 use std::ops::Add;
 use std::marker::PhantomData;
 
-/// Null enumerations to define unit types
+/// Null enumerations define unit types.
 #[derive(Debug, Clone, Copy)]
 struct Inch;
 #[derive(Debug, Clone, Copy)]
 struct Mm;
 
-/// `Length` is phantom type with hidden parameter `Unit`
+/// `Length` is phantom type with hidden parameter `Unit`.
 ///
-/// `f64` already implements the `Clone` and `Copy` traits
+/// `f64` already implements the `Clone` and `Copy` traits.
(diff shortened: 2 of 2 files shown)


examples/generics/phantom/testcase_units/input.md
```diff
@@ -3,20 +3,22 @@ for a phantom type. The `Add` `trait` is examined below:
 
 ```rust
 // This construction would impose: `Self + RHS = Output`.
+// where RHS and Self are of the same type.
 pub trait Add<RHS = Self> {
```
mdinger: This isn't quite right. `Add<RHS = Self>` seems to force `RHS` to default to `Self`. So:

```rust
impl Add<f64> for T {} // T + f64 (override default)
impl Add for T // T + T (RHS is Self because of the default)
```

There is no defaults section though probably because I couldn't figure them out when I was working on the associated types sections.

examples/generics/phantom/testcase_units/input.md
```diff
@@ -3,20 +3,22 @@ for a phantom type. The `Add` `trait` is examined below:
...
+}
 
-// Similarly, this imposes: `S<T> + S<T> = S<T>` can be
-// added only when `T + T = T`.
-impl<S<T>> Add<S<T> for S<T> where
-    T: Add<T, Output = T> {
-    type Output = S<T>;
+// Similarly, this imposes: `T<U> + T<U> = T<U>`
+impl<U> Add for T<U> {
+    type Output = T<U>;
```
mdinger: I would be inclined to drop one of these 2 blocks because that super heavyweight bounds section isn't there anymore however you as the reader have better insight than I do on whether these are helpful. Therefore deferring to your judgement seems the most proper thing to do.

mdinger:
This seems good other than that assuming @steveklabnik doesn't have issues with it.

Letheed:
I removed one of the blocks as you suggested. I didn't do it at first because I wasn't sure how much you wanted me to depart from the initial example. But you're right, keeping the two was probably not necessary.

mdinger:
Awesome. If the other nit is fixed this seems good to me.

examples/generics/phantom/testcase_units/input.md
mdinger: Note: this isn't your mistake. It just wasn't written correctly before.

It's mainly an issue because it uses `<X = Y>` but the actual example doesn't require that part.

Letheed: Apologies, I missed that comment yesterday... Well I didn't know about that defaulting mechanism indeed, it's nice.
I updated the comment to mention that. I would have removed that line but since it's not explained anywhere else, it does get confusing. Let me know if there's anything else.

mdinger:
Nice! Looks good to me. Now just wait for official review from @steveklabnik.

examples/generics/phantom/testcase_units/input.md
```diff
@@ -3,20 +3,16 @@ for a phantom type. The `Add` `trait` is examined below:
 
 ```rust
 // This construction would impose: `Self + RHS = Output`.
```
steveklabnik: since you've added below, the period at the end here should be removed

examples/generics/phantom/testcase_units/input.md
```diff
@@ -3,20 +3,16 @@ for a phantom type. The `Add` `trait` is examined below:
...
 
-// So, `Output` must be `T` and therefore, `T + T = T`.
-impl<T> Add<T, Output = T> for T {}
-
-// Similarly, this imposes: `S<T> + S<T> = S<T>` can be
-// added only when `T + T = T`.
-impl<S<T>> Add<S<T> for S<T> where
-    T: Add<T, Output = T> {
-    type Output = S<T>;
+// `Output` must be `T<U>` so that `T<U> + T<U> = T<U>`
```
steveklabnik: period at the end here please!

examples/generics/phantom/testcase_units/units.rs
```diff
@@ -8,26 +8,30 @@ struct Inch;
 struct Mm;
 
 /// Length is phantom type with hidden parameter `Unit`
+/// f64 already implements the Clone and Copy traits
```
steveklabnik: There should be an extra `///` between this line and the above one, and `f64`, `Clone`, and `Copy`, should be in backticks

examples/generics/phantom/testcase_units/units.rs
```diff
@@ -8,26 +8,30 @@ struct Inch;
...
 #[derive(Debug, Clone, Copy)]
-struct Length<Unit, T>(T,PhantomData<Unit>);
+struct Length<Unit>(f64,PhantomData<Unit>);
 
-// This is similar to the header except `T` must also
-// implement `Clone` and `Copy.
-impl<Unit, T> Add<Length<Unit, T>> for Length<Unit, T> where
-    T: Add<T, Output=T> + Clone + Copy {
-    type Output = Length<Unit, T>;
+/// The Add trait defines the behavior of the `+` operator
```
steveklabnik: period at the end, `Add` in backicks

examples/generics/phantom/testcase_units/units.rs
```diff
@@ -8,26 +8,30 @@ struct Inch;
...
+impl<Unit> Add for Length<Unit> {
+     type Output = Length<Unit>;
 
-    fn add(self, r: Length<Unit, T>) -> Length<Unit, T> {
-        Length(self.0 + r.0, PhantomData)
+	// add() returns a new Length struct containing the sum
```
steveklabnik: This comment is at the wrong indentation for its function, and should end with a period. Also, `Length` in backticks

examples/generics/phantom/testcase_units/units.rs
```diff
@@ -8,26 +8,30 @@ struct Inch;
...
+    fn add(self, rhs: Length<Unit>) -> Length<Unit> {
+		// `+` calls the Add implementation for f64
```
steveklabnik: `Add` and `f64` in backticks, period at the end.

examples/generics/phantom/testcase_units/units.rs
```diff
@@ -8,26 +8,30 @@ struct Inch;
...
 fn main() {
     // Specialize one_foot to have hidden parameter `Inch`
-    let one_foot:  Length<Inch, f32> = Length(12.0, PhantomData);
+    let one_foot:  Length<Inch> = Length(12.0, PhantomData);
     // one_meter has hidden parameter `Mm`
-    let one_meter: Length<Mm, f32>   = Length(1000.0, PhantomData);
+    let one_meter: Length<Mm>   = Length(1000.0, PhantomData);
 
+	// `+` calls the add() method we implemented for Length<Unit>
+	// Since Length implements Clone + Copy, add() does not consume
```
steveklabnik: there should be an extra `//` above this line, and `Length`, `Clone`, `Copy`, and `add()` in backticks

examples/generics/phantom/testcase_units/units.rs
```diff
@@ -8,26 +8,30 @@ struct Inch;
...
 
 fn main() {
     // Specialize one_foot to have hidden parameter `Inch`
-    let one_foot:  Length<Inch, f32> = Length(12.0, PhantomData);
+    let one_foot:  Length<Inch> = Length(12.0, PhantomData);
     // one_meter has hidden parameter `Mm`
-    let one_meter: Length<Mm, f32>   = Length(1000.0, PhantomData);
+    let one_meter: Length<Mm>   = Length(1000.0, PhantomData);
 
+	// `+` calls the add() method we implemented for Length<Unit>
```
steveklabnik: `add()` and `Length<Unit>` in backticsk

examples/generics/phantom/testcase_units/units.rs
```diff
@@ -8,26 +8,30 @@ struct Inch;
...
+	// Since Length implements Clone + Copy, add() does not consume
+	// one_foot and one_meter but makes a copy of them in `self` and `rhs`
```
steveklabnik: `one_foot` and `one_meter` in backticks

steveklabnik:
Hey @Letheed ! Thanks so much for this, and sorry it's taken me a while to review. I agree with @mdinger that this is much better, just some formatting nits. Thank you!

mdinger:
@steveklabnik you probably want to skim https://github.com/rust-lang/rust-by-example/issues/582 where other options are discussed for changing this as well. Note: I'm not inclined to make any of the changes suggested currently but someone else might take it and run with it.

Letheed:
No worries... You're welcome, the website was ~very~ helpful ! :+1: 
Rust's official doc is rather exhaustive but can be a little intricate for beginners. Also examples help a lot.

I've tried to follow your instructions, although there's still one or two details I'm not sure about, mostly regarding periods.

examples/generics/phantom/testcase_units/input.md
```diff
@@ -2,21 +2,17 @@ A useful method of unit conversions can be examined by implementing `Add`
 for a phantom type. The `Add` `trait` is examined below:
 
 ```rust
-// This construction would impose: `Self + RHS = Output`.
+// This construction would impose: `Self + RHS = Output`
+// where RHS defaults to Self if not specified in the implementation.
```
Letheed: I moved the period to the end of the sentence. I'm not sure if you wanted it removed completely, but I thought that would be inconsistent with line 13.

examples/generics/phantom/testcase_units/units.rs
```diff
@@ -7,27 +7,33 @@ struct Inch;
...
 #[derive(Debug, Clone, Copy)]
-struct Length<Unit, T>(T,PhantomData<Unit>);
+struct Length<Unit>(f64,PhantomData<Unit>);
 
-// This is similar to the header except `T` must also
-// implement `Clone` and `Copy.
-impl<Unit, T> Add<Length<Unit, T>> for Length<Unit, T> where
-    T: Add<T, Output=T> + Clone + Copy {
-    type Output = Length<Unit, T>;
+/// The `Add` trait defines the behavior of the `+` operator.
```
Letheed: I'm not sure why you wanted a period added here and at line 20, since the rest of the website never uses periods in the playpens.

mdinger: The official [policy](https://github.com/rust-lang/rfcs/blob/1f4cf0e02bb2fe3a0340cacecde9e22b8d35391d/text/0000-api-comment-conventions.md#formatting) is that comments should be punctuated. Generally that's fairly easy to check in small patches but rbe has had many many sweeping changes in the past year where it's hard to notice little things like that.

Ideally rbe would start conforming more though to the standard. The situation should get better with time and stability.

examples/generics/phantom/testcase_units/input.md
steveklabnik: Overall, we may have inconsistencies. Rust by Example was donated to the Rust organization, and I haven't had the chance to go over the entire thing yet. So, when adding new stuff, I tend to apply our already-existing conventions, even though the project as a whole may not have totally adopted them yet. Make sense?

examples/generics/phantom/testcase_units/units.rs
steveklabnik: @mdinger yes, thanks you beat me to it,  and i just mentioned it https://github.com/rust-lang/rust-by-example/pull/583#discussion_r31945527

mdinger:
I'm glad the sites helpful. The site doesn't get many critiques anymore. They're all directed at [the book](http://doc.rust-lang.org/book/)

examples/generics/phantom/testcase_units/units.rs
Letheed: So I should terminate the rest of the comments too ?

steveklabnik: If you'd like, yes. Thanks :)

steveklabnik:
Ah! I didn't see that this was fixed. Thanks so much @Letheed !

PR #584, opened 2015-05-31
meta/doc link to rustdoc leads to old location

asdofindia:
http://rustbyexample.com/meta/doc.html has a link to rustdoc which points to https://github.com/rust-lang/rust/blob/master/src/doc/rustdoc.md which has moved to book. 
This PR directly links to the rustdoc documentation.


steveklabnik:
Thanks!

PR #586, opened 2015-06-03
std/hash: HashMap::insert() returns Option<V>, not bool

Letheed:


steveklabnik:
Thanks so much! I apologize this took me a while.

PR #588, opened 2015-06-06
std_misc/process/pipe: Update comments to match current API.

sourcefrog:


examples/std_misc/process/pipe/pipe.rs
```diff
@@ -16,7 +16,9 @@ fn main() {
     };
 
     {
-        // Write a string to the stdin of `wc`
+        // Write a string to the stdin of `wc`. 
+        // stdin has type `Option<ChildStdin>`, but since we know this instance
```
steveklabnik: Two small nits: `stdin` should be in backticks, and there should be an extra `//` between your first and second line.

steveklabnik:
This looks great, just a few small nits

sourcefrog:
Thanks, please take another look.

steveklabnik:
Thank you! :)

steveklabnik:
Hehe whoops, forgot we don't use Bors here! Thanks again.

PR #590, opened 2015-06-08
Explain HOFs in terms of functions instead of closures

spastorino:


examples/fn/hof/hof.rs
```diff
@@ -26,11 +26,11 @@ fn main() {
...
         (0..).map(|n| n * n)             // All natural numbers squared
              .take_while(|&n| n < upper) // Below upper limit
-             .filter(|n| is_odd(*n))     // That are odd
+             .filter(is_odd)             // That are odd
              .sum();                     // Sum them
     println!("functional style: {}", sum_of_squared_odd_numbers);
 }
 
-fn is_odd(n: u32) -> bool {
+fn is_odd(n: &u32) -> bool {
```
steveklabnik: So one unfortunate thing about this change is that, while the `filter` call is more clear, you'd never want to take an `i32` by reference, because it's more expensive than taking it directly. I'm not sure what to do about this.

steveklabnik:
I'm in favor of the second part, but unsure about the first.

examples/fn/hof/hof.rs
spastorino: Yep I understand, the example ends being a bit weird.

Maybe the best thing we can do is change the example a bit more to be able to pass a pure function instead of a closure.

steveklabnik:
Yeah, I think that the tradeoff here isn't worth it. Thanks though!

examples/fn/hof/input.md
```diff
@@ -1,6 +1,6 @@
-Rust provides Higher Order Functions (HOF). These are functions that take a
-closure as an argument to produce a more useful function. HOFs and lazy
-iterators give Rust its functional flavor.
+Rust provides Higher Order Functions (HOF). These are functions that
+take one or more functions and/or produce a more useful function. HOFs
+and lazy iterators give Rust its functional flavor.
```
spastorino: Shouldn't we change anyway this ^^^ part?. If so I can provide a PR with that stuff.

steveklabnik: yeah, i'd take one for just this

spastorino: @steveklabnik can you reopen this PR? I've done a push -f but it seems to not be updating because it's closed.

steveklabnik: github says it won't let me re-open because you've force pushed, just send a new PR

spastorino: https://github.com/rust-lang/rust-by-example/pull/611

PR #595, opened 2015-06-14
This overlap is currently allowed

spastorino:


mdinger:
why?

mdinger:
That comment may not be precise enough but specialization still doesn't work. See [this](https://users.rust-lang.org/t/specialization-of-a-generic-method-for-a-struct/1309) page which includes [this](http://is.gd/ruH36a) example which doesn't work.

spastorino:
@mdinger I've tested the code shown here http://rustbyexample.com/generics/impl.html and does compile correct. Check http://is.gd/j58Fyq, am I missing something?

mdinger:
If specialization fails in the general cases then it makes sense to state specialization isn't currently allowed. Thus the note seems valid except with null `impl`s. It should be rephrased so it's accurate.

spastorino:
I have mainly sent the PR because `The 3 separate implementations of GenericTup above cannot coexist.` seemed to me very specific to just the example. So yeah maybe as you said we can rephrase to state something more accurate.

mdinger:
Yes but that specific note was targeted at answering [this thread](https://users.rust-lang.org/t/specialization-of-a-generic-method-for-a-struct/1309) so it wouldn't be a frequent question. It would definitely be better to rephrase so the statement is accurate. Once specialization is available, then the note can be removed.

steveklabnik:
@spastorino can you rebase this please?

spastorino:
@steveklabnik done!

steveklabnik:
Thanks! :)

PR #596, opened 2015-06-14
GenTup is not a tuple so name it as GenVal

spastorino:


wycats:
More notably, what's the value of using a tuple (with the confusing `(T,)` syntax) at this stage?

spastorino:
Have been talking with @wycats and yes agree with what he says, the thing is that all that adds just noise in this example so the commit message and pr title is not the best.

mdinger:
Not much. rbe has [historically](http://rustbyexample.com/custom_types/structs.html) distinguished between structs with like `S { a: b, c: d}` and tuplestructs like`T(a, b)` and so `T(a,)` might be useful for showing it uses the tuple syntax. Those 2 sections probably used to be much closer together also.

I have no preference. I typically call everything structures with varying format showing a distinction. Some people may like the distinction in names: for example: https://github.com/rust-lang/rust-by-example/commit/edd7059eb1f56e690fd4140f4a5923e049ea7dac was moving towards tuple structs. This suggests moving away.

I do think one way should be chosen and it should be followed if possible.

PR #598, opened 2015-06-19
std/str: Fix warning: deprecated item

jarcane:
This sample throws a warning in line 36 currently, which does not appear to be intended: 

```
warning: use of deprecated item: use `String::from` instead, #[warn(deprecated)] on by default
```
As making the suggested change produces this warning instead: 
```
warning: unused or unknown feature, #[warn(unused_features)] on by default
``` 
...the reference to `#![feature(collections)]` is removed as well.

### DIFF
--- a/examples/std/str/str.rs
+++ b/examples/std/str/str.rs
@@ -1,5 +1,3 @@
-#![feature(collections)]
-
 fn main() {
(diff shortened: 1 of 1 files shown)


steveklabnik:
Thanks!

PR #602, opened 2015-06-24
Update methods perimeter expression

ner0x652:
In my opinion this version is more readable, also theoretically faster.


steveklabnik:
Thanks!

Issue #603, opened 2015-06-24
cast: The example code makes no mention of warn(overflowing_literals)


mpe:
When you run the cast example there are lots of warnings, but the code doesn't make any mention of them, which is confusing. 

```
warning: literal out of range for u8, #[warn(overflowing_literals)] on by default
note: in expansion of format_args!
<std macros>:2:25: 2:56 note: expansion site
<std macros>:1:1: 2:62 note: in expansion of print!
<std macros>:3:1: 3:54 note: expansion site
<std macros>:1:1: 3:58 note: in expansion of println!
```

etc.

These seem to be caused by the literals that are cast, ie. not the casts themselves. eg. it's fixed by changing:

```println!("1000 as a u8 is : {}", 1000 as u8);```

to

```println!("1000 as a u8 is : {}", 1000u16 as u8);```

mdinger:
Please link to the example. Makes it easier to find. The [cast example](http://rustbyexample.com/cast.html) in question.

A simple fix is to add `#![allow(overflowing_literals)]` at the beginning of the example like https://github.com/rust-lang/rust-by-example/pull/605 does.

PR #605, opened 2015-06-24
Suppress warnings on cast by default

mdinger:
Fixes https://github.com/rust-lang/rust-by-example/issues/603


steveklabnik:
Thanks!

Issue #601, opened 2015-06-23
custom_types/enum: The first enum example is very complicated


mpe:
The enum example uses namespacing, boxes, and match, as well as talking about concrete types, references and borrowing, none of which has been introduced yet (@ 3.2).

I realise sometimes it's impossible to only use concepts which have already been introduced, but this seems like it could be simpler, or if not, moved later.

mdinger:
Well, it definitely needs references at the bottom for all those topics at a minimum. `Box` should probably stay or move to a separate page of `enum`s because a linked list is usually created via an `enum`.

FWIW, I'm pretty sure it used to be much later in the tutorial which is why it has so many concepts in it. I'm definitely for simplifying examples.

How do you like the [destructuring/enums example](http://rustbyexample.com/flow_control/match/destructuring/destructure_enum.html) for comparison?

mpe:
Moving it later would be best IMHO. Or drastically simplifying it, though I don't actually understand it well enough to say if that's even possible.

The destructuring/enums example is much better. That would be better at 3.2 I think, with a little more explanation in comments about what's going on.

mpe:
Just another note, the example at 3.2.1 does a much better job of explaning the use syntax:

````
// If you'd like to refer to the days without saying
// `Day::Monday`, then you use the following notation
// use Day::{
//     Monday,
...
````

Whereas in the List example it looks at first glance as if you're use'ing List and Cons from somewhere else:

````
// Allow Cons and Nil to be referred to without namespacing
use List::{Cons, Nil};
````

mdinger:
Well, the examples are no longer strictly in order because it's more direct to just link to simple explanations.

Regarding moving `enum` destructuring earlier, my preference is to explain destructuring in one general location so the span of what destructuring can do is more obvious. Adding simple examples to the `enum` section to elaborate on a specific point is something that could be done though.

PR #606, opened 2015-06-26
Simplify the enum section examples

mdinger:
* Moves the original enum example into a linked list testcase
* Moves `struct` visibility into the visibility section (people when learning how to create structs may not care about how `pub` interacts but when looking at visibility they would want to know that that it interacts with `structs` though. Seems a slightly positive change but I don't feel too strongly about it either way. Could use more visibility info there probably also).
* Adds a specific example aimed at `use` usage in regard to `enums`.
* Added *See also:* links where ever seemed good.
* Tried to drop methods and box usage when possible. It really complicates the examples.
* The new main *enum* example is very similar to the [destructuring/enums](http://rustbyexample.com/flow_control/match/destructuring/destructure_enum.html) which I tried to avoid but it's difficult. `enum` and `match` go hand in hand.

@steveklabnik Sorry for the messy large commit. I can shoehorn it into a little nicer order but it's messier than I like.

@mpe comments/critiques welcome as this aims at fixing your issue. Also, the testcase examples are *supposed* to be optional which you can look at if you're interested. They may cover more material than most examples at that point use so hopefully the reference linking is pretty decent.

As always, easiest to review if you manually build the docs.

Fixes https://github.com/rust-lang/rust-by-example/issues/601


examples/custom_types/enum/enum_use/use.rs
```diff
@@ -0,0 +1,36 @@
+// An attribute to hide warnings for unused code.
+#![allow(dead_code)]
+
+// Explicitly `use` each name so they are available without
+// manual scoping.
+use Status::{Poor, Rich};
```
mpe: The comment is good, but it would be even clearer I think if the ```use Status``` was below the definition of ```Status``` below. That makes it 100% clear that you're using the ```Status``` you just declared, vs importing it from somewhere.

examples/custom_types/enum/enum_use/use.rs
```diff
@@ -0,0 +1,36 @@
...
+
+// Automatically `use` each name inside `Work`.
+use Work::*;
```
mpe: And similarly here, I'd move this below the definition of ```Work```.

mpe:
Otherwise looks good to me.

examples/custom_types/enum/enum_use/use.rs
```diff
@@ -0,0 +1,37 @@
...
+    Civilian,
+    Soldier,
+}
+
+fn main() {
+    // Explicitly `use` each name so they are available without
+    // manual scoping.
+    use Status::{Poor, Rich};
+    // Automatically `use` each name inside `Work`.
+    use Work::*;
```
mdinger: Added a commit moving these into `main()`  (I was surprised it works. I don't think it used to work).

@steveklabnik You have any issues with this? I kinda think it goes against style guidelines. Typically they are all listed at the top of the module.

mpe: They could also be outside main but after the enum definitions and I think that would still be clear.

steveklabnik: it has worked for a long time. Most people put them outside it's true.

PR #607, opened 2015-06-28
Rewrite lifetimes section

### LINKED ISSUE #154: Reason for using generics in lifelines isn't clear
In [Lifetimes/Functions](http://rustbyexample.com/lifetime/fn.html), the example:

```rust
struct Triplet { one: int, two: int, three: int }
impl Triplet {
    // Fails
    fn mut_one(&mut self) -> &mut int { &mut self.one }
    // Fails
    fn mut_two<'s, 'f>(&'s mut self) -> &'f mut int { &mut self.two }
    // Works
    fn mut_three<'s>(&'s mut self) -> &'s mut int { &mut self.three }
}
```
You switch from`fn mut(...)` to `fn mut<...>(...)` without explanation. `Triplet` isn't generic so it isn't clear why the lifelines should need to be generic. Removing `<...>` definitely fails but it's not clear why.

mdinger:
~~[WIP]~~

* I tried pretty hard to make it read decently (I'm not that great at it though). Hopefully it is.
* ~~I completely glossed over elision and I may just insert a link to the book section where it discusses elision...not sure~~
* I tried to give answers to most of the questions asked on this [users thread](https://users.rust-lang.org/t/what-needs-to-be-explained-about-lifetime-parameters/291) but I'm not sure what they are all asking exactly.
* These examples are mostly simple. I'm not sure what kind of concepts the more complex ones use.

[TODO]

* Need to add `See also` sections
* ~~[This reddit thread](https://www.reddit.com/r/rust/comments/3aahl1/outside_of_closures_what_are_some_other_uses_for/) details a more advanced example which I don't get yet. Maybe I'll look at it again later.~~ [Defer to later]
* ~~Some testcases might be nice.~~ [Defer to later]

[EDIT]
* Added a section on lifetime coercion
* Added a section on elision. It's pretty brief and then defers to the book.

Fixes #154


mdinger:
r? @steveklabnik 

I think this is ready for review. I accidentally messed up my commit history so I went ahead and started new and rebased.

examples/scope/lifetime/elision/input.md
```diff
@@ -0,0 +1,14 @@
+Some lifetime patterns are overwelmingly common and so they may be elide
```
steveklabnik: elided

examples/scope/lifetime/explicit/explicit.rs
```diff
@@ -1,24 +1,34 @@
-struct Book {
-    // `String` is a heap allocated string
-    title: String,
-    author: String,
-    year: i32,
+// `print_refs` takes two references to `i32` which have different
+// lifetimes `'a` and `'b`. These two lifetimes must both outlive
```
steveklabnik: we usually say "be at least as long as"

examples/scope/lifetime/explicit/explicit.rs
```diff
@@ -1,24 +1,34 @@
...
+// the function `print_refs`.
+fn print_refs<'a, 'b>(x: &'a i32, y: &'b i32) {
+    println!("x is {} and y is {}", x, y);
+
+    // Suppose this `drop` worked. Then `print_refs` would outlive
+    // `x` and `four` from `main()` would refer to erased data.
+    // These two things go hand in hand and so both are banned.
+    //
+    //drop(*x);
```
steveklabnik: drop doesn't interact with borrows, exactly, so I'm not sure this is a great way to show this example

examples/scope/lifetime/explicit/input.md
```diff
@@ -1,16 +1,30 @@
...
-to the title field of a Book struct. The most generic function that we could
-write would look like this:
+Since lifetimes *currently* have no explicit type or name associated with them,
+usage will require generics (similar to [closures][anonymity]). Somewhat
+peculiarly, lifetime annotation has a second additional meaning. `foo<'a, 'b>`
+states:
+
+1. `'a` and `'b` will represent names for lifetimes with non-specifiable
+(generic) types.
+2. The lifetime of `foo` may not exceed either lifetimes `'a` or `'b`.
```
steveklabnik: isn't this for `'a: 'b`, not `'a, 'b'`?

examples/scope/lifetime/input.md
```diff
@@ -1,33 +1,13 @@
-The compiler enforces valid borrowing using its borrow checker. To accomplish
-this, it keeps track of the scope of blocks.
+A *lifetime* is a construct the compiler (also called the borrow checker)
+uses to ensure all borrows are valid. Specifically, the lifetime refers to
+the span which starts when the variable is created and ends when the variable
```
steveklabnik: i wonder if 'scope' isn't better than 'span', since it's a mor familliar term

mdinger: Maybe more familiar but also possibly confusable...scope is used in different ways I think...

I don't think vertical scope works though. I was specifically thinking of the vertical span because that's how error messages appear: variable appeared at line a and disappears at line b but this variable ...

Even if it isn't a perfect wording, I thought giving a very specific definition a good idea. [This confusion](https://www.reddit.com/r/rust/comments/2u9opy/lifetime_confusion_on_irc/) in particular seemed to stem from the documentation not stating what a lifetime is exactly; just assuming the reader knew and just wanted to know how to use one. Here I tried to avoid that entirely.

examples/scope/lifetime/explicit/explicit.rs
mdinger: I'm okay with changing this as I don't see a substantial difference but would carefully note this [answer to a users question](https://users.rust-lang.org/t/there-is-a-time-in-your-life-when-you-have-to-understand-lifetimes/1941/2) and then the rephrasing by the original the questioner:

```txt
ogeon: the lifetime of Foo<'a> is at most as long as 'a, but 'a may be longer
killrt rephrase: ... instances of the struct cannot outlive their reference fields
```

I found the rephrase much easier to interpret than the original statement by ogeon. So, while I approve of consistency, if reprasing it makes it easier to interpret, I would pick the rephrasing.

Mainly I just don't want it to sound too complicated I guess.

examples/scope/lifetime/explicit/input.md
mdinger: Are you sure? I've been interpreting `foo<'a, 'b>` to mean `foo <= min('a, 'b)`. Would you state it means `foo <= max('a, 'b)` so if `'a > 'b` then `'b < foo < 'a` would be valid? I don't really remember this being discussed with more than one lifetime...

steveklabnik: oh, i see what you're saying here. i just found the wording confusing. you are correct though...

mdinger: Oh good. I looked at `foo <= max('a, 'b)` and was thinking I really don't think it means that but was trying to give you the benefit of the doubt.

examples/scope/lifetime/elision/input.md
mdinger: Fixed

examples/scope/lifetime/explicit/explicit.rs
mdinger: Fixed.

examples/scope/lifetime/explicit/explicit.rs
mdinger: What about [this](http://is.gd/IyDt6X)?

```rust
struct Empty;

fn bad_consumer<'a>(x: &'a Empty) {
    {
        // Suppose this `move` worked. Then immediately after transferring
        // the data, the new variable would go out scope and be erased.
        // `bad_consumer` would outlive `x` and `empty` from `main()`
        // would refer to erased data. These two things go hand in hand
        // and so both are banned.
        let _eraser = *x;
    }
}

fn main() {
    let empty = Empty;

    bad_consumer(&empty);
}
```

mdinger: Anyway, it's not immediately obvious why `'a` should have to live longer than the function or the `struct` or whatever it is created in. These are attempts to show what would happen if it didn't live longer than the function/`struct`. If that can be understood then it should be easier to see why these rules exist.

examples/scope/lifetime/input.md
steveklabnik: well, borrowing is currently based on lexical scope, so it's certianly the precise term... hrm

mdinger: I'll think about this some. I may be able to word it better. In the mean time, please let me know what you think of my [other suggestion](https://github.com/rust-lang/rust-by-example/pull/607#discussion_r33693987). You may not think it is very different from the original though...

I think that type of example should be conceptually helpful.

examples/scope/lifetime/explicit/explicit.rs
steveklabnik: Hmm. I'm not totally opposed, though the extra `{}`s with nothing outside them feels a bit strange, though I know why they're there.

I wonder if anyone else might have some thoughts here. @huonw ? @alexcrichton ? @nikomatsakis ?

alexcrichton: Aren't move semantics in this case basically orthogonal to lifetimes? I'm not quite sure what this is trying to convey.

mdinger: If a borrow could and did accidentally consume the borrow, would that be bad? Why would it be bad? This is just an attempt to specifically examine that case so it isn't left up to the reader to figure it out on their own.

Before writing this, it wasn't immediately obvious to me why `'a` should have to outlive the function or `struct` or whatever.

Does that reasoning make more sense?

mdinger: The fact that it uses `move` is mainly because I don't know how else this could be examined.

examples/scope/lifetime/input.md
mdinger: @steveklabnik I tweaked this. I still think span is the correct term however, even before it used span as the explanation and then tried to leverage that into why scope is very appropriate and convenient as a lifetime size.

Mainly I tweaked the example to be a little more explicit about detailing lifetime size in terms of scope.

Also, trying to use scope directly doesn't seem to read properly and using both terms more often just seemed to make it more confusing.

steveklabnik: I agree that using both is confusing, but we don't use 'span' anywhere else in any other documentation, and use 'scope' instead: http://doc.rust-lang.org/stable/book/lifetimes.html#thinking-in-scopes

mdinger: @steveklabnik I reworded it to conform. I don't think it reads as well though...

examples/scope/lifetime/explicit/explicit.rs
alexcrichton: Hm I'm not sure I'm quite following just yet, what do you mean "if a borrow could and did accidentally consume the borrow"? I dunno what it means to consume a borrow I guess.

mdinger: @alexcrichton Thanks for following up. I will see if I can explain it more clearly below. Please let me know if this makes it understandable.

I'll try to use the typical book analogy to explain: Supposed I ask you "May I borrow your favourite book?" You might reply "Sure, but be careful. It's my favourite!" Then 2 weeks later when you asked for it back I told you "Sorry, you know how stuff happens? I accidentally burned it. It's gone. Sorry."  You'd rightfully be angry and perhaps demand something in reparation for it because I failed to meet the rules borrowing stipulates: I'm supposed to return it (preferably in the condition I received it).

Similar to the analogy, suppose I could create a function, perhaps like `fn borrow<'a>(b: &'a book)`, where the function cannot require that the `book` still exist when it exits. Maybe the type system doesn't ban such things. Suppose `borrow()` passes the book to a function called `burn()`, destroying the book. Then `borrow()` exits. This function would, similar to the analogy, violate the borrowing agreement.

How do lifetimes factor in? If I didn't understand lifetimes and was told `foo<'a>(...)` states `'a` must outlive `foo`, I might like to know why. Perhaps it's not obvious why. The obvious thing to do then is examine an example where `foo()` outlives `'a` (the opposite of what is supposed to happen) to see if anything bad happens.

These two analogies and the examples in the PR are attempts to make examples that would violate a functions' lifetime requirements to show what would happen. This would hopefully make the lifetime rules and the reasons they exist more understandable.

Note: The fact that I have to go into this detail suggests to me that the examples current form is clearly far from ideal. Readers shouldn't have to ask for extensive detail to understand an example. So this example seems clearly inadequate.

mdinger: @alexcrichton @steveklabnik I removed this and added some extra text explanation instead.

mdinger:
@steveklabnik r?

I tweaked a bunch of stuff but a lot of it is still the same. I'd happily squash to 1 or 2 commits if you like it as is but want fewer commits.

tomjakubowski:
cc #613 (I see that this pull request resolves that by adding a section on lifetime bounds)

examples/scope/lifetime/explicit/input.md
```diff
@@ -1,16 +1,34 @@
...
+how long references should be valid. Failure to annotate lifetimes[^1] is akin
+to banning the borrow checker from validating borrows and so accordingly,
+annotation is mandatory.
 
-Let's illustrate with an example: we want a function that returns a reference
-to the title field of a Book struct. The most generic function that we could
-write would look like this:
+Since lifetimes *currently* have no explicit type or name associated with them,
+usage will require generics (similar to [closures][anonymity]). Somewhat
+peculiarly, lifetime annotation has a second additional meaning. `foo<'a, 'b>`
```
nikomatsakis: This is not really how I think of it. I presume `foo<'a,'b>` refers to a type? In that case, `foo<'a, 'b>` is just as simple as "`foo`, where `'a` and `'b` are substituted for the lifetime parameters within the definition", just as `Vec<i32>` is "`Vec`, where `i32` is substituted for the type parameter in the definition". Now, it is also true that if you have `foo<'a ,'b>`, this type can only be referenced where `'a` and `'b` are in scope, so that implies that the structure is limited by the minimum of `'a` and `'b`.

mdinger: Yes, `foo<'a, 'b>`  is meant to represent a type such as a `struct` or a function or something else.

That makes sense and I think I can tweak it to use a different approach. I've been generally explaining `struct Num(i32)` declarations and usage independently. This has worked okay so far. A better approach might be to introduce `struct Ref<'a>(&'a i32)` via a "ref is subordinate to owner" approach (the ref must come from somewhere but I've kinda ignored that part).

I'll try it and see if I can improve it.

mdinger: @nikomatsakis I tweaked it to be more like you suggested.

examples/scope/lifetime/explicit/explicit.rs
```diff
@@ -1,24 +1,37 @@
...
 
-    println!("I just read {}", title);
+    // Attempting to use the lifetime `'a` as an explicit type
+    // annotation inside the function will fail because the
+    // lifetime of `&_x` is smaller than `y` has. A small lifetime
+    // cannot be coerced into a larger one.
+    //
+    //let y: &'a i32 = &_x;
+    // ERROR: `_x` does not live long enough
 }
```
mdinger: @nikomatsakis This function above was slightly problematic with your suggested explanation since you cannot rely on input borrowing being returned.

nikomatsakis: To be honest, this function makes no sense at all -- a random lifetime parameter that is not used anywhere is not particularly useful or meaningful. (Really, it ought to be illegal.)

That said, I don't quite follow how this example is in conflict with the explanation I suggested. I'm not even 100% sure what explanation you're referring to :). I thought I was explaining how to interpret a type like `Foo<'a>`, which does not appear in this example, right?

But in any case, the existing text is correct: whatever scope `'a` corresponds to (and, given that `'a` is completely unused, the caller can pick whatever scope they can name), it will be something larger than the current function, and `_x` is SMALLER than the current function, so this code is in error.

mdinger: Ah, [this explanation](https://github.com/rust-lang/rust-by-example/pull/607#discussion_r34061188). Anyway, the rules for `Foo<'a>` as a `struct` or `foo<'a>` as a function are basically the same. Rather than try to explain the same thing twice once for `structs` and once for functions, I tried to use `foo<'a, 'b>` as a stand-in for both. Maybe it wasn't successful...

Good. If the explanation is okay then I'm okay with it. As for this example being random...it's not really. It's due to trying the obvious thing following an unhelpful error message. Consider this:

```rust
fn test() { let x: &'a i32 = &7; }
```

Error:
```cmd
<anon>:1:21: 1:23 error: use of undeclared lifetime name `'a` [E0261]
<anon>:1 fn test() { let x: &'a i32 = &7; }
```

Oh, it needs declaring. Let's try:

```rust
fn test<'a>() { let x: &'a i32 = &7; }
```

Error:
```cmd
<anon>:1:35: 1:36 error: borrowed value does not live long enough
<anon>:1 fn test<'a>() { let x: &'a i32 = &7; }
```

Hmm...is manual annotation of lifetimes possible?

mdinger:
Made another slight tweak based on @nikomatsakis's comment. I'm not sure if I should keep asking for r? after fixing things or what. Or who I should be asking...

@nikomatsakis @alexcrichton Thanks for critiquing so far. Both your comments/critiques have been helpful. Especially @nikomatsakis.

mdinger:
Rebased

mdinger:
If someone could list an ETA for review by Friday that would be helpful to me. Stuff doesn't happen much over the weekends usually.

mdinger:
It has been a week. Review would still be nice. Maybe some things still need changes but I'm not sure how I should change them.

mdinger:
Still looking for review. It's been another week. Just trying to keep this from being forgotten.

steveklabnik:
with the run up to rustcamp, i am very behind. Sorry :(

steveklabnik:
@mdinger hoping to get to this sometime today, we'll see with the release

mdinger:
That would be so great! I've pretty much given up on someone reviewing this.

steveklabnik:
Actually finally reviewing this. built it locally, reading now.

examples/scope/lifetime/explicit/input.md
```diff
@@ -1,16 +1,32 @@
-When writing functions that return references, lifetimes must be explicitly
-annotated. These functions are generic and we must tell the compiler what is
-the relationship between the lifetimes of the objects that appear in the
-arguments and the output.
+The borrow checker utilizes explicit lifetime annotation to reason about
```
steveklabnik: I would say 'annotations" here, which is a small difference, but makes it way easier for me to parse.

examples/scope/lifetime/explicit/explicit.rs
```diff
@@ -1,24 +1,41 @@
...
-}
+// This example uses `main()` at the top so it can be read top to
+// bottom. `main()` is usually at the bottom.
+fn main() {
+    // Create variables which will be borrowed.
+    let (four, nine) = (4, 9);
+
+    // Borrows (`&`) of both variables are passed into the function.
+    // If a borrow is truly a borrow, the variable must be returned,
+    // otherwise a borrow would be transferring ownership. This means
```
steveklabnik: I would make this sentence

> When you pass a function a borrow as an argument, the borrow ends at the end of the function. The alternative to passing a borrow would be trasnferring ownership.

examples/scope/lifetime/explicit/explicit.rs
```diff
@@ -1,24 +1,41 @@
...
+    // other words, *Rule 2* must be true.
+    print_refs(&four, &nine);
 
-fn get_title<'a>(book: &'a Book) -> &'a str {
-    &book.title
+    // Uses a lifetime without any input. The lifetime sizedness will
+    // be chosen by the caller (which is here) from available lifetimes
+    // (scopes). Any lifetime chosen will then be larger than that of
+    // the function.
+    failed_borrow();
```
steveklabnik: I'm not sure about this, I'm finding it pretty confusing.

examples/scope/lifetime/explicit/explicit.rs
```diff
@@ -1,24 +1,41 @@
...
+// There are no references to force `'a` to be larger than the
+// function yet, `'a` remains larger than the function.
+fn failed_borrow<'a>() {
+    let _x = 12;
 
-    println!("I just read {}", title);
+    // Attempting to use the lifetime `'a` as an explicit type
+    // annotation inside the function will fail because the
+    // lifetime of `&_x` is smaller than `y` has. A small lifetime
+    // cannot be coerced into a larger one.
```
steveklabnik: I think this makes the example more clear. Maybe we can re-word the description above. I think the problem is that it says "uses a lifetime" but the syntax isn't anywhere.

examples/scope/lifetime/fn/input.md
```diff
@@ -1,4 +1,15 @@
-Explicit lifetimes are necessary when functions return references. Our case
-study will be returning a reference to one of the fields of a struct.
+Functions with lifetimes have a few different valid forms. Ignoring
+[elision][elision] for the time being, the rules for function parameters are:
+
+* any reference *must* have an annotated lifetime.
+* any reference being returned *must* have the same lifetime as an input.
```
steveklabnik: (or `'static`)

examples/scope/lifetime/fn/fn.rs
```diff
@@ -1,50 +1,45 @@
-#[derive(Debug)]
-struct Triplet {
-    one: i32,
-    two: i32,
-    three: i32,
+// One input reference with lifetime `'a` which must live
+// longer than the function.
```
steveklabnik: at least as long

examples/scope/lifetime/fn/fn.rs
```diff
@@ -1,50 +1,45 @@
...
+    *x += 1;
 }
 
-fn main() {
-    let mut triplet = Triplet { one: 1, two: 2, three: 3 };
-
-    println!("Before: {:?}", triplet);
+// Multiple elements with different lifetimes. This would
+// be equally acceptable if both references had the same
+// lifetime `'a`.
```
steveklabnik: I would phrase this more like "In this case, it would be fine for both to have the same lifetime `'a`, but in more complex cases, you may want two different lifetimes."

examples/scope/lifetime/fn/fn.rs
mdinger: I can change it but I'd note some people may find certain phrases easier to parse than others. Changing the wording around might make it so they don't understand one explanation but suddenly another slightly different one makes it click.

I think `at least as long` is slightly more abstract than live longer too.

examples/scope/lifetime/static_lifetime/input.md
```diff
@@ -4,6 +4,9 @@ program. There are two ways to make a variable with `'static` lifetime:
 * Make a `"string"` literal which has type: `&'static str`.
 * Make a constant with the `static` declaration.
```
steveklabnik: I know this isn't part of your diff, but while we're here, calling this a 'constant' is bad, because `const` is different than `static`. How about

> use a `static` item

with a link to a section on statics? do we have one of those yet?

steveklabnik:
Whew! Done. I think this is definitely an improvement, just some nits.

mdinger:
Okay cool. I'll try to get the changes fixed quickly. I'd like this to land if possible.

examples/scope/lifetime/fn/fn.rs
steveklabnik: The thing is that "must live longer" and "must live at least as long" are two different things, worded this way is inaccurate.

examples/scope/lifetime/explicit/explicit.rs
mdinger: I don't like this change. As written, I'm trying to leverage the idea that people understand that borrows must always come back. If they don't come back then they can't be borrows; they must be transfers. It's the only option.

With your suggestion, I would immediately ask "why must they be returned"? I wouldn't necessarily see any reason they must be returned other than that is what the docs say.

examples/scope/lifetime/fn/fn.rs
mdinger: Yeah, I know and the change should be made but it's still easier to understand as written. Especially if you don't already understand it.

examples/scope/lifetime/explicit/explicit.rs
mdinger: Any suggestions which would make this clearer?

examples/scope/lifetime/static_lifetime/input.md
mdinger: Well, they are both listed under constants. See the link at the bottom I think.

examples/scope/lifetime/explicit/explicit.rs
mdinger: I thought of something

mdinger:
Fixed

examples/scope/lifetime/explicit/explicit.rs
steveklabnik: okay cool, give it a shot :)

examples/scope/lifetime/static_lifetime/input.md
steveklabnik: ahh yeah this is fine, let's just keep it consistent then

examples/scope/lifetime/explicit/explicit.rs
steveklabnik: The issue is that you don't return, like

    fn foo<'a>(&'a i32)

Does not return anything, but is just fine.

examples/scope/lifetime/explicit/explicit.rs
mdinger: Oh, I already changed it. See the new commit: https://github.com/mdinger/rust-by-example/commit/0832d4eb84859c41599ef59ac5a9277e4c2815e7 for what is different

examples/scope/lifetime/explicit/explicit.rs
mdinger: I'm not talking about explicit returning. I'm talking about doing something which would preclude the reference from ever being able to return to it's owner. For example, if I tried this:

```rust
// Would this be the lifetime drawing?
fn foo<'a>(a: &'a NonCopyType) { // ---|
//                                     |
    let var = *a; // ------------------|

    ...    

    bar(var);
}
```

Why would it fail? Well, it would force a move but a move can't be done with a borrow. Why not? Because that would make it something other than a borrow. I consider this related to lifetime size. `foo()` above has lifetime size which must be at least as large as the function.

Why must this lifetime size be larger than the function? Because borrows must always be returnable. The borrow being returned at the end is just a minor point you could learn after this is realized.

examples/scope/lifetime/explicit/explicit.rs
steveklabnik: ah! i wonder why github is still showing me this then

mdinger: Probably because the fix is in a different section where you stated you find it confusing. Here was where you said you were less confused.

examples/scope/lifetime/explicit/explicit.rs
mdinger: @steveklabnik Should I just find you on IRC tomorrow to try and hash out what to do about this section?

steveklabnik: Maybe yeah. I"m not sure how much I'll be on. Still thinking about this.

mdinger:
rebased

steveklabnik:
I'm still not 100% sure about that one example, but this has been sitting around forever, and I want to get it landed.

Thanks again for being patient @mdinger :heart:

mdinger:
woooo!

@steveklabnik @nikomatsakis @alexcrichton Thanks for reviewing even if only for a little while!

PR #608, opened 2015-06-30
[Fn|Closures|OutputParameters] Update explanation

ner0x652:
I consider that it is important for the reader to know to which version of Rust Lang does refer the author.
** Version 1.1.0 was introduced, but I am not sure it is correct. If I am wrong, excuse me and please indicate the correct one. **


steveklabnik:
Generally speaking, this is why I don't like docs that read like this. `FnBox` isn't in 1.1, it's not in any Rust release yet, so this isn't accurate, yet your sentiment is accurate too.

steveklabnik:
That said, I'm going to merge, because it doesn't make it any worse. Thanks.

ner0x652:
Thank you for clarification! :)

Issue #609, opened 2015-07-01
Fix typo


kteza1:
Change 'eat' to 'consume' in the comment

https://github.com/rust-lang/rust-by-example/blob/master/examples/fn/closures/capture/capture.rs#L48-Lundefined

steveklabnik:
fixed by https://github.com/rust-lang/rust-by-example/pull/610

PR #610, opened 2015-07-01
Typo: eat to consume

mdinger:


mdinger:
Fixes https://github.com/rust-lang/rust-by-example/issues/609

PR #612, opened 2015-07-03
Added tuple destruct

kteza1:
Tuple destructuring is very useful. Is it discussed somewhere else? If so, why not here?


steveklabnik:
Thank you!

mdinger:
FYI, it's also [here](http://rustbyexample.com/flow_control/match/destructuring/destructure_tuple.html)

PR #614, opened 2015-07-06
Correct explanation of where clauses

tomjakubowski:
Where clauses don't apply bounds to arbitrary expressions, they apply bounds to arbitrary types (`Option<T>`). The other form is limited in that it can only apply bounds to a type parameter (`T`).


steveklabnik:
Thanks!

PR #617, opened 2015-07-10
Notify me of build changes

mdinger:
Seems this is only necessary if you don't have push privileges. Don't know how to test if it works without causing breakage to test it.


mdinger:
@steveklabnik Thanks for the tip!

steveklabnik:
Okay! Let's see how it goes :)

PR #618, opened 2015-07-13
lifetimes/structs expand unclear abbreviation

MatejLach:
The `'s` in `Let's` gets rendered like a lifetime, which could be confusing for some, especially in this particular chapter.


steveklabnik:
Ha! Nice catch :)

Issue #619, opened 2015-07-20
Comments should be better styled


eccegordo:
There is no font color distinction between commented text and code text. This makes it really hard to distinguish code from comment and hampers overall readability, IMHO. Even setting the comment color to 50% would help a lot.

![struct_visibility___rust_by_example](https://cloud.githubusercontent.com/assets/497659/8771114/b66c6bb8-2e71-11e5-9608-4df518452ffd.png)

mdinger:
This is recent. I'm not sure when it started but it's weird.

mdinger:
Seems to have changed from when this was updated: https://github.com/ajaxorg/ace/pull/2586

Playpen works properly so rbe is probably just doing something weird.

mdinger:
@bluss you monitored that PR going in. The [ace kitchen sink](http://ace.c9.io/build/kitchen-sink.html) shows the same issue so it seems to be ace which is the problem. I'm thrown by playpen working properly though. Any idea?

mdinger:
That PR merged July 11: [playpen](https://github.com/rust-lang/rust-playpen/commits/master) doesn't show anything obvious which would be related.

mdinger:
I think I see the problem. The operators regex is falsely stating `//` is an operator. I think I can fix it.

mdinger:
I fixed this in https://github.com/ajaxorg/ace/pull/2600 so this should be fixed the next time is updated.

PR #620, opened 2015-07-23
Use local highlighting file until ace-builds updates again

mdinger:
This pulls in the fix from https://github.com/ajaxorg/ace/pull/2600.

I changed the 2 regex parts from the local already minimized file. Since it's minimized the fix won't showup in a diff well. I'm not sure how I could make this PR clear either way.

This is a temporary fix until https://github.com/ajaxorg/ace-builds pushes a new update.

Fixes https://github.com/rust-lang/rust-by-example/issues/619


steveklabnik:
Sounds good to me, thanks!

PR #622, opened 2015-07-30
Fix typo in comment on Iterator::find

chriskrycho:
The comment appears to have been copy-and-pasted from the `Iterator::any` example on the previous page, and the comment still had `and` instead of `find` in its description of the method.


steveklabnik:
Thanks!

steveklabnik:
hehe, needed to click 'merge' instead of using bors.

PR #623, opened 2015-07-31
Add same scope binding shadowing

blaztinn:
Add an example of shadowing a variable binding in the same scope.


steveklabnik:
Thanks :)

PR #628, opened 2015-08-06
Fix a typo and update some code samples

chriskrycho:
- Typo: add an apostrophe to a possessive that was missing it.
- Code samples: switch from playpen to built-in sample type in all the file I/O examples, since their being editable is at best irrelevant and at worst confusing. Update the verbiage around them accordingly, no longer referencing the inability to run the samples in the playpen.

I took this route on these code samples to bring it in line with the FFI examples, which likewise do not use the playpen.


steveklabnik:
Thanks!

PR #629, opened 2015-08-09
Add simpler example to scope/borrow/ref

mdinger:
Fixes https://github.com/rust-lang/rust-by-example/issues/390

cc @pornel

I don't have a strong opinion about this but this seems to be the best place to put this and it seems an okay fit.


pornel:
:+1:

PR #635, opened 2015-08-15
Add enum variants to cover all cases

xunien:
Add variants, not sure if it's too complicated because nothing talk about borrowing ?

### DIFF
--- a/examples/custom_types/enum/enum.rs
+++ b/examples/custom_types/enum/enum.rs
@@ -36,10 +36,15 @@ fn inspect(p: Person) {
 
 fn main() {
     let person = Person::Height(18);
+    let danny  = Person::Weight(10);
     // `to_owned()` creates an owned `String` from a string slice.
     let dave   = Person::Info { name: "Dave".to_owned(), height: 72 };
-    // ^ TODO: Try changing these to a different variants.
+    let john   = Person::Fat;
+    let larry  = Person::Skinny;
 
     inspect(person);
+    inspect(danny);
     inspect(dave);
+    inspect(john);
+    inspect(larry);
 }


examples/custom_types/enum/enum.rs
```diff
@@ -18,28 +18,31 @@ enum Person {
 
 // A function which takes a `Person` enum as an argument and
 // returns nothing.
-fn inspect(p: Person) {
+fn inspect(p: &Person) {
     // Usage of an `enum` must cover all cases (irrefutable)
     // so a `match` is used to branch over it.
-    match p {
+    match *p {
```
mdinger: This won't work as written. `Person` isn't `Copy` so `*p` will force a move from a borrow which will not be allowed. You'd have to do

```rust
match p {
    &Person::Skinny => ...,
...
}
```

examples/custom_types/enum/enum.rs
```diff
@@ -18,28 +18,31 @@ enum Person {
...
         Person::Skinny    => println!("Is skinny!"),
         Person::Fat       => println!("Is fat!"),
         // Destructure `i` from inside the `enum`.
         Person::Height(i) => println!("Has a height of {}.", i),
         Person::Weight(i) => println!("Has a weight of {}.", i),
         // Destructure `Info` into `name` and `height`.
-        Person::Info { name, height } => {
-            println!("{} is {} tall!", name, height);
+        Person::Info { ref name, ref height } => {
```
mdinger: Don't need a `&Person` to do a `ref`.

mdinger:
Anyway, I'm not sure what this is trying to achieve. It complicates the example with `Copy`, `Clone`, and/or borrowing, and `iter()`. If you're looking for borrowing, it's discussed more in [match/destrucuring/pointers](http://rustbyexample.com/flow_control/match/destructuring/destructure_pointers.html), in [scope/borrowing](http://rustbyexample.com/scope/borrow.html), and in [scope/lifetimes](http://rustbyexample.com/scope/lifetime.html) (which was very recently changed).

xunien:
Yes to complicate. Sorry for this useless PR.

mdinger:
Just to clarify, I wasn't trying to belittle you over this PR. I just didn't see what the intent was or what problem you were trying to solve in light of my previous comment.

xunien:
I just want to add some variants to fix that :
    `// ^ TODO: Try changing these to a different variants.`

Nervermind, my PR was unclear ;) I will make a simpler.

mdinger:
This is okay and clearer than before.

steveklabnik:
Thanks!

PR #636, opened 2015-08-17
Move RBE to stable

steveklabnik:
This commit moves everything to stable Rust.

In it, we lose a few things:

* FnBox examples
* Associated Constants
* Benchmark tests
* PathExt stuff
* SIMD
* Inline Assembly

These are varying degrees of unfortunate. A better build system
might be able to do nightly vs stable builds for the relevant bits
in the future.


src/example.rs
```diff
@@ -37,7 +37,7 @@ impl Example {
             None => 1,
             Some(ref children) => 1 + children.iter()
                                               .map(|c| c.count())
-                                              .sum::<usize>(),
+                                              .fold(0, |sum, i| sum + i),
```
mdinger: That didn't occur to me. Glad you thought of it.

examples/fn/closures/output_parameters/input.md
```diff
@@ -1,30 +0,0 @@
...
-### See also:
-
-[Boxing][box], [`Fn`][fn], [`FnMut`][fnmut], [`FnBox`][fnbox], and
-[Generics][generics]
-
-[box]: /std/box.html
-[fn]: http://doc.rust-lang.org/std/ops/trait.Fn.html
-[fnmut]: http://doc.rust-lang.org/std/ops/trait.FnMut.html
-[fnbox]: http://doc.rust-lang.org/std/boxed/trait.FnBox.html
-[generics]: /generics.html
```
mdinger: Any reason this section was removed? You could just remove `FnBox` and add a comment stating `FnOnce` doesn't yet have a analog on stable.

steveklabnik: because I read `output_parameters.rs` wrong, and thought it was all about FnBox. i should restore this and the rest of that.

mdinger:
Missed the changes to `structure.json`. That's what those weird error messages are. It won't build without them.

examples/fn/closures/output_parameters/input.md
```diff
@@ -8,8 +8,6 @@ The valid types for returns are slightly different than before:
 
 * `Fn`: normal
 * `FnMut`: normal
-* `FnBox`: equivalent to `FnOnce` but specialized for this application
-because `FnOnce` currently(version 1.1.0) interacts badly with the type system.
```
mdinger: You still may want to comment on it though. People will probably expect the `FnOnce` analog and so might just think it missing is a typo.

mdinger:
Aside from that, I'm pretty okay with this. I would think the biggest loss here is losing slice patterns which are useful to know about (I was certain `step_by()` was used somewhere but I can't find it anywhere...so I guess I was mistaken). SIMD is being changed through that new RFC. bench and unsafe have better explanations in other locations such as the book and the rustonomicon.

This seems fine to me.

examples/fn/closures/output_parameters/input.md
steveklabnik: :+1:

steveklabnik:
@mdinger cool, glad we're on the same page.

mdinger:
Nice. +1 from me.

alexcrichton:
:confetti_ball: 

Nice work @steveklabnik!

sinistersnare:
Can we add a section of RBE: "Unstable of Rust", and recreate these sections?

steveklabnik:
@sinistersnare not without significant work on the build process.

mdinger:
Not just the build process. The playpen currently all uses the same environment and would have to be conditionally enabled for just those sections. Currently it doesn't pay attention to which example is being run, just that an example is being run.

steveklabnik:
And to be clear, I do want that work to happen at some point. It's just not simple.

PR #648, opened 2015-10-01
Fix playpen

mdinger:
Using ace-min-noconflict causes ace to not recognize the editor. I'm not sure if a bad build was uploaded or what but switching to min fixed it.


mdinger:
Oh, ace updated so this grabs the latest ace now. That means that local highlighting thing from a while back can be reverted.

mdinger:
I think the minification must have caused errors in the one ace set. Either that or a bad upload. I've notified upstream but haven't heard back yet so it's still unknown.

steveklabnik:
Sorry this took a while, GitHub mobile wouldn't let me merge without Travis, and then I went to bed.

PR #650, opened 2015-10-08
print_display/testcase_list: fix issue with zero-length vector

vthib:
The implementation given to implement fmt::Display with a Vec
works fine, as long as the vector is not empty. If it is,
vec[len-1] is accessed, leading to a run-time panic.

The issue is that a fmt::Result value must be returned in the end,
and as the example did not used any brackets to surround the printed
values, empty vector would have meant nothing to print. A "Ok(())"
could be returned, but this isn't elegant or best for an example of
pretty Rust code.

'[' and ']' are therefore added around the printed values, and the
code is reworked to print a comma before every element that is not
the first one.


steveklabnik:
Nice catch! Thanks.

PR #651, opened 2015-10-13
std_misc/arg/matching: fix match_args.rs to match input.md

ia0:
This patch fixes:

    $ ./match_args Rust
    thread '<main>' panicked at 'called `Result::unwrap()` on an `Err` value: ParseIntError { kind: InvalidDigit }', ../src/libcore/result.rs:732

to:

    $ ./match_args Rust
    This is not the answer.

which corresponds to the output expected by `input.md`.


steveklabnik:
Ah, thanks!

PR #655, opened 2015-10-31
Correct lifetime_coercion example

### LINKED ISSUE #654: Lifetime coercion example coneptually incorrect
The [example](http://rustbyexample.com/scope/lifetime/lifetime_coercion.html) should be:

```rust
fn coerce_first<'a: 'b, 'b>(x: &'a i32, _: &'b i32) -> &'b i32 {
    x
}

fn main() {
    let x = 800;

    // let coerce_test;
    let borrow_big = &x;
    {
        let y = 8; // <<< should be here; otherwise 'a and 'b would be equals
        let borrow_small = &y;

        let coerced = coerce_first(borrow_big, borrow_small);
        // let coerce_test = coerced; // this should not work, even though "safe"
        println!("`coerced` is {}", coerced);
    }
}
```

critiqjo:
Closes #654

### DIFF
--- a/examples/scope/lifetime/lifetime_coercion/coercion.rs
+++ b/examples/scope/lifetime/lifetime_coercion/coercion.rs
@@ -1,21 +1,25 @@
 // `x` is a reference with lifetime `'a` which is larger than `'b`.
-// Both `'a` and `'b` are larger than `coerce_first`. Since `'a` is
-// larger than `'b`, it may be coerced.
 fn coerce_first<'a: 'b, 'b>(x: &'a i32, _: &'b i32) -> &'b i32 {
-    x
+    x // Since `'a` is larger, it may be coerced to match `'b`
 }
 
 fn main() {
     let x = 800;
-    let y = 8;
 
     let borrow_big = &x;
+    //let coerce_test;
+
     {
-        // This reference is inside a scope and therefore is
-        // smaller than the other borrow.
+        let y = 8;
+        // Since y is inside a scope, this reference has smaller
+        // lifetime than borrow_big.
         let borrow_small = &y;
 
         let coerced = coerce_first(borrow_big, borrow_small);
+
+        // Since the returned reference has a lifetime of y,
+        //coerce_test = coerced; // this will not work
+
         println!("`coerced` is {}", coerced);
     }
 }


examples/scope/lifetime/lifetime_coercion/coercion.rs
```diff
@@ -1,21 +1,25 @@
...
 fn coerce_first<'a: 'b, 'b>(x: &'a i32, _: &'b i32) -> &'b i32 {
-    x
+    x // Since `'a` is larger, it may be coerced to match `'b`
 }
 
 fn main() {
     let x = 800;
-    let y = 8;
 
     let borrow_big = &x;
```
critiqjo: May be this line should also be moved inside the scope? Just to strengthen the point.

steveklabnik:
I just woke up, so maybe I'm missing something, but it's not clear to me why the original example is incorrect.

critiqjo:
Because both references have the same lifetime. Because [this](http://is.gd/GjLO81) works! The example *should* illustrate an example where `'a` > `'b`.

mdinger:
Eh, I thought the lifetime of a borrow is sized according to the scope it is declared in. The lifetime of the actual references may be larger but I didn't think that limited the borrow lifetime from being smaller. Having said that, I don't know why your counter example works.

@steveklabnik Can you explain this? If not, I'd probably cc someone for a more authoritative opinion.

critiqjo:
@mdinger I couldn't come up with a better example to make it more clear, sorry...

Anyway, to demonstrate coercion, I think [this](http://is.gd/6eXjVH) would be enough. Relevant line:
```rust
fn coerce_first<'a>(x: &'a i32, _: &'a i32) -> &'a i32 {
```

The larger lifetime(s) will get coerced to the smallest one during call. The original example also illustrated specification of bounds, but that could be an added complexity for the reader...

mdinger:
That works but this changes how lifetime sizedness is specified. If that is to be done, other sections such as the [lifetimes intro](http://rustbyexample.com/scope/lifetime.html) also need to be updated.

Also, your example drops the `a: b`. It is useful to use it at least twice at least for the reinforcement it gives when reading it twice.

critiqjo:
>the lifetimes intro also need to be updated.

A quote from it:

>Borrowing (via & for example) creates new lifetimes. A borrow is valid as long as the borrow ends before (inside) the lender is destroyed.

That is a confusing statement.

Every variable has a lifetime associated with it, even reference variables (since you can have `&&obj`, each `&` has an associated lifetime). So creation of a variable creates a lifetime. But it's better to phrase it simply as "A borrow is valid until it goes out of scope." (since (immutable) references follow copy semantics) (the lender cannot be destroyed during this time)

mdinger:
Your rephrase is very direct and proper but I wonder if it is enough. People get errors about lifetimes that confuse them when they try weird things. For example, that was an attempt to highlight that a borrow entirely depends on what it was borrowed from.

You want people to recognize that otherwise they might find many of the rules just weird and confusing.

critiqjo:
You're right! So a borrow should be properly defined. My attempt:

"A borrow gives you a reference. A reference has an associated lifetime (parameter), which is the lifetime of the object it borrows. As long as a reference lives (until it goes out of scope), the original object cannot be moved/destroyed."

(Additional constraints may be detailed when describing "mutable vs immutable" borrows)

mdinger:
So, I'm trying to be as fair as I can and I realize I'm biased because I wrote the original but I'm trying be reasonable.

This is really tricky. I don't think you can explain lifetimes by talking about lifetimes. The beginning of the page talks about lifetimes as a very vague concept which can be visualized as a vertical distance and then sorta leaves it to hopefully become more clear with context.

For example, [the gist in this thread](https://www.reddit.com/r/rust/comments/2u9opy/lifetime_confusion_on_irc/), there was a huge thread where someone was trying to understand what a lifetime is but no one explains what a lifetime is. As I remember (didn't reread it), they kept trying to explain lifetimes in terms of lifetimes (or how long it lives). That's not a good methodology.

So the original tries to piggyback on the meaning borrowing instead of lifetimes and hopefully eventually both can be seen as analogies to their real life counterparts.

Also, that ends up being twice as long. For how confusable the original is, it is somewhat compact.

Again, I realize I'm probably being defensive but I tried to be fair and I hope the critique is warranted and reasonable.

critiqjo:
(Just woke up.)
Lifetimes are always associated with the scope of the variable. Whether it's moved or dropped is an independent matter which does not really affect the lifetime. Because:
```rust
{
    let x = Foo; // a struct
    let y = &x;
    moving_call(x);
}
```
Where the `moving_call` occurs does not matter, as long as there is a borrow (`let y = &x;`) in the same scope as a move, the code is invalid.

So saying the lifetime is the vertical distance, is not 100% accurate, even though thinking that way is ok.

>I don't think you can explain lifetimes by talking about lifetimes.

May be you can define lifetime in terms of scope, and borrows in terms of lifetimes. Right after variable bindings are explained, may be mention that "Every variable (not the object) is said to have a 'lifetime', which is tied to the scope it is in. When it goes out of scope, the lifetime is said to have expired, and the object bound to it (if any) will be dropped."

critiqjo:
Please ignore the mail, and read directly from Github.

critiqjo:
Also see a [seems-like-coercion-but-not example](http://is.gd/kcEtMN)!

mdinger:
By the way, I'm really glad someone is critiquing this section carefully as I think you're the first to critique it at all basically since it merged. I hope you continue and critique at least the entire lifetime section if you find anything amiss.

---

Honestly I'd avoid mentioning lifetimes if possible in bindings because it is an early section and understanding lifetimes is difficult and problematic enough that I would defer it until later.

Isn't what the first paragraph in [lifetimes](http://rustbyexample.com/scope/lifetime.html) basically similar to your final paragraph? Even if vertical distance isn't precise, it seems okay for the readers first mention of lifetimes. Having something precise the reader latch onto allows their understanding to be refined through the following sections. Lacking anything precise to grasp, the reader is likely as not to make up something entirely odd and perhaps irrelevant and be mystified when follow up explanations don't gel with their mental model.

I think trying to give a very specific, understandable analogy should be very helpful. I'm re-asserting this because most explanations I've seen avoid this, possibly because a concrete understandable definition is so hard to come by (without relying on circular logic: a lifetime is a lifetime).

critiqjo:
>I hope you continue and critique at least the entire lifetime section if you find anything amiss.

It just doesn't seem to be having an effect. :disappointed: Are you convinced that the original example (the reason for this PR) should be changed -- that `y` should be moved inside the smaller scope?

>Also, your example drops the `'a: 'b`. It is useful to use it at least twice at least for the reinforcement it gives when reading it twice.

But, here's a confusing modification: http://is.gd/mvSUdF which works!
Relevant line:
```rust
let coerced = coerce_first(borrow_small, borrow_big); // swapped
```
Even though `'a: 'b`, lifetime of `borrow_big` gets coerced to match that condition at the time of call itself. So if it was up to me, I would keep the example simple (with just `'a`), and state: "even though the function seems to need 2 references of equal lifetimes, the bigger reference will get coerced to have smaller lifetime at the time of call (similar to an implicit typecasting)".

mdinger:
I think your argument is fairly persuasive but to change that example as you suggest would disagree with [this example](http://rustbyexample.com/scope/lifetime.html) at the least which states that borrows in inner scopes get smaller lifetimes than in larger scopes.

Even if the first lifetime page's interpretation can still be taken as correct, it'd be helpful to either provide an additional example clarifying the meaning of those lifetime sizes, or to extend the example to be more precise.

So, if your fix is to be accepted, I don't think it is large enough. More of the examples would need to be tweaked to make it plausible. Also, as I'm not an authority on lifetime interpretation though, I'd have to defer to someone more knowledgeable unless @steveklabnik just up and decided you're correct.

@nikomatsakis Could you chime in on if his interpretation is correct? Or recommend someone to chime in?

critiqjo:
@nikomatsakis TLDR; Why the original example is insufficient to explain lifetime coercion:
>Because both references have the same lifetime. Because [this](http://is.gd/GjLO81) works! The example *should* illustrate an example where `'a` > `'b`. [This](http://is.gd/594V3F) doesn't work.

critiqjo:
I believe Matsakis is very busy, making Rust awesome! Let's try \#rust IRC...

steveklabnik:
I _believe_ that the issue here is reborrowing, but am not entirely sure.

nikomatsakis:
I've tried to read the debate through but I guess not having read all the examples it's hard for me to tell which is the clearest explanation. Let me try to explain how the system works and clarify a few things a bit and you all tell me what you think that means for the example. :)

First, I usually pronounce `'a: 'b` as `"`'a` outlives `'b`" but it really means "`'a` >= `'b`". That is, the scope `'a` is **at least as large** as `'b`. In other words, it is fine for `'a` and `'b` to be the same scope.

Second, you can see that I used the word scope and not lifetime. I'm not sure that scope is the perfect word, but I've come to somewhat regret the choice of the word lifetime, because I think many people wind up confused. In particular, the lifetime of a reference (the `'a` in a `&'a T` type) refers to the smallest scope that contains *all the **uses** of the reference*. There is a separate concept that is the lifetime of the underlying, borrowed resource (the one that the reference refers to). Naturally the scope of the reference cannot be **larger** than the lifetime of the underlying resource.

I think this distinction is important to the examples in play here. That is, I think that the **scopes** of the borrows in the example both before and after this change are the same. This is because the borrow does not escape the inner block (in the new example, there is some commented out text that, were it uncommented, would cause the borrow to escape the inner block). What IS difference however is the lifetimes of the borrowed variables: before, both `x` and `y` had the same lifetime (the outer block, but see * below), whereas in the new example, `y` has a shorter lifetime (the inner block). This means that if you extend the scope of the borrow to the outer block, you see an error, because now it exceeds the lifetime of `y`.

So I guess I think a change like this is probably a good idea, but it should probably be part of an overall attempt to clarify the distinction that a *borrow* has a scope which is determined by the places where the reference is **used**, whereas a variable has a lifetime which is determined by where it is declared. Naturally the scope of the borrow cannot be **longer** than the lifetime of the thing being borrowed. Moreover, the thing being borrowed cannot be moved or mutated during the scope of the borrow (but it could be moved or mutated afterwards).

(*) One downside is that when I say "scope" it implies that the lifetime of a borrow is always an entire block. This is not the case. Borrows can be confined to any subexpression or statement; they only extend to a block if they grow beyond a single statement. So for example:

```rust
{
    let x = vec![1, 2, 3];
    foo(&x); // scope of this borrow is confined to the fn call itself
    move(x); // note: `x` not borrowed here, so we can move it
}
```

(**) Actually, even in the original example, `x` and `y` have a slightly different lifetime, because technically each `let` introduces a new scope, reflecting the fact that the destructor for `y` (if it had one) would execute first, and thus `y` lives a slightly shorter amount of time than `x`. But this distinction isn't really relevant here.

The main change that I see being proposed is that the lifetime of the underlying resource (`y`) is inner block, and hence when we attempt to borrow `y` for the scope of the *outer* block, we get an error. I agree that it makes sense to show an example highlighting this, and that since, in the original example, both of the borrowed variables `x` and `y` had the same lifetime,

nikomatsakis:
Note: I was just saying to @steveklabnik on IRC that I'm not sure what words to use, but I definitely think it's useful to distinguish the lifetime of a reference (derived from where it is used) and the lifetime of the resource (derived from where it is declared). I was using the word "scope" for the latter, but it occurs to me it might be a better fit for the *former*. Thus, the "lifetime" of a reference must not be larger than the "scope" of the borrowed variable. Food for thought.

critiqjo:
>In particular, the lifetime of a reference (the `'a` in a `&'a T` type) refers to the smallest scope that contains *all the uses* of the reference. *There is a separate concept that is the lifetime of the underlying, borrowed resource (the one that the reference refers to). Naturally the scope of the reference cannot be larger* than the lifetime of the underlying resource.

So let me annotate the lifetimes in the following example:
```rust
let x = 9;
{
    let r1 = &'a x;
}
let r2 = &'b x;
```

@nikomatsakis Do you mean to say that `'a` != `'b` ? My understanding was that regardless of the scope of reference variables, the associated lifetime would be that of the referring object (i.e. `'a` = `'b` = lifetime of x).

kteza1:
I've very recently started learning lifetimes. May be a new guy's perspective might be helpful here.

```
let x = 9;
{
    let r1 = &'a x;
}
let r2 = &'b x;
```
Summarizing the discussion here,

Does `a`, `b` represent

*  life time of reference (`r1`, `r2`) 

(or)

* the life time of underlying resource (`x`) which  @critiqjo  is referring to as `associated lifetime`?

I feel more comfortable with option 2. 

I would read this as, " `r1` is a reference which is referring to a **resource whose lifetime is `a`** and hence **lifetime of `r1`** should be <= `a`" . I.e **Lifetime of reference should be <= Lifetime of resource**

Now I can easily relate why [this](https://play.rust-lang.org/?code=fn%20coerce_first%3C%27a%3A%20%27b%2C%20%27b%3E%28x%3A%20%26%27a%20i32%2C%20_%3A%20%26%27b%20i32%29%20-%3E%20%26%27b%20i32%20{%0A%20%20%20%20x%0A}%0A%0Afn%20main%28%29%20{%0A%20%20%20%20let%20x%20%3D%20800%3B%0A%0A%20%20%20%20let%20borrow_big%20%3D%20%26x%3B%0A%20%20%20%20let%20coerce_test%3B%0A%20%20%20%20{%0A%20%20%20%20%20%20%20%20let%20y%20%3D%208%3B%0A%20%20%20%20%20%20%20%20let%20borrow_small%20%3D%20%26y%3B%0A%0A%20%20%20%20%20%20%20%20let%20coerced%20%3D%20coerce_first%28borrow_big%2C%20borrow_small%29%3B%0A%20%20%20%20%20%20%20%20coerce_test%20%3D%20coerced%3B%0A%20%20%20%20%20%20%20%20println!%28%22%60coerced%60%20is%20{}%22%2C%20coerced%29%3B%0A%20%20%20%20}%0A}&version=stable) is failing because `coerce_test`'s lifetime is greater than the lifetime of `y`. (Though we are returning reference to `x`, we are saying that its associated lifetime is that of `y`)

I don't know if I have got this completely wrong.

nikomatsakis:
@critiqjo 

> So let me annotate the lifetimes in the following example
> ...
> Do you mean to say that `'a` != `'b` ? 

Yes, that is what I mean.

> My understanding was that regardless of the scope of reference variables, the associated lifetime would be that of the referring object (i.e. `'a` = `'b` = lifetime/scope of x).

I don't know what you mean by "associated lifetime", but as I said there are two distinct things at play. The reference itself has a lifetime which is the smallest lifetime that encompasses **all uses** of the reference. Meanwhile, the value  being borrowed has a scope, which determines when its destructor will execute. The lifetime of the reference is only connected to this scope in that the reference cannot outlive the scope. (Here I used scope/lifetime in the inverse sense of my prior comment, per https://github.com/rust-lang/rust-by-example/pull/655#issuecomment-153306447, which I've decided is probably a better way to go, since we already use lifetime *primarily* to refer to the dynamic extent of a reference).

nikomatsakis:
@kteza1 

> I feel more comfortable with option 2. 

Nonetheless, this is not how Rust's type system works, and there is a very good reason for that. Whenever you create a reference to some value `x`, you also impose limitations on what you can do with `x` -- in particular, you cannot move `x`, nor can you mutate `x` (and, if this is a mutable reference, you can't read `x` either). Therefore, we want the lifetime of the reference to be as small as possible, so that these limitations are minimized. If the lifetime is larger than it has to be, you wind up with spurious errors from the compiler.

critiqjo:
@nikomatsakis Thank you so much for taking your time and explaining it... One final doubt:
```rust
{
    let s: &str = "string"; // &'a
}
```
Here, `'a` is `'static`, right? But that is the biggest lifetime. My doubt is -- if lifetime of a reference is dependent only on the enclosing scope where it was declared, `'a` should be something strictly less than `'static`, right?

mdinger:
@nikomatsakis Thanks. That was really very helpful and clarified the issue tremendously.

As to what specific action should be taken regarding this PR currently, I'm uncertain. I think it could be wordsmithed to be slightly more precise and in line with @nikomatsakis' explanation but then it would still be at odds with the current general theme of the section which I think could cause confusion.

I think my overall preference though would be to revise this when the entire section is revised so it is at least consistent.

kteza1:
Going in the direction that Niko suggested, I started associating 'a, 'b annotations with references rather than resources.

I  didn't understand this sentence completely 

> In particular, the lifetime of a reference (the 'a in a &'a T type) refers to the **smallest** scope that contains all the **USES** of the reference.

If any one could give an example for this in documentation, that would be great.

Going forward. Would it be correct to think as below

```
fn coerce_first<'a: 'b, 'b>(r1: &'a i32, r2: &'b i32) -> &'b i32 {
    r1
}

fn main() {
    let x = 800;

    let coerce_test;
    {
        let y = 8;

        let coerced = coerce_first(&x, &y);
        coerce_test = coerced;
        println!("`coerced` is {}", coerced);
    }
}
```
| REFERENCES  | ANNOTATION | RERERRING TO |
|-------------|:----------:|-------------:|
| r1          |     'a     |            x |
| r2          |     'b     |            y |
| coerce_test |     'b     |            x |

'b : largest scope of **references** annotated with 'b => scope of 'coerce_test'

Now any **resource** which has **reference** tagged with 'b' should not have lifetime less than `coerce-test`, which `y` is violating.

mdinger:
@kteza1 I think that's accurate but if the reader has to resort to compiling tables and line graphs to comprehend an example, the example isn't obvious enough. If making this change is going to result in an example which is extremely subtle, trying to find a clearer example might be in order.

critiqjo:
@mdinger What do you think about the doubt I have? Does the embedded `&str` variable (in the previous example) has a `'static` lifetime, or is it shorter?

Until now, I interpreted this:
```rust
let s: &'static str = "string";
```
as "`s` points to a resource having static lifetime".

---

>The "scope" of a reference variable must not be larger than the "lifetime" (`'a` in `&'a T`) it is having.

The reason this point of view seems better to me is because the reference variable itself contains all the information to constrain its own scope. Whereas (as @nikomatsakis wrote):

>the "lifetime" of a reference must not be larger than the "scope" of the borrowed variable.

refers to the "scope of the borrowed variable", which is not directly captured in the reference (at least syntax-wise `x: &'a T`).

Anyway, I don't think this PR holds meaning anymore. So closing...

mdinger:
I don't think you're thinking of it correctly. `'static` should be a lifetime of the program regardless of where it is declared. Hence it works when you do the following:

```rust
fn inner<'a>() -> &'a str {
    let s: &'static str = "words";

    s
}

println!("{}", inner());
```

Then you might validly state that the lifetime `'a` must be bigger than the function but it doesn't need to be bigger than `main` to work properly. In this case though, you might state differently:

```rust
fn inner<'a>() -> &'a str {
    let s: &'static str = "words";

    s
}

fn outer<'a>() -> &'a str {
    inner()
}

println!("{}", outer());
```

Now the lifetime still doesn't need to be larger than `main` but it has to be larger than both `inner` and `outer`. Here though it would be different because the lifetime leaving the function must be `'static`:

```rust
fn inner() -> &'static str {
    let s: &'static str = "words";

    s
}
println!("{}", inner());
```

It's probably okay to think of it as coercion with the caveat: if something `'static` was coerced so that it is too small for whatever your lifetime is, you didn't evaluate it big enough. Re-evaluate with it bigger than your newly found biggest lifetime.

kteza1:
@mdinger Yeah. I feel it intimidating myself

critiqjo:
@mdinger Thank you for explaining it. I still can't get a deep picture of it, but its okay...

One suggestion I have in general is that -- along with introduction of new concepts, an example which produces an error as a consequence should also be illustrated. I'm not clear how to produce an error as a consequence of lifetime coercion, though. But having that will help understand it a lot better, imo. People tend to hypothesize something that *works* in a way that is convenient to them. A counter will likely break a lot of such hypotheses. Thanks again...

mdinger:
> People tend to hypothesize something that works in a way that is convenient to them. A counter will likely break a lot of such hypotheses.

This is really very true and helping them to hypothesize the correct model in the first place should be a primary goal. Hopefully a more correct and better example can be merged at a later time to help more with these concepts.

critiqjo:
@mdinger Here's is a simple enough example to counter my argument:

```rust
fn cp<'a>(pp1: &mut &'a i32, pp2: &&'a i32) {
    *pp1 = *pp2;
}

fn main() {
    let i = 5;
    {
        let j = 8;
        let mut x = &i;
        let y = &j;
        cp(&mut x, &y); // this works
    }
}
```

If the associated lifetime of `x` (i.e. `'a`) was equal to the scope of `i`, then `cp` would not have succeeded (since `T` is invariant in `&mut T`). Not sure whether this example is useful enough to be put up...

Sorry for posting in this old thread.

PR #663, opened 2015-11-14
Add an exlicit activity.

absoludity:
After working through rust-by-example, one thing I missed from previous experience with the go-lang tour is explicit activities.

If it's wanted, I'd like to go through and add explicit bite-sized activities where appropriate, so that each step has a sense of accomplishment, and involves people in *doing* the learning, not just reading and running the examples.

No problem if it's not wanted, or if there's a different format the activities should take. If it is wanted, I'll keep working my way through again adding them.

Thanks for all the work!

### DIFF
--- a/examples/hello/input.md
+++ b/examples/hello/input.md
@@ -5,6 +5,14 @@ This is the source code of the traditional Hello World program.
 `println!` is a [*macro*][macros] that prints text to the
 console.
 
+Activity: Click 'Run' above to see the expected output. Next, add a new
+line with a second `println!` macro so that the output
+shows:
+```
+Hello World!
+I'm a Rustacian!
+```
+
 A binary can be generated using the Rust compiler: `rustc`.
 
 ```


steveklabnik:
Sure, that sounds great. Thanks so much!

mdinger:
I also like the idea. I wonder if the format could be improved. Other doc sets group sections under different headers like [this](https://msdn.microsoft.com/en-us/library/5z06z1kb.aspx#Anchor_5). Maybe in keeping with the `### See also:` at the bottom of some [pages](http://rustbyexample.com/hello/print/print_debug.html), you could have `### Exercises:` as a section directly above it.

That would make it like:

```txt
text
...
...

### Exercises:

blah blah blah

### See also:

blah blah blah
```

absoludity:
@mdinger Yep, I think it would be good to have some kind of header and organisation. I'm not sure whether a heading above the ### See also: will always work well though. For eg. on the one I added in this PR, I put it right after the information about println! (ie. above the text about rustc) as that's where the context is. Other pages might be longer and have two small bite-sized activities at different points - not sure?

It would be nice if it was formatted in a way that people can just identify and skip them if wanted. I'll try your option with ### Exercises. I'll also just try simply putting the "Activity:" prefix in bold:

**Activity:** Click 'Run' above to see the expected output. Next, add a new line with a second `println!` macro so that the output shows:
```
Hello World!
I'm a Rustacian!
```

Or do you think that would be too distracting? It might be a good format compromise that still allows the activity to be in the correct context, but easily identifiable (and skip-able) too. Ideally, we'd mark it with a separate CSS class or similar so that it could have a slightly different background identifying the complete activity, but that's probably overkill at this stage :)

mdinger:
As it is, it kinda just reads like the main text. I'm not sure if bold would differentiate it enough. It'd definitely be worth exploring both options though.

Stylistic options will be fairly limited since it is just parsed markdown. Hashes (`#`) for headers and `> foo` for quotes are probably about all the formatting choices that are available aside from regular text, bullets, and code blocks.

PR #664, opened 2015-11-15
Activities 2

absoludity:
Updates the first activity to use bold so it's easy to identify and skip if wanted.

Added a second example activities section for the print example.

The bold seems to work well - it's simple, and as long as we only ever have one paragraph for an activities section (ie. one sentence or a bullet list of activities like the print example), it's easy to identify the whole block.

I'll do larger branches from here on in - assuming the format is OK.

### DIFF
--- a/examples/hello/input.md
+++ b/examples/hello/input.md
@@ -5,7 +5,7 @@ This is the source code of the traditional Hello World program.
 `println!` is a [*macro*][macros] that prints text to the
 console.
 
-Activity: Click 'Run' above to see the expected output. Next, add a new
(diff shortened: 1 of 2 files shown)


mdinger:
The activity seems fine to me and I think the style is also okay. I could go back and forth over whether or not the style is ideal but the fact is, the activity is useful and the style isn't bad. So I'm okay with it.

absoludity:
Thanks @mdinger . I did try the alternatives, but let re-do them as separate PRs and you can see what you think.

PR #665, opened 2015-11-16
Activities 2b

absoludity:
One alternative mentioned by @mdinger . 

https://github.com/absoludity/rust-by-example/blob/activities-2b/examples/hello/print/input.md

Given that each page/example is short (is there nearly always just one example per page?), it seems to be fine having the activities at the bottom with a header. I don't mind either way.

Another one to come.

### DIFF
--- a/examples/hello/input.md
+++ b/examples/hello/input.md
@@ -5,14 +5,6 @@ This is the source code of the traditional Hello World program.
 `println!` is a [*macro*][macros] that prints text to the
 console.
 
-**Activity**: Click 'Run' above to see the expected output. Next, add a new
-line with a second `println!` macro so that the output
-shows:
-```
-Hello World!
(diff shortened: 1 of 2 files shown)


mdinger:
Yeah, always one example per page. We can add multiple code blocks but only one runnable example.

I think I like this the best and I don't really like the activity appearing in the middle of the page. At the bottom or right before the "See also:" (if it exists) seems best.

Maybe the colons on `See also` should be removed as well because it looks like it might look cleaner without. They've been there pretty consistently so it would be a bunch of really small fixes to change that.

Don't really like https://github.com/rust-lang/rust-by-example/pull/666 I guess. It works but it seems atypical to use that style for that purpose.

steveklabnik:
Seems good, I like this one too. Thanks!

PR #669, opened 2015-11-23
Activities 3

absoludity:
Adds activities for the print_display, format and tuples examples.

For the format page, I replaced the full list of formatting traits with a link. No problem to put it back if that's not wanted (it just seemed easy to open up if interested).

### DIFF
--- a/examples/primitives/tuples/input.md
+++ b/examples/primitives/tuples/input.md
@@ -19,8 +19,8 @@ use tuples to return multiple values, as tuples can hold any number of values.
     accepts a matrix as an argument, and returns a matrix in which two elements
     have been swapped. For example:
 ```
-print!("Matrix:\n{}", matrix)
-print!("Transpose:\n{}", transpose(matrix))
+println!("Matrix:\n{}", matrix)
+println!("Transpose:\n{}", transpose(matrix))
 ```
 results in the output:
 ```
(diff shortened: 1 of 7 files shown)


examples/primitives/tuples/input.md
```diff
@@ -4,3 +4,32 @@ using parentheses `()`, and each tuple itself is a value with type signature
...
+( 1.1 1.2 )
+( 2.1 2.2 )
+```
+    You may want to refer back to the example for [print display](print_display).
+ 2. Add a `transpose` function using the `reverse` function as a template, which
+    accepts a matrix as an argument, and returns a matrix in which two elements
+    have been swapped. For example:
+```
+print!("Matrix:\n{}", matrix)
+print!("Transpose:\n{}", transpose(matrix))
```
mdinger: Are these supposed to be `println!()`?

mdinger:
Cool. I didn't actually build the examples but those seem like nice additions. I like how these try to make people more familiar with `Display` by using it rather than just assuming they remember how to use it. Using something is definitely a more reliable way to get someone to remember something.

examples/primitives/tuples/input.md
absoludity: Yes - updated. Thanks.

absoludity:
Sorry @mdinger , @steveklabnik , I'd pushed the fix for the println! macro the other day but forgot to mention it. Let me know if there's anything else you think could be improved here, otherwise, land away.

steveklabnik:
Thanks!

PR #670, opened 2015-11-24
Remove ", that\n> "

kraai:
I'm not sure that this is the correct solution, but I'm pretty sure it should be changed in some way.

### DIFF
--- a/examples/mod/visibility/visibility.rs
+++ b/examples/mod/visibility/visibility.rs
@@ -15,7 +15,7 @@ mod my {
 
     // Items can access other items in the same module
     pub fn indirect_access() {
-        print!("called `my::indirect_access()`, that\n> ");
+        print!("called `my::indirect_access()`");
 
         // regardless of their visibility
         private_function();


kraai:
The same text also appears in at least "10.4 super and self".  I'll look into updating that as well.

mdinger:
Did you run this after changing it? This change results in:

```text
called `my::indirect_access()`called `my::private_function()`
```
instead of 

```text
called `my::indirect_access()` that
> called `my::private_function()`
```

The intent is to show that `indirect_access()` is indirectly accessing `private_function()`. `that` suggests something will be following and `>` is used to distinguish it from the other *self contained* lines while maintaining almost the exact same format. These two lines go together while still being distinguishable.

What specific problem are you trying to solve?

kraai:
Oops, I hadn't run the program and didn't realize that it was supposed to be that way.  Thanks for the explanation.

mdinger:
Sure

PR #672, opened 2015-11-29
Fixed broken link

queenp:
Link aliasing wasn't happening, caused a 404 (visible on running tests, might be worth making testrunning mandatory?). Fixed now.


steveklabnik:
Thanks!

Making the check mandatory means that we can't deploy if the site goes down for unrelated reasons... This happened before 😅

queenp:
Ah, fair does!

PR #673, opened 2015-11-29
Grammatical fixes for its/it's

IanWhitney:
Three very small fixes, each replacing its for it's, or vice versa. These were all of the examples I could find in the current master branch.


steveklabnik:
Thanks so much!

PR #674, opened 2015-12-01
Add new error handling section

mdinger:
I added a new error handling section which I based on the @burntsushi error handling blog doc. In general I think his docs motivate error handling really well however at the end of the docs where they diverge from strings as errors, it wasn't quite as good. You're left wondering why you'd ever bother implementing the `Error` trait and `From` and `Display` when you only need `Display`. Or perhaps you need `From` also. Anyway, I tried a slightly different approach by trying to approach the error trait and others as basically optional depending on the purpose and needs of the library/binary. He probably said more in the case study but I never actually looked at that part (so it may be absolutely wonderful).

In some sense, this may be quite practical because the examples are entirely usable and runnable in comparison to the original. The original, while each is runnable, the playpen links are basically useless which wouldn't work in this case.

I originally didn't like that I had to write the result error handling sections by loading files but after the fact, it seems like a good result and I'm fairly happy with it.

@steveklabnik r?

@burntsushi If you have any opinions/critiques, please add them.

I'm not sure who else I could cc. I could possibly rename some of the sections and perhaps add some more `See also`s because I might not have added enough but I'm not sure.

Sorry it took so long to get this up. I had this mostly to this point for a while but was taking a break and doing other things for a couple weeks I guess.

---

This branch is currently hosted [here](http://mdinger.github.io/rbe_error_handling). The error handling section is [here](http://mdinger.github.io/rbe_error_handling/error.html).

### DIFF
--- a/examples/error/option_with_result/combinator_combinations/result_try.rs
+++ b/examples/error/option_with_result/combinator_combinations/result_try.rs
@@ -3,14 +3,16 @@ use std::fs::File;
 
 type Result<T> = std::result::Result<T, String>;
 
-// Setup to make this work. Create two files with some info.
+// Setup to make this work. Create two files with some info. Ignore the
+// return values because we don't care about them here.
 fn setup() {
-    // Ignore the return value because we don't care about it.
-    let _ = File::create("a")
-        .and_then(|mut file| file.write_all(b"grape"));
+    File::create("a")
+        .and_then(|mut file| file.write_all(b"grape"))
+        .unwrap();
 
-    let _ = File::create("b")
-        .and_then(|mut file| file.write_all(b"fruit"));
+    File::create("b")
+        .and_then(|mut file| file.write_all(b"fruit"))
+        .unwrap();
 }
 
 // Get the data from each file with the data stored in a `Result`.
--- a/examples/error/option_with_result/combinator_combinations/result_try.rs
+++ b/examples/error/option_with_result/combinator_combinations/result_try.rs
@@ -1,0 +1,49 @@
+use std::io::prelude::*;
+use std::fs::File;
+
+type Result<T> = std::result::Result<T, String>;
+
+// Setup to make this work. Create two files with some info.
+fn setup() {
+    // Ignore the return value because we don't care about it.
+    let _ = File::create("a")
+        .and_then(|mut file| file.write_all(b"grape"));
+
+    let _ = File::create("b")
+        .and_then(|mut file| file.write_all(b"fruit"));
+}
+
+// Get the data from each file with the data stored in a `Result`.
+fn get_data(path: &str) -> Result<String> {
+    File::open(path)
+        .map_err(|err| err.to_string())
+        .and_then(|mut file| {
+            let mut contents = String::new();
+
+            // Read the data into `contents`.
+            file.read_to_string(&mut contents)
+                .map_err(|err| err.to_string())
+                // Ignore the output `read_to_string` returns and return `contents`.
+                .map(|_| contents)
+        })
+}
+
+// Concat the contents of the two files together into a new `Result`.
+fn concat(a: &str, b: &str) -> Result<String> {
+    let (data_a, data_b) = (get_data(a), get_data(b));
+    
+    data_a.and_then(|a|
+        // Return `Ok` when both `a` and `b` are `Ok`. Otherwise return
+        // whichever has the first `Err`.
+        data_b.and_then(|b| Ok(a + &b))
+    )
+}
+
+fn main() {
+    setup();
+
+    match concat("a", "b") {
+        Ok(n)  => println!("{}", n),
+        Err(e) => println!("Error: {:?}", e),
+    }
+}
--- a/examples/error/option_with_result/combinator_combinations/result_try.rs
+++ b/examples/error/option_with_result/combinator_combinations/result_try.rs
@@ -46,6 +46,6 @@ fn main() {
 
     match concat("a", "b") {
         Ok(n)  => println!("{}", n),
-        Err(e) => println!("Error: {:?}", e),
+        Err(e) => println!("Error: {}", e),
     }
 }
--- a/examples/error/reenabling_box/input.md
+++ b/examples/error/reenabling_box/input.md
@@ -18,9 +18,9 @@ types. In order to define a valid `Result<T, E>` type, the user has a few choice
 * convert it to `String` or some other intermediate choice
 * box it up into `Box<Error>` via type erasure
 
-Boxing it is a common choice. The only penalty is that the error type is only known
-at runtime and not statically determined. All that needs to be done to enable this
-is implement the `Error` trait:
+Boxing it is a common choice. The only penalty is that the underlying error type is only known
+at runtime and not [statically determined][dynamic_dispatch]. All that needs to be done to enable
+this is implement the `Error` trait:
 
 ```rust
 trait Error: Debug + Display {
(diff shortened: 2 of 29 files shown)


examples/error/option_with_result/combinator_combinations/result_try.rs
```diff
@@ -0,0 +1,49 @@
+use std::io::prelude::*;
+use std::fs::File;
+
+type Result<T> = std::result::Result<T, String>;
+
+// Setup to make this work. Create two files with some info.
+fn setup() {
+    // Ignore the return value because we don't care about it.
+    let _ = File::create("a")
+        .and_then(|mut file| file.write_all(b"grape"));
```
BurntSushi: I might be missing important context, but I would be very careful about using `let _ = ...`. Does using `unwrap` here work? If so, that would be much better.

steveklabnik: :+1:, `let _` is an antipattern imho

examples/error/reenabling_box/input.md
```diff
@@ -0,0 +1,41 @@
...
+```
+
+For example, a user may use a variety of libraries which each provide their own error
+types. In order to define a valid `Result<T, E>` type, the user has a few choices:
+
+* define a new wrapper error type around the external libraries error types
+* convert it to `String` or some other intermediate choice
+* box it up into `Box<Error>` via type erasure
+
+Boxing it is a common choice. The only penalty is that the error type is only known
```
BurntSushi: I might say "the underlying type" or the "type that impls `Error`" instead of "the error type." `Box<Error>` *is* a type and it is known statically. (It would be clearer to name the underlying type, but I don't think you have one in scope here?)

BurntSushi:
@mdinger This looks pretty good! I will say though, that while I read through everything, I had a really hard time evaluating it because it's out of order in Github's diff view. It makes it difficult to see whether the flow of information works. Is there a simple way to view this in its published form?

examples/error/option_with_result/combinator_combinations/result_try.rs
mdinger: I can fix that

examples/error/reenabling_box/input.md
mdinger: I like your recommendation. I'll probably update it tomorrow.

mdinger:
@BurntSushi I just uploaded it [here](http://mdinger.github.io/rbe_error_handling/). It's just a copy so while most things seem correct, the editor syntax highlighting seems to be missing. Syntax highlighting still works properly normally so it's probably that I didn't copy something properly but I'm not sure what.

mdinger:
Thanks. If you see any pages that need a `See also`, let me know. Some probably do but I missed them.

examples/error/reenabling_box/rethink.rs
```diff
@@ -0,0 +1,75 @@
...
+        match *self {
+            // No underlying cause so return `None`.
+            DoubleError::EmptyVec => None,
+            // The cause is the underlying implementation error type. Is implicitly
+            // cast to the trait object `&error::Error`. This works because the
+            // underlying type already implements the `Error` trait.
+            DoubleError::Parse(ref e) => Some(e),
+        }
+    }
+}
```
mdinger: Please check my `impl` of `error::Error`. I think the above case should be `None` but wasn't sure.

examples/error/reenabling_box/rethink.rs
```diff
@@ -0,0 +1,75 @@
...
+        })
+    }
+}
+
+impl error::Error for DoubleError {
+    fn description(&self) -> &str {
+        match *self {
+            // Trying to use `Display` here results in the string being immediately
+            // dropped which won't compile so we have to write out the phrase again.
+            DoubleError::EmptyVec => "please use a vector with at least one element",
```
mdinger: I really don't like how this repetition is necessary, especially in an explanation. It reeks of the library not being clean enough to avoid this repetition (and the explanation has to point out the issue to avoid people being puzzled by it when they think of it on their own). This led to me thinking that maybe if `Display` was implemented something like [this](http://is.gd/ztHFwx), then you could retrieve the slice by calling `&self.display()` but I think it would have the same issue.

Maybe if `description` returned `Cow` instead of `&str`. Anyway, I don't think it's ideal but for the time being seems okay. Hopefully the std library will make this nicer in the future.

examples/error/option_with_result/combinator_combinations/result_try.rs
mdinger: fixed

examples/error/reenabling_box/input.md
mdinger: Updated. I thought `underlying error type` might be better and I linked to dynamic dispatch also which seems like a good change. I think it should be fixed.

mdinger:
I updated the rendered site as well

examples/error/reenabling_box/rethink.rs
BurntSushi: LGTM

mdinger: Good.

examples/error/reenabling_box/rethink.rs
BurntSushi: `description` is meant to be a very short description of the error. The duplication in this case is kind of special, since you don't have any data attached to `DoubleError::EmptyVec` to interpolate into the error string (which is usually what happens in the `Display` impl).

For pedagogical purposes, you could purposefully shorten `description`. e.g., "empty vectors not allowed."

BurntSushi: Also, when implementing these things, I usually like to defer to the underlying impl when possible. e.g., `e.fmt(f)` instead of `e.to_string()` for the `DoubleError::parse` variant. (I suppose this is a matter of taste.)

BurntSushi:
@mdinger Thanks for uploading a rendered version of this! It was very helpful. I read through and things LGTM. Nice work. :+1:

examples/error/reenabling_box/rethink.rs
mdinger: Is there much benefit to doing so? `to_string()` is definitely more approachable (well, maybe just because I've never seen the other before) and using `fmt` in that fashion looks like it would any `write!` usage to happen on each arm. That is:

```rust
write!(f, "Error: {}", match *self {
	DoubleError::EmptyVec =>  "please use a vector with at least one element".to_owned(),
	DoubleError::Parse(ref e) => e.to_string(),
})
```

replaced with:

```rust
match *self {
    // Every arm which doesn't defer to something will need to spell this out
	DoubleError::EmptyVec => write!(f, "Error: please use a vector with at least one element"),
	DoubleError::Parse(ref e) => e.fmt(f),
}
```

After looking at it for a bit though, I can kinda see how it might be appealing but it doesn't seem like a huge difference.

Is it typical to include `Error: ...` in either display or the description or is that usually just at the very top level where it is printed out?

BurntSushi: It's definitely not a huge difference. I like it because it reads nicely: "oh okay, the implementation for fmt trivially defers to the same method call for variants that already impl Display." It's a code clarity thing, so reasonable people can disagree.

BurntSushi: Yeah, don't put Error: like that, because i can definitely see a main function adding that in. I guess prefer the smallest amount of context possible? You never know how deeply nested your error might be.

mdinger: Okay, I kinda like that. A plus is it points out something subtly useful that they might not have noticed.

mdinger:
Sweet! I committed some more fixes and updated the render. The commit message states what changed. I'm pretty happy with the result.

@BurntSushi Thanks. You've been tremendously helpful.

examples/error/reenabling_box/rethink.rs
mdinger: Plus, I was able to change the comment on `description`. It no longer sounds so much like a library weakness; just a distinction between the two different traits.

steveklabnik:
Thanks so much!

mdinger:
Awesome! Fastest rbe review for a big PR ever!

steveklabnik:
As far as I'm concerned, that's due to @burntsushi :)

PR #683, opened 2015-12-29
Fix whitespace in playpen result

parasyte:
This fixes the results display with any leading whitespace, as in the "formatted print" code on chapter 1.2. Specifically this line:

```rust
    // You can right-align text with a specified width. This will output
    // "     1". 5 white spaces and a "1".
    println!("{number:>width$}", number=1, width=6);
```


steveklabnik:
Thank you!

Issue #98, opened 2014-06-14
Create an accurate list of operators and their precedence


mdinger:
It would be useful to have a page which contains an accurate list of all the operators and their precedence.  In addition, it might be good to have an example usage of each but that is less important.

The reference manual has a precedence list [here](http://doc.rust-lang.org/reference.html#operator-precedence).

steveklabnik:
https://github.com/rust-lang/rust-by-example/pull/686

PR #686, opened 2015-12-30
Added link to precedence list of Rust reference

Xmasreturns:
#98 
Technically a bandaid, but it seems a little excessive to add another chapter/section on content that is so similar to other languages.


mdinger:
Yeah, this should be okay. That bug was filed 18 months ago and all rust docs are in a different place than they were at the time. I would close #98 after this merges.

steveklabnik:
Thanks!

PR #688, opened 2016-01-04
std/option Better? comment wording for unwrap

clemens-tolboom:
As a novice it took some time to get this. Splitting the comment makes it easier to me. What do you think?


mdinger:
This fix seems okay. I'd also check out the [error handling section](http://rustbyexample.com/error.html) which may be helpful.

steveklabnik:
Thanks1

clemens-tolboom:
tnx

PR #689, opened 2016-01-04
Updates to Chapter 9

Xmasreturns:
Readability/grammar, renamed DRY for clarity, and changed a macro name
for clarity

### DIFF
--- a/examples/macros/designators/designators.rs
+++ b/examples/macros/designators/designators.rs
@@ -1,23 +1,26 @@
 macro_rules! create_function {
-    // this macro takes an argument of "type" `ident`
-    // the `ident` designator is used for variable/function names
+    // This macro takes an argument of type `ident` and creates
+    // a function named `$func_name`.
+    // The `ident` designator is used for variable/function names.
     ($func_name:ident) => (
-        // this macro creates a function with name `$func_name`
         fn $func_name() {
-            // the stringify! macro converts an `ident` into a string
+            // The `stringify!` macro converts an `ident` into a string.
             println!("You called {:?}()",
                      stringify!($func_name))
         }
     )
 }
 
+// Create functions named `foo` and `bar` with the above macro.
 create_function!(foo);
 create_function!(bar);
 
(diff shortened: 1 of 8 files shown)


examples/macros/designators/designators.rs
```diff
@@ -1,23 +1,26 @@
 macro_rules! create_function {
-    // this macro takes an argument of "type" `ident`
-    // the `ident` designator is used for variable/function names
+    // This macro takes an argument of type `ident` and creates
```
steveklabnik: 'type' was in quotes here because it's not _really_ a type, exactly. I might just rephrase this to say "this macro takes an `ident` argument

steveklabnik:
Historically, we haven't used graves in code comments, since they don't get rendered differently anyway. So I'm tempted to say that we shouldn't be changing that. What do you think?

Xmasreturns:
I think graves in code comments are reasonable since they're visually noticeably different, which is the goal. In cases where the english ends up being the same word as a keyword or similar, it can get confusing. That being said, in these cases I made the additions/changes so that the chapter is at least internally consistent. Is there a style guide somewhere that can be referenced?

steveklabnik:
>  Is there a style guide somewhere that can be referenced?

Not one that mentions this, exactly. We haven't been working on the style guide until after rustfmt is fully working, so what we do have is not exactly great.

mdinger:
Actually, it's fairly [common](http://rustbyexample.com/generics.html) to use them on this site and I think it's useful for emphasis anyway lacking easier methods (like bold/italics). I think this is just making the docs more consistent and wouldn't bar it's use here.

I would postpone basing this on the style guide until the guide is at least in better shape (at least for now).

steveklabnik:
Okay, if @mdinger likes them, let's leave them in :)

Xmasreturns:
Sounds good, any other changes you'd like me to make before merging?

steveklabnik:
Nope, that's it!

PR #692, opened 2016-01-07
Updates to 13. Generics and rename 12.1 to `dead_code`

Xmasreturns:
12.1 to `dead_code` to more accurately reflect contents.
13. Generics - Readability and reducing confusion with wording.

### DIFF
--- a/examples/generics/generics.rs
+++ b/examples/generics/generics.rs
@@ -15,11 +15,12 @@ fn main() {
     // `Single` is concrete and explicitly takes `A`.
     let _s = Single(A);
     
-    // Instantiating generic types can be explicit or implicit.
-    // `SingleGen` explicitly specialized:
+    // Here, `SingleGen` is explicitly specialized. This reads as:
+    // Create a variable `_char` of type `SingleGen<char>`
+    // and give it the value `SingleGen('a')`
     let _char: SingleGen<char> = SingleGen('a');
 
-    // `SingleGen` implicitly specialized:
+    // `SingleGen` can also be implicitly specialized:
     let _t    = SingleGen(A); // Uses `A` defined at the top.
     let _i32  = SingleGen(6); // Uses `i32`.
     let _char = SingleGen('a'); // Uses `char`.
--- a/examples/generics/generics.rs
+++ b/examples/generics/generics.rs
@@ -1,25 +1,26 @@
-// A concrete type `T`.
-struct T;
+// A concrete type `A`.
+struct A;
 
-// The first use of `T` was not preceded by `<T>` so `Single` must
-// be a concrete type. `T` is defined at the top.
-struct Single(T);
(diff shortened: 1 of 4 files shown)


examples/generics/generics.rs
```diff
@@ -1,25 +1,26 @@
-// A concrete type `T`.
-struct T;
+// A concrete type `A`.
+struct A;
```
mdinger: I have reservations about this. The questioning pattern goes: "Why  does `Generic<T>(T)` need to be written with `<T>` before `(T)`?" and the answer is because `T` may actually be a concrete type somewhere else in this code base. Generally people don't write code this way but it is actually a possibility. By using `T` as a concrete type at the top, this highlights that the generic *must* use `<T>` so it may be understandable.

If you use `A`, then the questioning pattern now is implicit instead of explicit which should make the rule less obvious. That is, you can state the reason the `<T>` is required but the reader must make the theoretical change in their mind rather than decipher the example as written.

I realize this is a slightly confusing example but I don't see a great way around that. If you have any suggestions, I'd be happy to hear them.

Xmasreturns:
My opinion on documentation is generally that it should be easy to read first and foremost. From the perspective of someone new to a language, every example in the documentation should be unambiguous in all ways. I think a reader would expect specific use cases or edge cases being properly supported by the examples, but not necessarily used.

In this case, at the simplest level, "Why does Generic<T>(T) need to be written with <T> before (T)?" is essentially answered by listing the rules. If the rules aren't clear enough, I would rather change the wording on them than make the example harder to understand. Similarly, your highlight that the generic requiring use of `<T>` in the case of `T` existing as a concrete type would also not be out of place in the .md

mdinger:
Okay, I think your reasoning is sound. If you think you could tweak the .md file and it still read reasonably, I'd be interested in that too. This same type of thing also comes up in [generics/`impl`](http://rustbyexample.com/generics/impl.html) though someone tweaked it similarly sometime in the past. I mention the second page because it is just another case of `<T>` basically being used to protect the inner type from being concrete; the earlier in generics that the reader can clearly recognize that fact, the better. Then they might even make a good guess at the cases which rbe doesn't cover (which I'm sure there are at least a few).

Xmasreturns:
Alright, hopefully those changes are along the same lines as what you were thinking.

examples/generics/input.md
```diff
@@ -1,17 +1,27 @@
...
+`T` of any type:
 
-These rules play out like this:
+```rust
+fn foo<T>(T) { ... }
+```
+
+Because `T` has been specified as a generic type parameter, it is considered
+generic when used here as `(T)`. This is the case even if `T` has previously
+been defined as a `struct`.
```
mdinger: I like it. It should definitely be clearer and more understandable than it used to be so good job!

steveklabnik:
Thanks!

PR #696, opened 2016-01-14
std::box: remove a lint that is not needed

tshepang:


mdinger:
@steveklabnik bors doesn't work here yet

steveklabnik:
hehe i opened like 7 tabs at once, and so i was in the habit

PR #698, opened 2016-01-19
Updates to Ch13 - Readability, wording, grammar

Xmasreturns:
Renamed 13.7.1 "The problem" -> "The Problem"
Removed some unhelpful code and reworded some description to be less redundant between sections

### DIFF
--- a/examples/generics/assoc_items/input.md
+++ b/examples/generics/assoc_items/input.md
@@ -1,19 +1,13 @@
-Associated Items is an extension to `trait` generics which encompasses a set
-of rules pertaining to [`item`s][items] of various types. It allows `trait`s
-to internally define a few new items including:
+"Associated Items" refers to a set of rules pertaining to [`item`][items]s
+of various types. It is an extension to `trait` generics, and allows 
+`trait`s to internally define new items.
 
-* `const`s via the keyword `const`.
-* `type`s via the keyword `type`. This is called an *output* type.
-
-Some advantages are:
-
-* `trait`s now have access to `const`s.
-* `type`s get simpler usage patterns when the `trait` is generic over
-containers.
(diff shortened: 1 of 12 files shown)


examples/generics/assoc_items/input.md
```diff
@@ -1,19 +1,13 @@
...
 
 ### See also:
 
-[RFC](
-https://github.com/rust-lang/rfcs/blob/master/text/0195-associated-items.md
-)
+[RFC][RFC]
+
 [items]: http://doc.rust-lang.org/reference.html#items
+[RFC]: https://github.com/rust-lang/rfcs/blob/master/text/0195-associated-items.md
```
mdinger: I think this was a carryover from when the max line lengths were shorter and the parser wouldn't accept `[abc]:` and `url` on different lines (as a historical note).

Xmasreturns: Any idea if this is still the case? If so, then I can revert this to make sure it passes checks.

mdinger:
This is really really great!

examples/generics/assoc_items/input.md
mdinger: If you're building the docs, then just manually check the page to make sure the link appears properly. If it does, it should be fine. If you aren't building the docs, make sure the line length is less than [100 characters](https://github.com/rust-lang/rust-by-example/blob/master/check-line-length.sh#L8) (actually, I'm not sure that's the right link but I couldn't find any others). Maybe we don't check md files for line length...I thought we did...

Xmasreturns: Line should be 83 characters, so I think we're good here

steveklabnik:
:confetti_ball:

PR #699, opened 2016-01-19
Add reason for casting arithmetic

prasannavl:


examples/cast/cast.rs
```diff
@@ -22,6 +22,7 @@ fn main() {
     println!("1000 as a u16 is: {}", 1000 as u16);
 
     // 1000 - 256 - 256 - 256 = 232
+    // Basically, it just uses the 8 bits from the least significant bit, and truncates the MSB
```
steveklabnik: Hey @prasannavl ! I like the idea of expanding this, but maybe it would be a bit better if you removed the "basically" and "just" and expanded the "most significant bit"?

prasannavl: Makes sense. 


How about “Under the hood, the 8 bits from the least significant bit (LSB), and truncates the most significant bit (MSB).”

I feel using the abbreviations helps with the quick glances :)

prasannavl: Ah. Had too many typos. Updated above comment.

mdinger:
@steveklabnik Can you look at this? It's pretty old and it's pretty bad this has sat this long since it was updated.

steveklabnik:
Sorry about that! I apologize for letting this slip through the cracks.

PR #704, opened 2016-01-20
std_misc/process/pipe: remove unnecessary enclosing block

### LINKED ISSUE #702: std_misc/process/pipe might have an unnecessary block
http://rustbyexample.com/std_misc/process/pipe.html has an extra block around the write to stdin, but the program works identically without the block (tested `rustc 1.7.0-dev (257d3244f 2016-01-18)` on Ubuntu 15.10).

ivan:
Fixes #702.

Thanks to mbrubeck and sfackler for helping me out with this.

### DIFF
--- a/examples/std_misc/process/pipe/pipe.rs
+++ b/examples/std_misc/process/pipe/pipe.rs
@@ -15,22 +15,21 @@ fn main() {
         Ok(process) => process,
     };
 
-    {
-        // Write a string to the `stdin` of `wc`.
-        //
-        // `stdin` has type `Option<ChildStdin>`, but since we know this instance
-        // must have one, we can directly `unwrap` it.
-        match process.stdin.unwrap().write_all(PANGRAM.as_bytes()) {
-            Err(why) => panic!("couldn't write to wc stdin: {}",
-                               Error::description(&why)),
-            Ok(_) => println!("sent pangram to wc"),
-        }
-
-        // `stdin` gets `drop`ed here, and the pipe is closed.
-        //
-        // This is very important, otherwise `wc` wouldn't start processing the
-        // input we just sent.
+    // Write a string to the `stdin` of `wc`.
+    //
+    // `stdin` has type `Option<ChildStdin>`, but since we know this instance
+    // must have one, we can directly `unwrap` it.
+    match process.stdin.unwrap().write_all(PANGRAM.as_bytes()) {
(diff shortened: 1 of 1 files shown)


examples/std_misc/process/pipe/pipe.rs
```diff
@@ -15,23 +15,22 @@ fn main() {
...
+                           Error::description(&why)),
+        Ok(_) => println!("sent pangram to wc"),
     }
 
+    // Because `stdin` does not live after the above calls, it is `drop`ed,
+    // and the pipe is closed.
+    //
+    // This is very important, otherwise `wc` wouldn't start processing the
+    // input we just sent.
+
```
mdinger: This block (5 lines above this) should probably be removed because it is related to the braces you just removed. It makes little sense without them.

ivan: I left it in (and rephrased it slightly) because it seemed important to mention that the implicit dropping causes ChildStdin to close.  (If someone has a `let mut stdin = process.stdin;`, the program will hang forever.)

Is the phrasing incorrect?

mdinger: Ah, I see. If it is important to emphasize that point, you should save it to a variable and include the braces which were just backed out to emphasize the dropping. IDK, I guess it might be nitpicky though. Not sure.

ivan: I don't object to putting the braces back and adding a `let mut stdin =` to emphasize the dropping, if that's the Rust way of doing things.

mdinger: Maybe the wording is just odd. Can't think of anything better right now.

steveklabnik:
Any idea what w  should do here?

ivan:
To summarize, it could be

1) left as-is with a slightly confusing unnecessary block

2) changed as per my PR to remove the block, and rely on a non-obvious but commented implicit Drop

3) left as-is except with an extra `let stdin = ` local inside the block, to emphasize the Drop

But I really don't know what to do here because I don't know how this is written in real Rust programs.

steveklabnik:
I think I vote #1, after sleeping on it.

Thank you!

ivan:
I think you voted for number 2 since you merged it?

steveklabnik:
.... yes. I am very, very jetlagged, hehe. merging was what I wanted to do.

PR #707, opened 2016-01-26
Updates to Ch14.3 - Borrowing

Xmasreturns:
Grammar, readability, variable name changes, some reordering of
text/examples

### DIFF
--- a/examples/scope/borrow/freeze/input.md
+++ b/examples/scope/borrow/freeze/input.md
@@ -1,4 +1,4 @@
-When data is borrowed, it also *freezes*. *Frozen* data can't be modified via
-the original object, until all the references to it go out of scope.
+When data is immutably borrowed, it also *freezes*. *Frozen* data can't be 
+modified via the original object until all references to it go out of scope:
 
 {freeze.play}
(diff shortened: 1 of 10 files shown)


examples/scope/borrow/freeze/input.md
```diff
@@ -1,4 +1,4 @@
-When data is borrowed, it also *freezes*. *Frozen* data can't be modified via
-the original object, until all the references to it go out of scope.
+When data is immutably borrowed, it also *freezes*. *Frozen* data can't be 
+modified via the original object until all references to it go out of scope:
```
steveklabnik: This isn't super relevant here, but I wonder if all of this should be re-worked. "Freeze" used to mean something specific in Rust, but hasn't for a long time, and using these terms seems like a holdover from that time period.

Xmasreturns: I'm sort of on the same page, but it felt just a tad more redundant than anything else. As I was going through, I noticed that the wording on the Freeze page and the main Borrowing page are very similar. 

I kept them separate because the examples are different in complexity level, but I think that Borrowing could be turned into a short summary, and followed by its old content and the current content in Freeze

PR #708, opened 2016-01-27
hello/input: fix typo

viodlen:
According to the Rust book ([https://doc.rust-lang.org/book/getting-started.html](url)), Rustacian should be written Rustacean.


steveklabnik:
Thanks!

PR #710, opened 2016-02-02
Updates to 14.4 and added a link to Ch8

Xmasreturns:
Rewordings for clarification, general readability improvements

Added link to generics chapter in Ch8 where it sees use

### DIFF
--- a/examples/scope/lifetime/explicit/explicit.rs
+++ b/examples/scope/lifetime/explicit/explicit.rs
@@ -5,7 +5,7 @@ fn print_refs<'a, 'b>(x: &'a i32, y: &'b i32) {
     println!("x is {} and y is {}", x, y);
 }
 
-// A function which has no input, but has a lifetime parameter `'a`.
+// A function which takes no arguments, but has a lifetime parameter `'a`.
 fn failed_borrow<'a>() {
     let _x = 12;
 
@@ -17,7 +17,7 @@ fn failed_borrow<'a>() {
 }
 
 fn main() {
-    // Create variables which will be borrowed.
+    // Create variables to be borrowed below.
     let (four, nine) = (4, 9);
     
     // Borrows (`&`) of both variables are passed into the function.
--- a/examples/scope/lifetime/explicit/explicit.rs
+++ b/examples/scope/lifetime/explicit/explicit.rs
@@ -1,24 +1,3 @@
-// This example uses `main()` at the top so it can be read top to
-// bottom. `main()` is usually at the bottom.
-fn main() {
-    // Create variables which will be borrowed.
-    let (four, nine) = (4, 9);
-
-    // Borrows (`&`) of both variables are passed into the function.
-    // If a borrow is truly a borrow, the variable must be returned,
-    // otherwise a borrow would be transferring ownership. This means
-    // no matter what, *any* input which is borrowed via function or
-    // otherwise, *must* still exist after the borrower ceases. In
-    // other words, *Rule 2* must be true.
-    print_refs(&four, &nine);
-
-    // Note that there is no input even though the function specifies
-    // a lifetime. The sizedness of that lifetime will be determined
-    // by the caller (which is here) from available lifetimes (scopes).
-    // Any lifetime chosen will then be larger than that of the function.
-    failed_borrow();
-}
-
 // `print_refs` takes two references to `i32` which have different
 // lifetimes `'a` and `'b`. These two lifetimes must both be at
 // least as long as the function `print_refs`.
@@ -26,16 +5,29 @@ fn print_refs<'a, 'b>(x: &'a i32, y: &'b i32) {
     println!("x is {} and y is {}", x, y);
 }
 
-// There are no references to force `'a` to be larger than the
-// function, yet `'a` remains larger than the function.
+// A function which has no input, but has a lifetime parameter `'a`.
 fn failed_borrow<'a>() {
     let _x = 12;
 
-    // Attempting to use the lifetime `'a` as an explicit type
-    // annotation inside the function will fail because the
-    // lifetime of `&_x` is smaller than `y` has. A small lifetime
-    // cannot be coerced into a larger one.
-    //
-    //let y: &'a i32 = &_x;
     // ERROR: `_x` does not live long enough
+    //let y: &'a i32 = &_x;
+    // Attempting to use the lifetime `'a` as an explicit type annotation 
+    // inside the function will fail because the lifetime of `&_x` is shorter
+    // than that of `y`. A short lifetime cannot be coerced into a longer one.
+}
+
+fn main() {
+    // Create variables which will be borrowed.
+    let (four, nine) = (4, 9);
+    
+    // Borrows (`&`) of both variables are passed into the function.
+    print_refs(&four, &nine);
+    // Any input which is borrowed must outlive the borrower. 
+    // In other words, the lifetime of `four` and `nine` must 
+    // be longer than that of `print_refs`.
+    
+    failed_borrow();
+    // `failed_borrow` contains no references to force `'a` to be 
+    // longer than the lifetime of the function, but `'a` is longer.
+    // Because the lifetime is never constrained, it defaults to `'static`.
 }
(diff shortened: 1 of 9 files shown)


examples/scope/lifetime/explicit/explicit.rs
```diff
@@ -26,16 +5,29 @@ fn print_refs<'a, 'b>(x: &'a i32, y: &'b i32) {
     println!("x is {} and y is {}", x, y);
 }
 
-// There are no references to force `'a` to be larger than the
-// function, yet `'a` remains larger than the function.
+// A function which has no input, but has a lifetime parameter `'a`.
```
steveklabnik: What about 'takes no arguments'?

examples/scope/lifetime/explicit/explicit.rs
```diff
@@ -26,16 +5,29 @@ fn print_refs<'a, 'b>(x: &'a i32, y: &'b i32) {
...
     let _x = 12;
 
-    // Attempting to use the lifetime `'a` as an explicit type
-    // annotation inside the function will fail because the
-    // lifetime of `&_x` is smaller than `y` has. A small lifetime
-    // cannot be coerced into a larger one.
-    //
-    //let y: &'a i32 = &_x;
     // ERROR: `_x` does not live long enough
+    //let y: &'a i32 = &_x;
```
steveklabnik: put a space between `// let`, please :smile:

examples/scope/lifetime/explicit/explicit.rs
```diff
@@ -26,16 +5,29 @@ fn print_refs<'a, 'b>(x: &'a i32, y: &'b i32) {
...
+    // Attempting to use the lifetime `'a` as an explicit type annotation 
+    // inside the function will fail because the lifetime of `&_x` is shorter
+    // than that of `y`. A short lifetime cannot be coerced into a longer one.
```
steveklabnik: Is there a reason this was moved until after? It's that way below as well, but aren't most explanations before, not after? Maybe I'm confusing the style of RBE with the style of The Book.

examples/scope/lifetime/explicit/explicit.rs
```diff
@@ -26,16 +5,29 @@ fn print_refs<'a, 'b>(x: &'a i32, y: &'b i32) {
...
+}
+
+fn main() {
+    // Create variables which will be borrowed.
```
steveklabnik: Saying 'will be borrowed' is a bit ambiguous here. They'll eventually be borrowed, but right now, they're owned. Maybe something like "which we will be borrowing in a few lines" or something to make that a bit more clear would be good.

examples/scope/lifetime/explicit/input.md
```diff
@@ -1,21 +1,31 @@
-The borrow checker utilizes explicit lifetime annotations to reason about
-how long references should be valid. Failure to annotate lifetimes[^1] is akin
-to banning the borrow checker from validating borrows and so accordingly,
-annotation is mandatory.
+The borrow checker uses explicit lifetime annotations to determine
+how long references should be valid. In cases where lifetimes are not
+elided[^1], failure to annotate lifetimes is akin to banning the borrow
```
steveklabnik: I know that you're just re-flowing this, but "banning the borrow checker" sounds strange here. Maybe something more like

> The borrow checker uses explicit lifetime annotations to determine how long references should be valid. In
> cases where lifetimes are not elided, Rust requires explicit annotations to determine what the lifetime of a
> reference should be. The syntax for explicitly annotation a lifetime uses an apostrophe character as follows:

Thoughts?

examples/scope/lifetime/explicit/input.md
```diff
@@ -1,21 +1,31 @@
...
+```rust
+foo<'a>
+// `foo` has a lifetime parameter `'a`
+```
 
-1. `'a` and `'b` will represent names for lifetimes with non-specifiable
-(generic) types.
-2. The lifetime of `foo` may not exceed either lifetimes `'a` or `'b`.
+Usage of lifetimes requires generics (similar to [closures][anonymity]) 
+since lifetimes have no type or name associated with them. Additionally,
```
steveklabnik: This is a bit weird, given that `"a` is a name. Hm.

examples/scope/lifetime/fn/fn.rs
```diff
@@ -16,19 +16,17 @@ fn print_multi<'a, 'b>(x: &'a i32, y: &'b i32) {
...
-// dropped leaving a reference to invalid data to be returned.
-//
-// The reason the problem is caught is because of the restriction
-// `<'a>` imposes: `'a` must live longer than the function.
-//fn invalid_output<'a>() -> &'a i32 { &7 }
-
-// While returning references without input is banned, returning
-// references that have been passed in are perfectly acceptable.
-// One restriction is the correct lifetime must be returned.
+// Returning references without input is banned, but returning
```
steveklabnik: This isn't strictly true, they must be `'static`.

```rust
fn foo() -> &'static str {
    "this works"
}
```

examples/scope/lifetime/fn/fn.rs
```diff
@@ -16,19 +16,17 @@ fn print_multi<'a, 'b>(x: &'a i32, y: &'b i32) {
...
+// references that have been passed in is acceptable.
+// One restriction is that the correct lifetime must be returned.
```
steveklabnik: Hm, I feel like we could improve this whole section a lot, but am not totally sure what I'd put here.

examples/scope/lifetime/fn/input.md
```diff
@@ -1,16 +1,18 @@
-Functions with lifetimes have a few different valid forms. Ignoring
-[elision][elision] for the time being, the rules for function parameters are:
+Ignoring [elision], function signatures with lifetimes have a few constraints: 
 
 * any reference *must* have an annotated lifetime.
 * any reference being returned *must* have the same lifetime as an input or
 be `static`.
 
+Additionally, note that returning references without input is banned, as it
```
steveklabnik: Same comment as above.

steveklabnik:
Thanks a lot for this! We've got some stuff to work on, but this is great overall :+1:

examples/scope/lifetime/explicit/explicit.rs
Xmasreturns: Agreed, this change will show up

examples/scope/lifetime/explicit/explicit.rs
Xmasreturns: This formatting is in line with other example code that contains commented out lines. For example, see the examples on [13.4.1](http://rustbyexample.com/generics/bounds/testcase_empty.html), [13.8](http://rustbyexample.com/generics/phantom.html), and [10.2](http://rustbyexample.com/mod/struct_visibility.html)

examples/scope/lifetime/explicit/explicit.rs
Xmasreturns: This is done to show that the "let" statement causes an error, so the "ERROR:..." line is a flag for that. The description then follows, so that it's clearly a flag for improper code

examples/scope/lifetime/explicit/explicit.rs
Xmasreturns: Agreed, a change to this will show up

examples/scope/lifetime/explicit/input.md
Xmasreturns: Agreed, this is definitely an improvement and will show up

examples/scope/lifetime/explicit/input.md
Xmasreturns: Hm. I'm making some reductions here in wording, but hopefully it doesn't become too vague.

examples/scope/lifetime/fn/fn.rs
Xmasreturns: This line was somewhat unnecessary to begin with and has been removed to favor the below changes

examples/scope/lifetime/fn/input.md
Xmasreturns: This has been made less strict: 
"Additionally, note that returning references without input is banned **if**..."

examples/scope/lifetime/fn/fn.rs
Xmasreturns: This has been slightly reworded, but still feels improve-able.

examples/scope/lifetime/explicit/explicit.rs
steveklabnik: Ouch, okay.

mdinger:
Cool! Lots of minor flow improvements are great!

Issue #658, opened 2015-11-06
Lifetime explanations don't currently account for lifetimes inherent flexibility


mdinger:
PR https://github.com/rust-lang/rust-by-example/pull/655 pointed out that the section on lifetimes doesn't properly specify the size of a lifetime. A specific confusion was whether the lifetime size is chosen by the scope of the reference or the scope of the owned variable. In actuality, the size may be chosen as one or the other or anywhere in between depending on context. Changing the context can change the lifetime. This should be detailed more properly.

See https://github.com/rust-lang/rust-by-example/pull/655#issuecomment-153303105 in particular for more specifics of how it should be improved.

Xmasreturns:
As an update, I tried to address some of this in #711 and #710

steveklabnik:
I'm going to close this one out since those PRs were merged; @mdinger if you have more specifics, please open a new issue!

PR #711, opened 2016-02-03
Updated 14.4.5, 14.4.6

Xmasreturns:
Improved readability, changed coercion example to involve less confusion
with respect to scoping (and reduce redundancy)

### DIFF
--- a/examples/scope/lifetime/lifetime_bounds/bounds.rs
+++ b/examples/scope/lifetime/lifetime_bounds/bounds.rs
@@ -25,6 +25,6 @@ fn main() {
     let x = 7;
     let ref_x = Ref(&x);
     
-    print(ref_x);
     print_ref(&ref_x);
+    print(ref_x);
 }
--- a/examples/scope/lifetime/lifetime_bounds/bounds.rs
+++ b/examples/scope/lifetime/lifetime_bounds/bounds.rs
@@ -24,7 +24,7 @@ fn print_ref<'a, T>(t: &'a T) where
 fn main() {
     let x = 7;
     let ref_x = Ref(&x);
-    
+
     print_ref(&ref_x);
     print(ref_x);
 }
--- a/examples/scope/lifetime/lifetime_bounds/bounds.rs
+++ b/examples/scope/lifetime/lifetime_bounds/bounds.rs
@@ -1,11 +1,11 @@
 use std::fmt::Debug; // Trait to bound with.
 
 #[derive(Debug)]
-// `Ref` contains a reference to type `T` where `T` is unknown
-// with an unknown lifetime `'a`. `T` is bounded such that any
-// references in `T` must outlive `'a`. In addition, the lifetime
-// of `Ref` may not exceed `'a`.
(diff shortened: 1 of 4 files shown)


steveklabnik:
Looks good to me, other than the Travis failure. @mdinger what do you think?

Xmasreturns:
I swapped two lines that ended up with pointers to invalid data, so that should be reverted with this commit

examples/scope/lifetime/lifetime_bounds/bounds.rs
```diff
@@ -24,7 +24,7 @@ fn print_ref<'a, T>(t: &'a T) where
 fn main() {
     let x = 7;
     let ref_x = Ref(&x);
-
+
```
mdinger: I think this is kinda weird. I suspect the convention is blank lines are blank but am not sure

examples/scope/lifetime/lifetime_coercion/input.md
```diff
@@ -1,4 +1,6 @@
-A lifetime may be coerced into another if the the original is larger than
-the new lifetime:
+A longer lifetime can be coerced into a shorter one 
+so that it works inside a scope it normally wouldn't work in.
+This comes in the form of inferred coercion by the Rust compiler,
+and also in the form of declaring a lifetime difference:
```
mdinger: This sounds interesting and seems like a good place for a link to where this is elaborated on if such a place exists.

mdinger:
I basically think it sounds good but I took only a very cursory look at it. If those two nits can be addressed or commented (or show what the actual convention is in the first case), I'd be fairly happy with this.

As wordsmithing, I'd definitely be happy with it. Regarding whether it addresses the problems highlighted in https://github.com/rust-lang/rust-by-example/issues/658, I'd have to do a more thorough review and comparison with niko's comment to see how they compare which I probably won't have time for until at least the weekend if not later. However, since this section isn't ideal as it is anyway, reevaluating this section sometime in the future after this was to merge wouldn't be too bad in my opinion.

Basically, I'm good with it.

examples/scope/lifetime/lifetime_bounds/bounds.rs
steveklabnik: Yes, blank lines should be blank

Xmasreturns: This was unintentional, and will be changed.

examples/scope/lifetime/lifetime_coercion/input.md
Xmasreturns: In this particular case, I'm letting the example elaborate on the claim.

As far as I can find, no single explanation currently exists that doesn't add confusion to a (relatively) simple lifetime explanation by talking about other aspects of lifetimes, cryptic compiler behaviors, or scopes. My solution to the issue was to combine a couple viewpoints from IRC and reword as succintly as possible.

steveklabnik:
Sounds great to me then!

PR #719, opened 2016-02-26
Updated Ch15.2

Xmasreturns:
Readability changes to input.md and update to current description in the .rs example.


steveklabnik:
Thanks so much, as always :)

PR #720, opened 2016-03-05
Fix the build

mdinger:


examples/std/hash/alt_key_types/input.md
```diff
@@ -2,12 +2,11 @@ Any type that implements the `Eq` and `Hash` traits can be a key in `HashMap`.
 This includes:
 
 * `bool` (though not very useful since there is only two possible keys)
-* `int`, `uint`, and all variations thereof 
-(see [`VecMap`][vecmap] for a more streamlined map implementation keyed by `uint`)
-* `String` and `&str` (protip: you can have a `HashMap` keyed by `String` 
+* `int`, `uint`, and all variations thereof
```
mdinger: These should probably be `i32` or `isize` or something but I'm not sure what. I'm not gonna look into it but if you happen to know what they should be, I'll happily change it. I'm just mainly trying to keep rbe building.

steveklabnik: Yes, they should be, but I agree, let's keep it building.

PR #721, opened 2016-03-05
RAII: Prefix unused box variables with "_"

stevegury:
This example explains the concept of RAII but never use the `box*` variables.
It currently triggers compiler warning, which is confusing for new users and could
be easily avoided by prefixing the variables with an underscore.


steveklabnik:
Thanks!

PR #723, opened 2016-03-10
More accurate grammar terms in example

sdwebster:
"Predicate" here would be the entire latter part of the sentence, verb + object. See https://en.wikipedia.org/wiki/Predicate_(grammar) . Could alternatively write "verb_phrase" for "verb". This is admittedly a petty commit. Thanks so much for excellent starter materials!


steveklabnik:
> This is admittedly a petty commit.

Not at all! Grammar is important. Thank you!

PR #725, opened 2016-03-12
Updates to Ch15.2,15.4; Removed 15.3

Xmasreturns:
15.3 removed due to redundancy; information delta shifted to Ch13.4 (also bounds). Language changes to 15.2 and 4


examples/generics/bounds/input.md
```diff
@@ -11,13 +11,8 @@ fn printer<T: Display>(t: T) {
...
 ```
 
-One consequence is that generic instances are allowed to access the methods
-of the traits specified in the bounds. For example:
-
-{bounds.play}
-
-Bounding also restricts the generic from being specialized to
-types that do *not* conform to the bounds. That is:
+Bounding restricts the generic from being specialized to types that do
```
steveklabnik: So, I am kinda not sure about saying "specialized" here, since specialization is a specific feature that means something in this context, and hasn't actually landed yet. 

What about something like "restricts the generic to types that" instead?

mdinger: Maybe "from being applied to types"

Xmasreturns: I think I'm partial to "Bounding restricts the generic to types that conform to the bounds." Any other suggestions before I switch it to that?

mdinger: +1

steveklabnik:
I didn't see this was updated, sorry!

PR #727, opened 2016-03-18
generics/gen_fn: fix warning about unused variable

jesrui:
This is a small fix to remove the following
```
warning: unused variable: `s`, #[warn(unused_variables)] on by default
```
in the code of the `generics/gen_fn` example.


steveklabnik:
Thanks so much!

Issue #270, opened 2014-11-06
Rust by Example in PDF


afiskon:
It would be great if Rust by Example would be available in PDF. You can just create a single HTML-page with whole book content, in this case anyone could create a PDF using Print Friendly & PDF plugin for Google Chrome and other solutions.

mvallerie:
+1

Probably not an ideal workaround (neither for users, nor for rust maintainers, nor for rust servers, sorry about that) but this basically works (expect some minor issues anyway, and it's NOT single page) :

wget -r -k -np http://rustbyexample.com/

milanaleksic:
+1
For newcomers that wish to dive in, this seems like a reasonable approach.
Not only PDF, though - I'd recommend making MOBI and EPUB as well...

funivan:
+1
 `wget -r -k -np http://rustbyexample.com/` is good solition but sometimes pdf/mobi/epub is better

Anachron:
Just created this little script to download & update the most important rust-docs:

```bash
#!/bin/bash

wget -r -k -np http://rustbyexample.com/
wget -r -k -np -X "book,src,std" https://doc.rust-lang.org/reference.html
wget -r -k -np https://doc.rust-lang.org/nightly/rustc/
```

mdinger:
FYI: the [book: installation](http://doc.rust-lang.org/1.0.0-beta.2/book/installing-rust.html)

> This installer also installs a copy of the documentation locally, so you can read it offline. On UNIX systems, /usr/local/share/doc/rust is the location. On Windows, it's in a share/doc directory, inside wherever you installed Rust to.

filib:
:+1:

a0x:
I've been dealing with this issue in this whole morning, and I found something.
After `make book`, in the `stage/` dir, there'are all markdown files, which are well constructed in the way of the table of contents.
![_20160417125350](https://cloud.githubusercontent.com/assets/3829546/14585141/1a6cfdd4-049c-11e6-9452-d47870c162d4.png)


So things become how to generate a single PDF file via a group of markdown files.

a0x:
I'm still working on this issue, in progress, but met a problem.

In the source file, `rust-playpen` is used to display and run the code in web page, and it is used as a plugin written in `book.json` file. 

[...]

PR #737, opened 2016-04-13
EPUB and PDF generation

tpasternak:
Initial support for PDF and EPUB generation. Implements #270.

For sure some CSS will be needed in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,9 @@ script:
   - rustc --version
   - make
   - make test
-  - make book
+  - make html
+  - make pdf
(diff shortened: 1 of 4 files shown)


mdinger:
Have you run this? I think I tried this in the past but the runnable examples never looked decent. Did adding `<pre></pre>` help?

tpasternak:
Yes, I have. The tag `<pre>` is for keeping the original indentation of the code.It's not beautiful, but readable.

steveklabnik:
Thanks!

PR #738, opened 2016-04-13
Updated Ch16,16.1,16.2,16.3, 16.4

Xmasreturns:
Renamed Ch16.3, 16.4
Changes to prose that hopefully improve readability.
Minor changes to example code to unify voice.
Apologies in advance for not splitting the rename commit with edits in Ch16.3


mdinger:
Oh, I think you renamed/moved something in the table of contents which can't be found anymore. That error is always weird. Or you didn't rename an example file or something.

mdinger:
I like the updates here though I only looked at them fairly briefly. +1

Xmasreturns:
Hmm, after changing the id/title in the table of contents file, I renamed the folders, as well as the .rs files in those folders. If I recall correctly, I did the same thing with other chapters before, but someone might need to doublecheck that I did it properly this time around.

Xmasreturns:
Actually, I wonder if it's because the `unwrap` section has a panic in it? It's on purpose and happens with the current .rs file as well, so unless something else is getting borked, that particular build issue should be fine.

examples/error/and_then/input.md
```diff
@@ -0,0 +1,20 @@
...
+in the nested `Option<Option<T>>`. Chaining multiple calls together can 
+then become confusing.
+
+That's where `and_then()` comes in. Known in some languages as flatmap, `and_then()` calls its function input with the wrapped value or returns `None` if the `Option` is `None`.
+
+In the following example, `cookable_v2()` results in an `Option<Food>`. 
+Using `map()` instead of `and_then()` would have given an `Option<Option<Food>>`, 
+which is an invalid type for `eat()`.
+
+{combinators.play}
```
mdinger: The problem is this should now be called `and_then.play`

Xmasreturns: Nice catch! That definitely slipped by me.

steveklabnik:
Thanks!

Issue #642, opened 2015-09-03
Suggestion for simple translate system for this repo?


Narann:
I was wondering if it could be doable to have a simple translate system. For example:

`examples/attribute/input.md` could be rename `examples/attribute/input-en.md` and be the default.

This mean, if user switch to `french` (for example), it will provide: `examples/attribute/input-fr.md` if it exists. If not, it will provide a warning and display the `english` version.

With such approach it would be easy for users to PR translated version of the original `input-en.md` in there native language.

Does it make sense?

steveklabnik:
I haven't heard much demand to actually do said translations, but yeah, having it translatable eventually would be nice.

Narann:
As rust-by-example is full of "one shot" pages, it's a good candidate for fast translating (IMHO).

mdinger:
It'd be nice if it was handled by the book generator but without integrating a language translation tool, it would have a tendency for getting out of date. Already rust-by-example tends to let things get out of date, especially when they're whitelisted examples because they don't cause failure when things change.

The translation tools I've seen will actually tell you when the translation page if lagging behind the base english page and someone must decide how to proceed with the translation.

It'd be nice to have though.

joemphilips:
Interesting. What is that translation tools name ?
Maybe I will translate this page into Japanese. In that case I will send PR lately.

mdinger:
Well rbe doesn't have a tool setup for it yet. The one I saw freecad use is called [crowdin](https://crowdin.com/project/freecad). They advertise free usage for open source. It would need setting up and a button to switch between translations would have to be added.

joemphilips:
Looks fascinating!

Oh, by the way, I have decided to translate rbe into Japanese.
https://github.com/rust-lang-ja/rust-by-example-ja

mdinger:
Cool. Might be useful to put up a link on the main readme as a stopgap for
proper language support soon then.
On Apr 17, 2016 11:15 AM, "Joe Miyamoto" <notifications@github.com> wrote:

Looks fascinating!

Oh, by the way, I have decided to translate rbe into Japanese.
https://github.com/rust-lang-ja/rust-by-example-ja

—
You are receiving this because you commented.
Reply to this email directly or view it on GitHub
<https://github.com/rust-lang/rust-by-example/issues/642#issuecomment-211042740>

steveklabnik:
Now that we've ported to mdbook, this is a duplicate of https://github.com/azerupi/mdBook/issues/5

PR #745, opened 2016-04-30
add link to Japanese translation

joemphilips:
as I mentioned in,
#642 
I am now translating rbe to Japanese.
I would appreciate if someone gives me suggestion, especially for switching languages or how to track update.


steveklabnik:
Wonderful!

These questions are ones we don't have good answers to yet :(

PR #746, opened 2016-05-01
Adding semicolons to example code – primitives/tuples.html

JackDanger:
Copying and pasting these lines results in a compilation error unless semicolons are added.
This wasn't immediately obvious from the error message.


JackDanger:
I love @rust-highfive so much.

steveklabnik:
Ah thanks! This is a great change.

(And I love @rust-highfive too :heart_eyes: )

PR #747, opened 2016-05-01
Diversifying enum person examples – custom_types/enum.html

JackDanger:
I was distracted a little when working through this example by the `Fat`/`Skinny` dichotomy. I changed that to `Engineer`/`Scientist` and I edited the names of the example people to be a slightly more interesting set.

### DIFF
--- a/examples/custom_types/enum/enum.rs
+++ b/examples/custom_types/enum/enum.rs
@@ -3,12 +3,12 @@
 
 // Create an `enum` to classify someone. Note how both names
 // and type information together specify the variant:
-// `Skinny != Fat` and `Height(i32) != Weight(i32)`. Each
+// `Engineer != Scientist` and `Height(i32) != Weight(i32)`. Each
 // is different and independent.
 enum Person {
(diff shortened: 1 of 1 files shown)


examples/custom_types/enum/enum.rs
```diff
@@ -3,12 +3,12 @@
...
 // Create an `enum` to classify someone. Note how both names
 // and type information together specify the variant:
-// `Skinny != Fat` and `Height(i32) != Weight(i32)`. Each
+// `Engineer != Scientist` and `Height(i32) != Weight(i32)`. Each
 // is different and independent.
 enum Person {
     // An `enum` may either be `unit-like`,
-    Skinny,
-    Fat,
+    Engineer,
```
JackDanger: It was actually pretty hard to come up with an equally dichotomous pair that wasn't offensive but I suspect the target audience of this tutorial will appreciate this pair.

steveklabnik:
Awesome, thanks a ton!

PR #748, opened 2016-05-02
Moving macro section to after traits

JackDanger:
Even as a lisp veteran none of the macro examples were useful without understanding the basics first. Some of the more complex macro examples introduce trait restrictions on generics. A new Rustacean needs to learn about traits and scope and ownership before they can follow the `macro_expand!` section.

### DIFF
--- a/examples/structure.json
+++ b/examples/structure.json
@@ -69,12 +69,6 @@
         ] },
         { "id": "hof", "title": "Higher Order Functions", "children": null }
     ] },
-    { "id": "macros", "title": "macro_rules!", "children": [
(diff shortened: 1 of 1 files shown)


steveklabnik:
Sounds great, I certainly haven't thought a lot about the optimal ordering here. There may be a bunch of other issues like this...

mdinger:
FWIW, one of my first PRs attempting to expand the macros section but it never got merged. Even if the approach was good, it probably doesn't exactly follow the current rules since it was written before release. In case that's useful to you.

Issue #742, opened 2016-04-20
Chapter 5.1 Literals has a weird first sentence


symingz:
The first sentence of Chapter 5.1 is this:

"Numeric literals can be type annotated by adding the type as a suffix, with the exception of usize that uses the usize suffix and isize that uses the isize suffix."

According to this sentence, usize and isize are not different from other types, why call them exceptions?

mdinger:
This is probably because the `isize` and `usize` used to use `is` and `us` as suffixes which made them distinct from the others. So the types were properly updated but the sentence was not. Feel free to submit a PR fixing it. Thanks for pointing it out!

steveklabnik:
https://github.com/rust-lang/rust-by-example/pull/754 fixed this!

PR #754, opened 2016-05-14
Remove exception about isize/usize and add example #742

arvidj:


mdinger:
Fixes https://github.com/rust-lang/rust-by-example/issues/742

Thanks

PR #755, opened 2016-05-18
flow_control/match/struct: delete duplicated entry

joemphilips:
since `examples/flow_control/match/struct` and `examples/flow_control/match/destructuring/destructure_structures/` seems duplicated, I erased a former one.


steveklabnik:
Thanks!

PR #756, opened 2016-05-18
custom_types/structs.html: add activities

arvidj:
Adding some more activities. Is it okay if I create pull-requests directly, or should I create issues first?


steveklabnik:
No need to open issues! Thanks for this.

Issue #741, opened 2016-04-20
Chapter 5 Casting: incorrect use of the term "two's complement"


symingz:
In the code example, the last comment says "two's complement of 232 is -24". This statement is false according to the definition of two's complement in Wikipedia, which is consistent with my memory of text book definitions. I think the correct way to say this is "the 8bit unsigned representation of 232 is the same as the 8bit two's-complement representation of -24".
To generalize, other uses of "two's complement" in this code example are also wrong.

srinivasreddy:
@steveklabnik  please close this issue

PR #757, opened 2016-05-18
cast/cast.rs: try to clear up two's complement #741

arvidj:
Not sure if this is the ideal way to explain either, but perhaps a little less confusing than the original explanation. Also improve the wording of MSB/LSB.


steveklabnik:
I agree, this wording is much nicer. Thank you!

arvidj:
Thanks for the review! I think #741 can be closed now!

PR #758, opened 2016-05-23
Use method syntax when getting the description of an error

shepmaster:
These get copied enough by newbies that then show up on Stack Overflow that I decided to nip it at the source. :-)


steveklabnik:
Thanks!

PR #759, opened 2016-05-28
Fix dead link

aakloxu:


steveklabnik:
We try to keep links relative so that they work locally too; would you mind fixing that in that way?

aakloxu:
@steveklabnik It's a good idea to keep links relative so that translations of other languages which are located in the subdirectory of root, i.e. Chinese and Japanese, don't have to modify the links. At this moment, the reference links of the Japanese translation located in the subdirectory are all dead links because they don't fixed the absolute links.

By the way, the reference links of the RBE are all absolute except this one of the chapter, and  is it necessary to replace all the absolute links with the relative ones of the RBE?

joemphilips:
> At this moment, the reference links of the Japanese translation located in the subdirectory are nearly all dead.

Thanks, I didn't recognize it until you mentioned. 
In the case of Japanese it's probably better to make it relative e.g.  `../lifetime/static_lifetime.html`, since Chinese translations are in that way.

maybe it's better for this repository to follow the same way.

steveklabnik:
Yes, in general, in most of Rust's docs, we make them relative. I would certainly take PRs to update the rest. In almost all of Rust's docs, we use relative links, so this repo is being strange.

Thank you for the PR!

mdinger:
I usually did them relative to the root of rbe, which in this case would be `/scope/lifetime/static_lifetime.html` although I can see now why it could be an issue. It has never been mentioned as an issue before. Also, we've never had clones which modify the base url in the way these translations do which is probably why it's appearing now.

PR #767, opened 2016-06-29
Rename unit conversion

joemphilips:
the title of section 12.8.1 is `unit conversion` but there is no conversion really so I suppose `unit clarification` is better

### DIFF
--- a/examples/structure.json
+++ b/examples/structure.json
@@ -101,7 +101,7 @@
             { "id": "types", "title": "Associated types", "children": null }
         ] },
         { "id": "phantom", "title": "Phantom type parameters", "children": [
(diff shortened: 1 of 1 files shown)


steveklabnik:
@joemphilips it seems like there's a lot of unrelated commits in here; maybe double check what you did?

joemphilips:
excuse me ...

steveklabnik:
No worries!

Uh, one last thing: it seems like I'm the author of the commit somehow. can you fix that as well? Happy to merge after.

joemphilips:
Hmmm........ wonder why that happend.
now it's fixed.

Issue #770, opened 2016-07-10
8.2.2 Fn FnMut FnOnce text needs updating


alexandermerritt:
http://rustbyexample.com/fn/closures/input_parameters.html

~~Most of that is incorrect and~~ does not demonstrate how FnOnce executes once.

mdinger:
If you wish anything to be fixed, you need to be very specific. As it is, this issue has provided nothing which is actionable and may never be fixed.

steveklabnik:
Yes, I'm not sure what specifically is incorrect here.

talchas:
The section

>   -  Fn: takes captures by reference (&T)
>   -  FnMut: takes captures by mutable reference (&mut T)
>   -  FnOnce: takes captures by value (T)

is very visible, and while it is corrected in the next paragraph, is still misleading - it certainly sounds like the default changes depending on the context's closure type, when it doesn't, and even the last line seems likely to further this confusion for thread::spawn's 'static bound.

The error messages (even in the example) expose this, though the more direct `apply(|| ...)` now gives a friendly error pointing at the move location if you change apply to Fn/FnMut.

Also if this section is about specifying closures as parameters it seems odd to never mention the limitations on the caller for each of these, since you generally want the most general you can manage. If this is about understanding API documentation/source that uses them this is less of a problem I guess?

Something like "can only take captures by reference (&T); can also take captures by mutable reference (&mut T)" would be clearer, IMO. I don't know how or where best to mention the closure caller's limitations.

mdinger:
FWIW, this was based upon a [huonw blog post](http://huonw.github.io/blog/2015/05/finding-closure-in-rust). [This](http://huonw.github.io/blog/2015/05/finding-closure-in-rust/#structs-and-captures) might be a point fairly close to where he discusses things like this.

Why do you say it does not change based on context? The article I linked seemed to state that they are flexible in that fashion and this page attempted to reflect that sentiment. Have closures changed since that article was written to invalidate the article? I'm not sure what you're implying regarding `thread::spawn's 'static` bound. Could you clarify?

Noting the caller limitation and refining the wording both seem reasonable.

@Xmasreturns You're really good at making text read cleanly. Do you have ideas for this page? It looks to me like it could possibly be shorter by a bit but I haven't done more than just look at it. I wouldn't have time to play with it until the weekend I'm fairly sure.

talchas:
[...]

PR #772, opened 2016-07-13
Reordering and rewording in Ch8.2, 8.2.2

Xmasreturns:
Based on some comments in issue #770:

The new wording in 8.2 should better introduce closures as they work in
Rust.

Did some reordering in 8.2.2 to make it less misleading if reader skims
the page. Length stayed roughly the same, but hopefully it's clearer in
general.

### DIFF
--- a/examples/fn/closures/input.md
+++ b/examples/fn/closures/input.md
@@ -1,13 +1,18 @@
-Closures in Rust, also called lambdas, are functions 
-that can capture the enclosing environment. Their syntax and capabilities make them
-very convenient for on the fly usage. Some characteristics include:
+Closures in Rust, also called lambdas, are functions that can capture 
+the enclosing environment. For example, a closure that captures the x 
+variable:
+```Rust
+|val| val + x
+```
 
-* uses `||` instead of `()` around input variables.
-* *both* input and return *types* can be inferred.
-* input variable *names* must be specified.
-* body delimination (`{}`) is optional for a single expression. Mandatory
-otherwise.
-* the outer environment variables *may* be captured.
-* calling a closure is exactly like a function: `call(var)`.
+The syntax and capabilities of closures make them very convenient for 
+on the fly usage. Calling a closure is exactly like calling a function.
+However, both input and return types *can* be inferred and input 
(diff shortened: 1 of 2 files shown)


examples/fn/closures/input.md
```diff
@@ -1,13 +1,18 @@
...
-* calling a closure is exactly like a function: `call(var)`.
+The syntax and capabilities of closures make them very convenient for 
+on the fly usage. Calling a closure is exactly like calling a function.
+However, both input and return types *can* be inferred and input 
+variable names *must* be specified.
+
+Other characteristics of closures include:
+* using `||` instead of `()` around input variables.
+* optional body delimination (`{}`) for a single expression (mandatory otherwise).
+* optional capturing of the outer environment variables.
```
mdinger: I'm not sure about optional for captures since it implies you can just choose to not capture as opposed to capturing just being a detail regarding what you're trying to do. @steveklabnik You think it's okay here?

mdinger:
This looks good to me aside from that one question noted above.

Note: I was discussing this with @Xmasreturns over in https://github.com/Xmasreturns/rust-by-example/pull/2 so I'm hardly impartial but looking it over again, it seems good to me and a definite improvement.

examples/fn/closures/input.md
mdinger: Maybe:
```
possible capturing of the outer environment variables
the ability/capacity/potential to capture the outer environmental variables
```

I'm not sure what the right word is but I don't think `optional` is quite right.

Xmasreturns: I like the word choice in `ability`, but I'll wait on changes until we're sure the details are correct.

mdinger:
@steveklabnik Can you look this over. I think it's pretty okay. There's one minor issue noted above but no decision was made.

examples/fn/closures/input.md
steveklabnik: I like "ability" as well :+1: I will merge after that's updated.

Sorry for the long delay here :(

steveklabnik:
Thanks!

PR #781, opened 2016-07-27
Fix up process examples

### LINKED ISSUE #728: 18.5.2. Wait - doesn't work as described
[18.5.2. Wait](http://rustbyexample.com/std_misc/process/wait.html) claims the following:

> When a `Process` goes out of scope, its drop method will wait until the child process finishes before releasing the resource.

It seems that it is not true anymore.

### LINKED ISSUE #736: process wait example doesn't actually wait
Shouldn't the examples/std_misc/process/wait/wait.rs example look more like the following?

http://is.gd/sL9iL8

The way it is now, it doesn't wait for the sleep to be done.

### LINKED ISSUE #769: 18.5.2: It seems there is no wait while process goes out of scope
image
It seems not match with [the example](http://rustbyexample.com/std_misc/process/wait.html)
Any thing wrong?

ubsan:
Fix #728 

Fix #736 

Fix #769

### DIFF
--- a/examples/std_misc/process/input.md
+++ b/examples/std_misc/process/input.md
@@ -1,5 +1,5 @@
-The `ProcessOutput` struct represents the output of a finished child process,
-and the `Command` struct is a process builder.
+The `process::Output` struct represents the output of a finished child process,
+and the `process::Command` struct is a process builder.
 
 {process.play}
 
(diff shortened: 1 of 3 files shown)


brson:
Thanks @ubsan.

PR #787, opened 2016-08-07
Switch from f64 to f32

### LINKED ISSUE #786: Page: Struct, Section: Activity. f32 vs f65
On [this](https://github.com/rust-lang/rust-by-example/blob/master/examples/custom_types/structs/input.md) page, the Activity says `takes a Point and a f32 as arguments `.But the Point has f64 types so I would need to cast from f32 to f64 in some way. Is this on purpose or it's a bug?

ubsan:
Fixes #786 

Allows the activity to work without casting

stylistic choice - could also change the activity to take f64.

Unknown which is better.

### DIFF
--- a/examples/custom_types/structs/structs.rs
+++ b/examples/custom_types/structs/structs.rs
@@ -2,12 +2,12 @@
 struct Nil;
 
 // A tuple struct
-struct Pair(i32, f64);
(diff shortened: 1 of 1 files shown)


steveklabnik:
Seems good, thanks!

PR #789, opened 2016-08-17
Removed derive attribute, to avoid confusion

slayerjain:
I think the '#[derive(Debug)]' is redundant here. Removing it could save people like me from confusion.


steveklabnik:
Thanks!

PR #790, opened 2016-08-19
Use structs, not tuples, for generics impl example.

Havvy:
Over IRC, had a user say they found the usage of tuple get syntax
to be confusing (`self.0`) so this patch replaces the usage of
that with more mundane syntax so that learners can focus on
the impl syntax.


steveklabnik:
Agree wholeheartedly, thanks!

PR #792, opened 2016-08-21
unsafe: cast::transmute is now called mem::transmute

xfix:


steveklabnik:
Hey, thank you for this! So sorry that I let it slide for so long, not sure how that happened.

PR #794, opened 2016-08-31
fn/methods: Clarify method comparison to static call with reference.

johnthagen:
When going through the example, I think it would be more clear if the full `Rectangle::` resolution was provided to make it clear that a static method is being compared against.

If the user were to copy and paste `perimeter(&rectangle)` into their editor to try it out, they will get a `unresolved name` error.


steveklabnik:
I agree, thank you!

PR #799, opened 2016-09-22
closures/output_parameters: fix typo

autra:
I'm not English native, but I believe it is the action of "Using closures as input parameters" that is possible, thus the verb should be singular right?

### DIFF
--- a/examples/fn/closures/output_parameters/input.md
+++ b/examples/fn/closures/output_parameters/input.md
@@ -1,4 +1,4 @@
-Using closures as input parameters are possible, so returning closures as 
+Closures as input parameters are possible, so returning closures as 
 output parameters should also be possible. However, returning closure types 
 are problematic because Rust currently only supports returning concrete 
 (non-generic) types. Anonymous closure types are, by definition, unknown 


mdinger:
Honestly, I'm not certain. I could argue `are` is valid but I could also see how your interpretation could be valid. `are` sounds more natural to me but I'm not an expert. I am native though.

autra:
Oh I see your point. I'm not sure any more. Maybe I'm transposing the rules of my own language! [This link](http://www.grammarbook.com/grammar/subjectVerbAgree.asp), and though this special case is not directly mentioned, it tends to argue in favor of `is`. However, quite a lot of people of [this english stackexchange thread](http://english.stackexchange.com/questions/126726/each-of-these-is-vs-each-of-these-are) seems to agree with you and find it awkward...

Or maybe just:

> Closures as input parameters are possible

Let me know what you prefer, and I'll amend if necessary.

mdinger:
Your last suggestion sounds fine. I don't have a strong opinion about the others. I'd normally advocate fixes that are sure wins. Your final suggestion reads fine so if you like how it reads better, then submit that.

autra:
@mdinger updated :-)

steveklabnik:
I think both are fine, but am happy to take this patch :smile: Thanks!

PR #804, opened 2016-10-03
generics/phantom/testcase_units/units.rs: more info

francozappa:
- More info to final `println!` messages:
	- I think that the whole point of this exercise is to understand
	that we can use a generic Phantom parameter to enforce type constraints 
	at compile time, indeed is good to explicitly show the result of the sum with
	the two different unit of measure in the printed messages.
	
- More doc info about `Length` struct:
	- While I was re writing the example I wanted to add a `Bit` enum
	and use `Length` over `u32` (something like `let one_byte = // sum 8 one_bit `)
        value but then I realized that `Length` is not generic over the length itself.	
	
- I think it would be nice to extend this example using `Length`
	that is generic over `T`, is it feasible ?

### DIFF
--- a/examples/generics/phantom/testcase_units/units.rs
+++ b/examples/generics/phantom/testcase_units/units.rs
@@ -7,11 +7,12 @@ enum Inch {}
 #[derive(Debug, Clone, Copy)]
 enum Mm {}
 
-/// `Length` is a type with phantom type parameter `Unit`.
+/// `Length` is a type with phantom type parameter `Unit`,
+/// and is not generic over the length type (that is `f64`).
 ///
 /// `f64` already implements the `Clone` and `Copy` traits.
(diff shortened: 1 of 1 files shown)


mdinger:
> I think it would be nice to extend this example using Length that is generic over T, is it feasible ?

Yeah, it used to be that way. See https://github.com/rust-lang/rust-by-example/issues/582 and https://github.com/rust-lang/rust-by-example/issues/392 and both referenced PRs. In particular, see the first one. I think adding an example showing the full generic version as a more complicated case was mentioned.

mdinger:
Didn't know what the review button did but I wanted to try even if I can't officially approve it.

steveklabnik:
@mdinger yeah, anyone can approve :smile:

steveklabnik:
Thanks so much @francozappa !

PR #805, opened 2016-10-04
scope/lifetime: add a lifetime testcase

### LINKED ISSUE #802: Lifetime section should include an example where you need to finely specify them.
After going through the "lifetime" section, I failed to understand the point of lifetimes. I didn't see where I needed Foo<'a> or Foo<'a, 'b>. I tried different program I thought wouldn't compile, but they did! So at the end, I was like "ok, I know the syntax of lifetimes parameters, but when do I need to bother about these? Why can't the compiler put some 'a everywhere?". I mean, I _felt_ there was something going on, but couldn't _see_ it yet.

After going through [this section of the rust doc](https://doc.rust-lang.org/0.12.0/guide-lifetimes.html#parameter-lifetimes), things are a lot clearer to me. Basically, this example shows where you need to specify precisely the lifetime of each parameters

``` rust
// This does not compile yet, because 'r in inferred to be B lifetime. 
// it would if we specified different lifetimes for a and b on one hand, 
// and shape on the other hand.
fn select<'r, T>(shape: &'r Shape, threshold: f64,
                 a: &'r T, b: &'r T) -> &'r T {
    if compute_area(shape) > threshold {a} else {b}
}
                                                     // -+ r
fn select_based_on_unit_circle<'r, T>(               //  |-+ B
    threshold: f64, a: &'r T, b: &'r T) -> &'r T {   //  | |
                                                     //  | |
    let shape = Circle(Point {x: 0., y: 0.}, 1.);    //  | |
    select(&shape, threshold, a, b)                  //  | |
}                                                    //  |-+
                                                     // -+

```

 it really helped me to click things together.

To explain a bit more toroughly, the `coercion` section is not far from this, but failed to help me. 

``` rust
// Here, Rust infers a lifetime that is as short as possible.
// The two references are then coerced to that lifetime.
fn multiply<'a>(first: &'a i32, second: &'a i32) -> i32 {
    first * second
}

// `<'a: 'b, 'b>` reads as lifetime `'a` is at least as long as `'b`.
// Here, we take in an `&'a i32` and return a `&'b i32` as a result of coercion.
fn choose_first<'a: 'b, 'b>(first: &'a i32, _: &'b i32) -> &'b i32 {
    first
}
```

indeed `choose_first` could be written with only one lifetime parameter, just like `multiply`, and the coercion would work just fine. So I didn't see when I needed 'a *_and_ 'b, so I didn't see why I should specify 'a at all (why couldn't the compiler add them for me in all cases?).

I think this section would really benefit from an example like the one above WDYT? If you agree with this, I can make a PR. If so, what would be the best place for this example? Maybe we can add a "The problem" subsection like in "Associated items"? I do think we can find something better than shape and threshold though.

autra:
This commit adds an example where we need to specify different lifetimes
parameters for different pointers. It shows how the compiler can be
overly restrictive if we don't annotate them carefully.

Close #802.

### DIFF
--- a/examples/scope/lifetime/testcase_multiple_lifetimes/input.md
+++ b/examples/scope/lifetime/testcase_multiple_lifetimes/input.md
@@ -1,0 +1,102 @@
+Now that we know how to specify lifetime parameters, let's see what difference
+they can make in practice.
+
+Consider the following example and let's see what happens:
+
+```rust
+#[derive(Debug)]
+// Rust compiler cannot use elision here, because there is 2 pointers. But we
+// decide to use only one lifetime parameter here, 'foo will get coerced to
+// the intersection of the lifetimes of a and b at runtime.
+struct Foo<'foo> {
+    a: &'foo i32,
+    b: &'foo i32,
+}
+```
+
+We can write a `get_b` function that returns a reference to b:
+
+```rust
+// Here we are saying that foo must outlive the return pointer. In other word,
+// that the returning pointer is tied to foo.
+// Indeed, if we discard foo, &foo.b won't be available any more.
+fn get_b<'foo>(foo: &'foo Foo) -> &'foo i32 {
+    println!("Extracting b from {:?}", foo);
+    &foo.b
+}
+
+fn main() {
+  let a = &1;
+  let b = &2;
+  // 'foo lifetime must be included in the intersection of a and b lifetime,
+  // because it borrows a and b. It is the case here.
+  // Note that we must declare foo after a and b, because variables go out of
+  // scope in the reverse order they are declared.
+  // In practice 'foo lifetime will be coerced to 'b lifetime.
+  let foo = Foo {a: a, b: b};
+
+  println!("&foo.b is {}", get_b(foo));
+}
+```
+
+Here, things works just fine. However, we are coercing the lifetime of `get_b`
+to the intersection of a and b lifetimes. This is overly restrictive, because
+the return value is in fact only conditioned by b's lifetime. This prevents this
+other program from compiling:
+
+```rust
+#[derive(Debug)]
+struct Foo<'foo> {
+    a: &'foo i32,
+    b: &'foo i32,
+}
+
+// actual signature (the lifetime is elided):
+// fn create_foo_and_get_b<'bparam>(b: &'bparam i32) -> &'bparam i32
+fn create_foo_and_get_b(b: &i32) -> &i32 {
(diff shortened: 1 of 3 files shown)


autra:
I don't know if I should change the r? But I guess @mdinger should be notified a least :-)

examples/scope/lifetime/testcase_multiple_lifetimes/input.md
```diff
@@ -0,0 +1,102 @@
...
+lifetime, and thus `get_b` lifetime parameter.
+
+The solution here is annotating lifetimes more precisely:
+
+{multiple_lifetime.play}
+
+Incidentally, we see here that lifetimes are not the same thing as scope: the
+lifetime of `first_b` propagate to `get_b` function, even though `first_b` is
+not in scope there. The lifetime of a value owner is indeed the union of its
+scope and all its borrowers' scope.
```
autra: After reading #655 I'm not even sure this makes sense... :-/

autra: What I want to say is maybe more something like "The lifetime of a resource is the union of the scope of its owner and borrowers", but I'm not sure it's true.

examples/scope/lifetime/testcase_multiple_lifetimes/input.md
```diff
@@ -0,0 +1,102 @@
+Now that we know how to specify lifetime parameters, let's see what difference
+they can make in practice.
+
+Consider the following example and let's see what happens:
+
+```rust
+#[derive(Debug)]
+// Rust compiler cannot use elision here, because there is 2 pointers. But we
```
steveklabnik: this isn't true; the compiler never does elision on structs currently.

examples/scope/lifetime/testcase_multiple_lifetimes/input.md
steveklabnik: yeah, i'm vaguely not liking this either, i dunno

examples/scope/lifetime/testcase_multiple_lifetimes/input.md
steveklabnik: @autra any interest in fixing this up so we can merge?

autra:
@steveklabnik yep! I don't have a lot of time these days. I'm trying to get back at it in the following weeks.

steveklabnik:
I'm going to give this a close due to inactivity; @autra if you get the time to update things, please let me know and I'm happy to re-open!

PR #806, opened 2016-10-05
scope/borrow/borrow.rs: more info

francozappa:
I think that `borrow_i32` is a better name in this case
because it is not restricted to `Box<i32>` but it can borrow
also a stack-allocated `i32`.

More emphasis is given to the type system using `i32` instead of `int`
in variables names. We can revert back to `int` if you don't like 
the book's syntax highlight.

Comments are changed accordingly.


steveklabnik:
Thank you! I agree that this is better.

PR #807, opened 2016-10-07
scope/lifetime/methods/methods.rs: minor fix

francozappa:
Put a semicolon and implicitly return the unit type.


steveklabnik:
Thank you!

PR #808, opened 2016-10-08
scope/borrow/alias/alias.rs: more info

francozappa:
Clarify single mutable reference vs multiple references

p.s. Maybe `alias` is not a super good name in this example, because I think is more correct
in the context of `type` aliasing. In this context we are just creating multiple references 
(different type from the owner) and not aliases (same type of the owner with different names).

Hope it helps :smiley:


steveklabnik:
> p.s. Maybe alias is not a super good name in this example,

I agree.

steveklabnik:
Thanks!

PR #809, opened 2016-10-10
fn/closures/closures.rs: more info

francozappa:
Removed the part about `professor_x` because it will be presented in the next example,
and because it is not true that is *impossible* to capture a value from another scope
with functions.

### DIFF
--- a/examples/fn/closures/closures.rs
+++ b/examples/fn/closures/closures.rs
@@ -19,14 +19,4 @@ fn main() {
     let one = || 1;
     println!("closure returning one: {}", one());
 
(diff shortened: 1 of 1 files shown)


steveklabnik:
:+1: thanks!

mdinger:
> it is not true that is impossible to capture a value from another scope with functions

Please elaborate. From that statement, I would have assumed this should compile: https://is.gd/9fbx0f

francozappa:
Hi @mdinger,

> Please elaborate. From that statement, I would have assumed this should compile: https://is.gd/9fbx0f

https://is.gd/3CDMQy

PR #811, opened 2016-10-11
fn/closures/capture/capture.rs: minor fix

francozappa:
Remove duplicate `print()`

### DIFF
--- a/examples/fn/closures/capture/capture.rs
+++ b/examples/fn/closures/capture/capture.rs
@@ -1,6 +1,6 @@
 fn main() {
     use std::mem;
-    
+
     let color = "green";
 
     // A closure to print `color` which immediately borrows (`&`)
@@ -11,7 +11,6 @@ fn main() {
     let print = || println!("`color`: {}", color);
 
     // Call the closure using the borrow.
-    print();
     print();
 
     let mut count = 0;
@@ -34,7 +33,7 @@ fn main() {
 
     //let reborrow = &mut count;
     // ^ TODO: try uncommenting this line.
-    
+
     // A non-copy type.
     let movable = Box::new(3);
 


examples/fn/closures/capture/capture.rs
```diff
@@ -12,7 +12,6 @@ fn main() {
 
     // Call the closure using the borrow.
     print();
-    print();
```
mdinger: BTW, this is intentional to showcase that `print` can be called multiple times. `inc` is the same. `consume` in contrast, cannot be called twice.

francozappa: I see

However I still think that is redundant because `inc` is explicitly changing something in the output while `print`ing twice the same thing is less expressive, maybe we can print a different color the second time to show that we can capture (borrow) any variable in the outer scope.

p.s. the `consume` part can be improved  as well, because it is giving more emphasis on `mem::drop` rather than on the closure itself. For example, if we remove `mem::drop(movable)` then we can call `consume()` many times (like `print` and `inc`) without getting any compilation error. Maybe explaining why, and explaining (as in the Rust book) the difference between a closure and a `move` closure is more interesting than the current use case.

However, I think that the closure's explanation on the Rust book is much better that the one in rbe.

@mdinger thanks for the feedback.

:see_no_evil:

mdinger: The book avoid details. Why does this example from the [book](https://doc.rust-lang.org/book/closures.html#closures-and-their-environment) not compile?

```rust
let nums = vec![1, 2, 3];

// These first 2 are equivalent
let takes_nums = || nums;
let takes_nums1 = || {
    nums
};
// This is not
let takes_nums2 = || {
    nums;
    // The semicolon means the type is not being returned when the closure is called
    // whereas, in the above example, it is.
};

println!("{:?}", nums);
```

Why is the 3rd example okay but the others are not? It is because the first 2 closures require ownership of the type in order to return the type when the closure is called. References are not sufficient. `mem::drop` is at least explicit as opposed to implicit.

mdinger: Examine this from below. For the closure, how is it decided if the closure takes by reference? 

```rust
let consume = || {
    println!("`movable`: {:?}", movable);
    // ^ At this point, the closure would determine that it should take `movable` by reference. Why?
    // Because that is all that is required in order to call `println!`
    mem::drop(movable);
    // Here, it changes because `movable` now must be moved because of the type restriction
    // mem::drop requires.
};
```

mdinger: Clearly this section is not as helpful as it might be if you find the book section better however, it would be better to provide more detailed information than less like the book does. Maybe some tweaks would make things more clear. I'm not sure where the problem came from. See [this](http://huonw.github.io/blog/2015/05/finding-closure-in-rust/) for a more in depth analysis of closures. Maybe it will clarify some of the things rbe was attempting to show.

steveklabnik:
@francozappa any interest in keeping up this PR? I agree with @mdinger that this could use more clarity, but this isn't the right way to do it.

francozappa:
> @francozappa any interest in keeping up this PR? I agree with @mdinger that this could use more clarity, but this isn't the right way to do it.

:+1:

PR #813, opened 2016-10-11
fn/closures/input_functions: refactor

francozappa:


examples/fn/closures/input_functions/input.md
```diff
@@ -1,12 +1,12 @@
-Since closures may be used as arguments, you might wonder if the same can be said 
-about functions. And indeed they can! However, because a function can 
-*never* capture variables, closures are strictly more flexible. Therefore, any 
-function which can take a closure as an argument can also take a function.
+Since closures may be used as arguments, you might wonder if the same can be said
+about functions. And indeed they can! If you declare a function that takes a
+closure as parameter, then any function that satisfy the trait bound of that
```
mdinger: `s/satisfy/satisfies/`

mdinger: So, maybe the text is slightly vague but I'm not sure. It looks like you're changing the intent of the article, maybe unintentionally. I think the same goes for https://github.com/rust-lang/rust-by-example/pull/809.

So, on the first page of [closures](http://rustbyexample.com/fn/closures.html), capturing is specified as using a variable from the environment without passing it via parameters. Until regular functions gain the ability to do that, discussions on closures must talk about that. If the nomenclature is slightly wrong, then it should be updated. I don't see how deletions of statements highlights such facts are advantageous.

francozappa: > So, maybe the text is slightly vague but I'm not sure. It looks like you're changing the intent of 
> the article, maybe unintentionally. I think the same goes for #809.

Well. maybe is true because the PRs reflect my opinion about the whole closure chapter. I think that it is focusing too much on explaining why a closure is "different" from a function instead of explaining why the language is offering closures as "another type of" functions, and why it is convenient to use them and how. For example the chapter could give more emphasis on the difference between a normal closure and a `move` closure like in the official Rust book.

mdinger: Lots of people don't know what makes a closure a closure. Explaining the details behind it is well worth it. Furthermore, knowing when to use it and what advantages it brings is more difficult. Providing details helps with that.

I'm not sure why `move` needs expanded treatment. The important point is that closures have different capture modes which change on demand. All `move` does is enforce a specific one. Pointing out when it is advantageous to do so is a different topic.

> why the language is offering closures as "another type of" functions

Not sure what you mean by this

> why it is convenient to use them and how

It could probably use expanding in this area

steveklabnik:
Thank you!

PR #815, opened 2016-10-14
error/option_unwrap/map/map.rs: minor print messages

francozappa:


steveklabnik:
Is this actually needed? Since it derives Debug, it's fine to not use .0, and it does compile.

francozappa:
>  Is this actually needed? Since it derives Debug, it's fine to not use .0, and it does compile.

@steveklabnik , true ! 

I've added it to obtain a nicer `println!` output.

steveklabnik:
> I've added it to obtain a nicer println! output.

hmm, interesting. @mdinger do you have opinions?

mdinger:
Opinions on this "looking better" may vary from person to person. I think it is okay that it says `Cooked(Apple)` when it could have said `Peeled(Apple)`; this is hardly vectors nested 4 deep direly in need of cleaner printing.

I don't see any advantage gained from accepting this PR other than pointing out that `.0` works. A newtype provides clarity when printing anyway. I wouldn't merge this but I don't feel strongly about it either.

PR #816, opened 2016-10-15
std/result: Non-positive logarithm

chenl:
Check for non-positive logarithm value rather than negative


chenl:
same change should be made in std/result/try, should I close this PR and create a new one?

steveklabnik:
@chenl adding it to this PR is fine! Thank you

steveklabnik:
I'm going to merge this as-is, @chenl if you want to send a follow-up, please do!

PR #822, opened 2016-11-17
fn/hof: destructure in `filter` like `take_while`

josephfrazier:
This makes it more consistent with the fn/closures/closure_examples.


steveklabnik:
Thank you!

Issue #821, opened 2016-11-17
Usage of std::result in "aliases for Result" chapter


realcr:
In the chapter of aliases for Result there is a use of `std::result` to create a new alias for result.
For reference, I'm talking about this chapter:
http://rustbyexample.com/error/result/result_alias.html

```rust
use std::result;
...
type AliasedResult<T> = result::Result<T, ParseIntError>;
...
```
The `Result` type was used in the previous section ("map for Result") and `std::result` was not mentioned there at all. I was wondering if the statement `use std::result` has any special meaning in the section.

To check myself I changed the above code to the following:
```rust
// use std::result;
...
type AliasedResult<T> = Result<T, ParseIntError>;
...
```
And the code seems to compile and work correctly. Am I missing something here? Is the `Result` I'm using the modified code the same as `std::result::Result` used in the original code?

Except for this question, thank you for writing this book. It is a great addition to "The book", and it helped me to understand many things that I couldn't understand earlier: Especially the subject of lifetimes.

steveklabnik:
You're not missing anything; this project is just very old, and contains some stuff that's not needed today. I'd happily take a PR to remove that line!

PR #825, opened 2016-11-18
Removed unneeded usage of std::result.

realcr:
Fixed issue #821: Usage of std::result in "aliases for Result" chapter.


steveklabnik:
Thank you!

Issue #765, opened 2016-06-17
touch() must also include write(true) or append(true)


casey:
I believe that the touch example is incorrect:

https://github.com/rust-lang/rust-by-example/blob/f4fd0b50c10b337bf6eb7706dea62688a53426de/examples/std_misc/fs/fs.rs#L26

If you look at `std::fs::OpenOptions`, it states that if `create(true)` is specified, then `write(true)` or `append(true)` must be set:

https://doc.rust-lang.org/std/fs/struct.OpenOptions.html#method.create

On OS X, I get this error: `Invalid argument (os error 22)`

realcr:
I am experiencing the same issue on Ubuntu 16.04. Adding the `.write(true)` option solves the problem.

xftroxgpx:
This issue seems to be in need to be closed as per previous merge.
I think github didn't auto-close it because of the extra `issue` word in `Fixes issue #765`

PR #827, opened 2016-11-27
Added missing write(true) option. Fixes issue #765.

realcr:


mdinger:
Looks good to me. Is inline with [openoptions](https://doc.rust-lang.org/std/fs/struct.OpenOptions.html) docs

mdinger:
Though, it should probably mirror the docs main example and be `read` as well:

```rust
use std::fs::OpenOptions;

let file = OpenOptions::new()
            .read(true)
            .write(true)
            .create(true)
            .open("foo.txt");
```

realcr:
@mdinger: The example shows how to implement touch, which doesn't need the `.read` option, so maybe we shouldn't add it here. What do you think?

mdinger:
Isn't the intent of touch to make a new file which is readable?

realcr:
Maybe the person who originally wrote this should comment on this.
I think that the touch function should be some kind of imitation of the Unix touch command.

For example, see here: http://www.linfo.org/touch.html

>  The touch command is the easiest way to create new, empty files. It is also used to change the timestamps (i.e., dates and times of the most recent access and modification) on existing files and directories. 

To me it doesn't sound like touch should be able to read from the file.

steveklabnik:
Yup, I agree @realcr . Thanks!

mdinger:
Oooooh, I was thinking read/write was going to configure the OS read/write permissions, not if the operation could just read the file. Silly me.

PR #837, opened 2017-01-16
Add example struct field init shorthand

hngnaig:


Havvy:
Travis fails on stable and beta where field init shorthand syntax doesn't exist. It succeeds on nightly where it does. I'm presuming this patch will sit here until a release that has field init shorthand syntax in stable?

steveklabnik:
Yes.

steveklabnik:
@hngnaig this syntax is now stable, can you remove the `#![feature]` stuff and then we can merge? Thanks!

steveklabnik:
I've made the change. Thanks!

PR #839, opened 2017-01-25
closures/iter_find: Fix typo in comment

garious:


steveklabnik:
I don't believe this is a typo; the iterator methods return `&`, but filter takes an `&`, which leads to the `&&`.

maybe we can make this more clear somehow?

garious:
I see, thanks! I thought it was a copy-paste bug from the `any` example. No need for clarification.

PR #842, opened 2017-02-09
hello/display: Change 'Point2' to 'Point2D'

### LINKED ISSUE #841: Point2 vs Point2D
Reading through Hello World -> Display, I was a little confused as to why the struct is named `Point2`, since there is no `Point` or `Point1`. However, at the end of that page, it says this:

```rust
 // println!("What does Point2D look like in binary: {:b}?", point);
```
This name is implying it's a 2D point, which makes the name make sense. Would it be acceptable if I changed to struct's name in this book from `Point2` to `Point2D` to help others avoid the confusion?

J3RN:
I had some confusion over the naming of the 'Point2' struct, and whether it was supposed to be named 'Point2D'. This change renames the struct.

Closes #841

### DIFF
--- a/examples/hello/print/print_display/display.rs
+++ b/examples/hello/print/print_display/display.rs
@@ -15,13 +15,13 @@ impl fmt::Display for MinMax {
 
 // Define a structure where the fields are nameable for comparison.
 #[derive(Debug)]
-struct Point2 {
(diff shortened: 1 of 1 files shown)


steveklabnik:
Thanks!

PR #844, opened 2017-02-16
Add field init shorthand example.

chriskrycho:
**Note:** this should *not* be merged until 1.17 lands and the field init shorthand feature lands.

Though this isn't *required* for https://github.com/rust-lang/rust/issues/37340, I'd call it a nice bonus to have it ready to go when it does.

### DIFF
--- a/examples/custom_types/structs/structs.rs
+++ b/examples/custom_types/structs/structs.rs
@@ -33,6 +33,13 @@ fn main() {
         p2: point,
     };
 
+    // Create a struct using field initialization shorthand. When local names
+    // match the names of the public fields of the struct, you don't have to
+    // type them twice!
+    let p1 = 0.1;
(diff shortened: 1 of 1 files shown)


mdinger:
It's pretty nice you're trying to update an external repo other than just normal official targets for this. This seems fairly unusual.

chriskrycho:
It should be increasingly common, since whether a change should make its way into Rust By Example is one of the questions under consideration in writing, accepting, and stabilizing RFCs now that RFC 1636 was merged! Some RFCs may even go so far as to require updating RbE before stabilizing the feature. This one didn't, but it's also an easy win, so here we are!

steveklabnik:
Closing in favor of https://github.com/rust-lang/rust-by-example/pull/837, which was first. Thanks though! ❤️

PR #845, opened 2017-02-16
print/print_display: changing 'Point2' to 'Point2D' under Activity

projektir:
I think this last part was missed for issue #841.


steveklabnik:
Nice catch, thanks!

PR #849, opened 2017-03-06
generics/gen_fn: Fix typo

Gordon-F:


steveklabnik:
I don't think this is a typo; they're both plural: functions and structs. Right?

Gordon-F:
Right. But, look at the other pages, eg: [Traits](http://rustbyexample.com/generics/gen_trait.html), [Implementation](http://rustbyexample.com/generics/impl.html).

Character after `` looks little bit ugly. Maybe `structs` ?

steveklabnik:
> Character after `` looks little bit ugly. Maybe structs ?

This is historically how we've done it.

PR #850, opened 2017-03-09
Add definition of 'word'.

russmack:
This commit adds a short definition of "word" in the context of Rust, in the description of slices.


steveklabnik:
Thanks!

PR #853, opened 2017-03-10
hello/print/print_display/testcase_list: replace try! with ?. Close #834

### LINKED ISSUE #834: The try! macro should be replaced with new construct "?"
For example in 1.2.2.1:
`write!(f, "{}", value)?;`

russmack:
This commit replaces the try! macro with the ? operator in 1.2.2.1 hello/print/print_display/testcase_list, and mentions their equivalence.


steveklabnik:
Thank you!

Issue #854, opened 2017-03-15
cast/alias: Confusing wording -> typing 


jeiting:
https://rustbyexample.com/cast/alias.html

"The main use of aliases is to reduce typing; for example the IoResult<T> type is an alias for the Result<T, IoError> type."

Is "typing" here referring to the programming concept or the physical act of typing on a keyboard?

If it refers to the physical act, perhaps a better word would be "boilerplate" or "repetition". Glad to submit a PR if my understanding is correct here.

steveklabnik:
The latter. PR very welcome!

steveklabnik:
https://github.com/rust-lang/rust-by-example/pull/855 should have closed this!

PR #855, opened 2017-03-15
cast/alias: Fix typing ambiguity.

jeiting:
Issue: #854

This caused my a little bit of confusion because I thought maybe typing was some sort of concept that was good to reduce, turns out it just means punching in letters on the keyboard.

Changed typing -> boilerplate.


steveklabnik:
Thank you!

PR #856, opened 2017-03-15
Add pub(restricted) example

cramertj:
See https://github.com/rust-lang/rust/issues/32409
cc @steveklabnik

### DIFF
--- a/examples/mod/visibility/visibility.rs
+++ b/examples/mod/visibility/visibility.rs
@@ -1,39 +1,55 @@
-// A module named `my`
-mod my {
+// A module named `my_mod`
+mod my_mod {
     // Items in modules default to private visibility.
     fn private_function() {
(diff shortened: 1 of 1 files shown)


cramertj:
(Travis is obviously going to fail since the feature isn't stable yet.)

steveklabnik:
Yup! RBE deploys to master always, so we'll have to wait until the release to merge.

cramertj:
To add:
- path in pub(in path) item must be some parent module of the item.
- pub(super) and pub(self) are shortcuts for pub(in super) and pub(in self) respectively
- pub(self) is the same thing as private

steveklabnik:
We'll be able to merge this Thursday; it now is green on beta 🎊

steveklabnik:
Woo!

PR #867, opened 2017-05-13
Add an example about returning values with break

pietroalbini:
This pull request adds an example for the `loop_break_value` feature (which should be stabilized after all the documentation is ready), tracked in issue rust-lang/rust#37339.

The example doesn't do much, but without introducing random things, threads or the network in it there aren't many things to retry.

### DIFF
--- a/examples/flow_control/loop/return/return.rs
+++ b/examples/flow_control/loop/return/return.rs
@@ -1,0 +1,13 @@
+fn main() {
+    let mut counter = 0;
+
+    let result = loop {
+        counter += 1;
+
+        if counter == 10 {
+            break counter * 2;
+        }
+    };
+
+    assert_eq!(result, 20);
+}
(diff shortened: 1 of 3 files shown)


examples/flow_control/loop/return/return.rs
```diff
@@ -0,0 +1,14 @@
...
+    let result = loop {
+        counter += 1;
+
+        if counter == 10 {
+            break counter * 2;
+        }
+    };
+
+    // This should print "The result is 20"
+    println!("The result is {}", result)
```
GuillaumeGomez: Why not putting an assert then?

pietroalbini: Yep, that's better.

GuillaumeGomez:
👍

steveklabnik:
Thank you! This looks great. Now we wait for it to hit stable.

pietroalbini:
The stabilization PR was merged \o/ Should we merge this now?

steveklabnik:
We have to actually wait for it to land in stable rust; RBE doesn't have a train structure.

steveklabnik:
At long last, this has landed. Thanks so much @pietroalbini !

PR #869, opened 2017-05-21
Use [Open|Closed]Box

svenfuchs:
Similar to how `[safe|block]list` is better terminology I think the same concerns apply here.

This commit changes this exapmle to use `[Open|Closed]Box`.

### DIFF
--- a/examples/mod/struct_visibility/struct.rs
+++ b/examples/mod/struct_visibility/struct.rs
@@ -1,19 +1,19 @@
 mod my {
     // A public struct with a public field of generic type `T`
-    pub struct WhiteBox<T> {
+    pub struct OpenBox<T> {
         pub contents: T,
     }
 
     // A public struct with a private field of generic type `T`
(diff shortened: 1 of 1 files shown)


examples/mod/struct_visibility/struct.rs
```diff
@@ -22,22 +22,22 @@ mod my {
 
 fn main() {
     // Public structs with public fields can be constructed as usual
-    let white_box = my::WhiteBox { contents: "public information" };
+    let open_box = my::OpenBox { contents: "public information" };
 
     // and their fields can be normally accessed.
-    println!("The white box contains: {}", white_box.contents);
+    println!("The open box contains: {}", white_box.contents);
```
Gordon-F: s/white_box/open_box

svenfuchs: thanks @Gordon-F! fixed.

steveklabnik:
❤️

PR #870, opened 2017-05-22
std/result/try: Non-positive logarithm

meganetaaan:
Check for non-positive logarithm value rather than negative.
(according to the PR of std/result https://github.com/rust-lang/rust-by-example/pull/816)


steveklabnik:
Thanks!

PR #880, opened 2017-07-15
Fix error response handling for rust playground

AlexRiedler:
In the case of compilation failure, the api returns an object with an empty result; and an error in its own field.
This makes it so this error bubbles back to the requestor that is running the program.

I feel really weird editing this in the node_modules folder; however it seems like this package is not on npm ?

I noticed this when doing https://rustbyexample.com/hello/print.html and running it without any change.

Suggestions welcome! :)

### DIFF
--- a/node_modules/gitbook-plugin-rust-playpen/book/editor.js
+++ b/node_modules/gitbook-plugin-rust-playpen/book/editor.js
@@ -169,11 +169,15 @@ function runProgram(program, callback) {
   req.open('POST', "https://play.rust-lang.org/evaluate.json", true);
   req.onload = function(e) {
     if (req.readyState === 4 && req.status === 200) {
-      var result = JSON.parse(req.response).result;
+      var response = JSON.parse(req.response);
+      var result = response.result;
 
       // Need server support to get an accurate version of this.
       var statusCode = SUCCESS;
-      if (result.indexOf("error:") !== -1) {
+      if (response.error) {
+        statusCode = ERROR;
+        result = response.error;
+      } else if (result.indexOf("error:") !== -1) {
         statusCode = ERROR;
       } else if (result.indexOf("warning:") !== -1) {
         statusCode = WARNING;


steveklabnik:
> I feel really weird editing this in the node_modules folder; however it seems like this package is not on npm ?

It's not, yeah.

However!

I think this is not a bug here, but a bug in the playpen. https://github.com/integer32llc/rust-playground/issues/179

So let's hold off until we see what they say.

shepmaster:
Can you explain a little more what the previous code expected to see and what the change should do? 

The Rust homepage checks for an `error` key **and** the string `error:`:

```js
result = JSON.parse(req.response);

// handle application errors from playpen
if (typeof result['error'] === 'string') {
    statusCode = ERROR;
    result = 'Playpen Error: ' + result['error'];
} else if (typeof result['result'] === 'string') {
    statusCode = SUCCESS;
    result = result['result'];

    // handle rustc errors/warnings
    // Need server support to get an accurate version of this.
    if (result.indexOf("error:") !== -1) {
        statusCode = ERROR;
    } else if (result.indexOf("warning:") !== -1) {
        statusCode = WARNING;
    }
}
```

The first and second editions of the book assume that everything will be in `result`, but don't care about errors:

```
    $.ajax({
        url: "https://play.rust-lang.org/evaluate.json",
        method: "POST",
        crossDomain: true,
        dataType: "json",
        contentType: "application/json",
        data: JSON.stringify(params),
        success: function(response){
            result_block.text(response.result);
        }
    });
```

Looks like this does a mix of the two - assuming that the result will always be in `result` but checking for the string. Sigh.

I don't see any reason to not just duplicate the error text to both `result` and `error` in the playground. This is what we get for not designing a feature in the first place. 😿

steveklabnik:
> Can you explain a little more what the previous code expected to see and what the change should do?

Honestly, I don't remember. I just know this worked and now it's broken.

I thought that the others would break too, as I thought they'd be doing the same thing, but I guess not.

If this is desired behavior, I'm fine just updating RBE.

AlexRiedler:
It appears (based on reading: https://github.com/integer32llc/rust-playground/blob/master/ui/src/main.rs#L424) that the reason it is separated is to show the difference between compilation and runtime errors.

We might be able to change it to just take either or since the http codes do not seem to match.

shepmaster:
> I thought that the others would break too, as I thought they'd be doing the same thing, but I guess not.

I'm guessing that the other cases don't fail, or fail so rarely that no one noticed. The homepage shouldn't fail; any idea if there are any doc pages with code examples that are *supposed* to fail?

> that the reason it is separated is to show the difference between compilation and runtime errors.

This route exists *solely* for backwards compatibility, so if we're doing something on the playground side such that any previous consumer breaks, it should be fixed on the playground.

steveklabnik:
> any idea if there are any doc pages with code examples that are supposed to fail?

I believe `mdbook` drops the play button for anything that's supposed to fail; not 100% sure though.

shepmaster:
This may be closed now. Thanks for the report; sorry for the breakage!

steveklabnik:
@shepmaster no worries! <3

PR #883, opened 2017-07-31
std_misc/path: traits outdated

Songbird0:
The `BytesContainer` trait and `FileStat` struct aren't used anymore. I've replaced `BytesContainer` with `OsStr` and `FileStat` with `Metadata`.

A `See also` header and some links were added too.


steveklabnik:
Thanks!

PR #887, opened 2017-08-14
fix typo in examples/trait/ops/input.md

d-e-s-o:
'list' is singular so 'is' is the correct form of be to pair with it,
not 'are'.


steveklabnik:
Thanks!

PR #893, opened 2017-09-03
fs: add alternative format for ? notation

anna-liao:
I was referring to this example for some Rust code I am writing, and @burntsushi suggested the shorthand notation `?` that was introduced in Rust 1.3.  I thought it would be useful to add that to the `fs` example.

### DIFF
--- a/examples/std_misc/fs/input.md
+++ b/examples/std_misc/fs/input.md
@@ -33,6 +33,17 @@ a
 1 directory, 2 files
 ```
 
+An alternative way to define the function `cat` is with `?` notation:
+
+```
+fn cat(path: &Path) -> io::Result<String> {
+    let mut f = File::open(path)?;
+    let mut s = String::new();
+    f.read_to_string(&mut s)?;
+    Ok(s)
+}
+```
+
 ### See also:
 
 [`cfg!`][cfg]
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+_book/*
 bin/*
 stage/*
 node_modules/*
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -59,3 +59,12 @@ name = "utf8-ranges"
 version = "0.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
+[metadata]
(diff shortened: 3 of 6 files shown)


examples/std_misc/fs/input.md
```diff
@@ -33,6 +33,15 @@ a
 1 directory, 2 files
 ```
 
+An alternative way to define the function `cat` is with `?` notation:
+```
+fn cat(path: &Path) -> io::Result<String> {
+    let mut f = File::open(path)?;
+    let mut s = String::new();
+    f.read_to_string(&mut s).map(|_| s)
```
steveklabnik: this is a bit awkward, I'd prefer

    f.read_to-string(&mut s)?;
    Ok(s)

examples/std_misc/fs/input.md
```diff
@@ -33,6 +33,15 @@ a
 1 directory, 2 files
 ```
 
+An alternative way to define the function `cat` is with `?` notation:
+```
```
steveklabnik: can you put a newline above this line, to give some space between the text and the example?

anna-liao:
@steveklabnik thanks for the review!  I made the changes, but it wasn't clear to me how to run the commands to view the changes locally.  When I ran gitbook serve, it only showed the Introduction page.  Also, with the CI build failure, I received the same error with `Error: Cannot find module 'internal/fs'`.  I'm not sure if this is an error as a result of this addition.

steveklabnik:
Hey @anna-liao ! So, yeah, the build system is very... baroque. Honestly I don't even run it locally, and re-doing it entirely is something that's been on my mind, I just haven't gotten to it yet.

Could you maybe try removing all other changes than just the one to the example? I think maybe that's causing problems, I'm not sure though.

steveklabnik:
Hey there! So, given the above conversation about the build being complex, I did something I've been meaning to do for a long time: I ported it to mdBook! This makes this PR out of date, though. I made https://github.com/rust-lang/rust-by-example/pull/898 to fix it; I've rebased your change, and kept your authorship information. 

Thank you again for the patch!

PR #884, opened 2017-08-01
std_misc/path: bad substitution

Songbird0:
I did a simple substitution for `BytesContainer`, that's an error. `BytesContainer` is a trait while `OsStr` is a struct. The sentence is fixed.

PR #898, opened 2017-09-14
Rebase other open PRs

steveklabnik:
Continuation of https://github.com/rust-lang/rust-by-example/pull/884 and https://github.com/rust-lang/rust-by-example/pull/893


steveklabnik:
I've added one more commit for something else I've found that's unrelated, but since I hadn't merged this one yet, just putting it in this PR.

PR #909, opened 2017-09-19
Fix open subcommands in readme

BORN2LOSE:
Hi. When i type in terminal: `mdbook open` i get some error:
> error: Found argument 'open' which wasn't expected, or isn't valid in this context.

I found a new subcommand that launches book on localhost by default: `serve`.
*Serve the book at localhost:3000. Rebuild and reload on change.*

My mdbook version: `mdbook v0.0.25`.


steveklabnik:
Whoops! Thank you!

Issue #896, opened 2017-09-12
Update try! section to use ?


johnthagen:
https://rustbyexample.com/error/multiple_error_types/enter_try.html

# 16.4.2 Introducing `try!`

I would suggest that this section be updated to use `?`, since that is the common idiomatic Rust way to return an error early now.

Edit: Now that I look, RustByExample has a lot of sections where it is using `try!`, so probably best to hit them all at once for uniformity.

steveklabnik:
Yup!

steveklabnik:
Closed by https://github.com/rust-lang/rust-by-example/pull/912#pullrequestreview-63724054 🎊

PR #912, opened 2017-09-19
Use `?` instead of `try!`

Zengor:
As mentioned in issue #896. I kept some mentions of the `try!` macro as an older variant that might be seen in older Rust.


steveklabnik:
This looks great, thank you!

Issue #589, opened 2015-06-07
long_tuple can't debug


tennix:
In [Tuples chapter](http://rustbyexample.com/primitives/tuples.html), if I added one more element in `long_tuple` and tried to print it with `{:?}`, then it would be an error

`
error: the trait `core::fmt::Debug` is not implemented for the type `(u8, u16, u32, u64, i8, i16, i32, i64, f32, f64, char, bool, bool)` [E0277]
`

In [Debug chapter](http://rustbyexample.com/hello/print/print_debug.html)

> All std library types automatically are printable with {:?} too

I read the `std::fmt::Debug` source code and found there are some limitations for Debug trait. For arrays and tuples, it can't be too long. Since I can't find it documented in trpl or reference, I think we should point it out in rust-by-example

mdinger:
See https://github.com/rust-lang/rust-by-example/issues/127 also.

tennix:
@mdinger yes, there can be long tuples but it can't use Debug trait. If not pointed this out somewhere, people will get confused if they try to print long tuples or long arrays using Debug trait

steveklabnik:
I'd be happy to help mentor someone to fix this bug; we should point out this behavior of tuples for sure.

marioidival:
As @vishalsodani say in PR #915 this is not possible, so this issue can be closed? @steveklabnik

PR #915, opened 2017-09-20
Add an example to demonstrate long tuple cannot be printed

vishalsodani:
An attempt to fix the bug #589 .

### DIFF
--- a/src/primitives/tuples.md
+++ b/src/primitives/tuples.md
@@ -31,7 +31,7 @@ fn main() {
 
     // Tuples can be tuple members
     let tuple_of_tuples = ((1u8, 2u16, 2u32), (4u64, -1i8), -2i16);
-    
+
     // Tuples are printable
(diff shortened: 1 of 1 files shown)


src/primitives/tuples.md
```diff
@@ -31,9 +31,14 @@ fn main() {
 
     // Tuples can be tuple members
     let tuple_of_tuples = ((1u8, 2u16, 2u32), (4u64, -1i8), -2i16);
-
+
```
steveklabnik: ✂️

vishalsodani:
Why does this temporary restriction exists?

steveklabnik:
> Why does this temporary restriction exists?

Because Rust doesn't have a way to write a function generically over integers, which would be needed to implement a trait for tuples of arbitrary lengths, which has an RFC accepted but not yet implemented.

(Changes look good!)

srikwit:
We still don't have this implemented. Any suggestions on how this can be done?

PR #917, opened 2017-09-21
Fix small typo in Vec chapter

Zengor:
Lack of comma on this line would make example not compile.


steveklabnik:
Thanks so much! :metal:

Issue #832, opened 2017-01-10
default type for number is i64 not i32 (Chapter 2 Primitives)


jido:
In the example, I can assign an i64 value to "mutable" as follows:
```
    let mut mutable = 12;

    mutable = 3000000000000000000i64;
```

However there is a type error if I force the type of mutable to i32:
```
    let mut mutable: i32 = 12; // Mutable `i32`.

    // Error! The type of a variable can't be changed
    mutable = 3000000000000000000i64;
```
This indicates that integers default to i64.

steveklabnik:
This isn't true; "unconstrained" numbers will be u32, but given that you said `300000....i64`, you provide a constraint that makes it `i64`. If you'd have left that off, then the type would be `i32`.

Any idea on how to word this so that it makes more sense?

jido:
I may have misinterpreted the results but I really believe that integers defaults to i64 based on my testing.

I updated the issue description to explain better.

steveklabnik:
Yes, I understand your issue. My comment was trying to explain why your interpretation is incorrect.

ubsan:
@jido 

Note that this also works:

```rust
let mut x = 0;
x = 0u8;
let mut y = 1;
y = 1u16;
// and similarly, with all the integer types
```

jido:
@ubsan @steveklabnik 
does that mean that the integer variable is unconstrained, but an integer literal defaults to i32?

In that case when is the default literal type i32 ever relevant since it is not applied to the variable in case of mut and it is resized accordingly when assigning to a different byte length variable?

`let mut x = 10 // x is an unconstrained integer (not i32)`
`let y: i16 = 5 // 5 is converted to i16 on assignment`

I suggest removing any reference to i32 by default, instead say that the type is unconstrained integer when declared that way.

ubsan:
@jido for example, `println!("{}", 13);` shall make that `13` an `i32`. That is the type of default it is talking about.

jido:
@ubsan so when it is assigned to a non-mutable variable, the type is i32 and when it is assigned to a mutable variable the type is unconstrained integer? Is this observable in any way?

jido:
This answers the "observable" although I don't know if there are practical uses:
```
let x = 30;
println!("x occupies {} bytes", mem::size_of_val(&x));
```

mdinger:
Sounds like you're confusing how rust which is strongly typed (strictly?)
with other languages which have things like 'objects' which are coercible
[...]

Issue #783, opened 2016-07-31
"The type of a variable can't be changed" - but could maybe mention shadowing?


icanrealizeum:
This `mutable = true;`
here: 
http://rustbyexample.com/primitives.html

Makes me wish that `let mutable = true;` was mentioned as a way to still be able to change the type of the variable by shadowing the old one. I think it would be interesting to have in that context. 

Thanks for your consideration.

steveklabnik:
Yes, shadowing is important, and would be a great addition; possibly adding its own section!

I'm happy to mentor anyone who wants to fix this.

zsck:
Given that there appears to now be a whole section that (briefly) discusses binding [scope and shadowing](https://rustbyexample.com/variable_bindings/scope.html), would it be more appropriate to simply add a link to that section?  I personally feel that having a separate section for these things is perfectly reasonable. The concepts deserve at least a little bit of dedicated attention, and I'm not sure we necessarily need to tie the explanation of primitives to the technique of shadowing. In light of that, maybe this issue can be closed?

steveklabnik:
Yeah, that seems good to me.

jansol:
Link and example are in place, please close this.

PR #920, opened 2017-09-22
Improve primitives section

dwbrite:
Fix/Improve: #783, #832

- Distinguish scalar and compound types
- Add an example of inference
- Improve example of mutability
- Add an example of shadowing
- Add `mut` to "see also:"
- Add shadowing to "see also:"
- Add inference to "see also:"

### DIFF
--- a/src/primitives.md
+++ b/src/primitives.md
@@ -10,13 +10,15 @@ Rust provides access to a wide variety of `primitives`. A sample includes:
 * floating point: `f32`, `f64`
 * `char` Unicode scalar values like `'a'`, `'α'` and `'∞'` (4 bytes each)
 * `bool` either `true` or `false`
+* and the unit type `()`, whose only possible value is an empty tuple: `()`
 
+Despite the value of a unit type being a tuple, it is not considered a
+compound type because it does not contain multiple values. 
 
 ### Compound Types
 
 * arrays like `[1, 2, 3]`
 * tuples like `(1, true)`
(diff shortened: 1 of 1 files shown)


src/primitives.md
```diff
@@ -2,18 +2,25 @@
...
 
+
+### Scalar Types
+
 * signed integers: `i8`, `i16`, `i32`, `i64` and `isize` (pointer size)
 * unsigned integers: `u8`, `u16`, `u32`, `u64` and `usize` (pointer size)
 * floating point: `f32`, `f64`
 * `char` Unicode scalar values like `'a'`, `'α'` and `'∞'` (4 bytes each)
 * `bool` either `true` or `false`
-* and the unit type `()`, whose only value is also `()`
```
steveklabnik: So, what's interesting here is that `()` *is* a scalar type...

src/primitives.md
```diff
@@ -2,18 +2,25 @@
...
+
+
+### Compound Types
+
 * arrays like `[1, 2, 3]`
 * tuples like `(1, true)`
+* and the unit type `()`, whose only value is also `()`
```
steveklabnik: ... so having it here doesn't make a ton of sense. I get why it may feel like it, given the line above is tuples generally, but I'm not sure that we should move this.

dwbrite:
Updated to clarify that change.

steveklabnik:
Thanks so much, and sorry it took me a while to review!

PR #922, opened 2017-09-24
Editable any

mdinger:
Was this intended? The change is untested.


mdinger:
Sorry, gives [this](https://rustbyexample.com/fn/closures/closure_examples/iter_any.html) page an editable example.

steveklabnik:
It may have been a mistake from the port, or maybe not.

Issue #918, opened 2017-09-21
Links to the first edition of rust book are dead


vishalsodani:
On some pages like [Error Handling](https://rustbyexample.com/error.html) and [Combinators](https://rustbyexample.com/error/option_unwrap/map.html), the links https://doc.rust-lang.org/book/error-handling.html and https://doc.rust-lang.org/book/glossary.html#combinators to the first edition of rust book do not work.

steveklabnik:
Those pages do link to the new page, but we should update these for sure!

maccoda:
I have had a quick look over the remaining references to the first book and am having a little difficulty finding the appropriate mapping of the following for the second book:
* [doc.md](https://github.com/rust-lang/rust-by-example/blob/master/src/meta/doc.md) - Documentation
* [testcase_mapreduce.md](https://github.com/rust-lang/rust-by-example/blob/master/src/std_misc/threads/testcase_mapreduce.md) - Loops: Enumerate
* [map.md](https://github.com/rust-lang/rust-by-example/blob/master/src/error/option_unwrap/map.md) - Glossary: Combinators

Any ideas on these ones?

marioidival:
[Error Handling](https://rustbyexample.com/error.html) has link to second version of book, [Combinators](https://rustbyexample.com/error/option_unwrap/map.html) has link but the user have choice a version of book, can be consider broken too... @steveklabnik

marioidival:
@QuietMisdreavus

peter-kehl:
https://doc.rust-lang.org/book/index.html?search=combinator doesn't show up any matches for "combinator" or "combinators" (only for "combination" - is must be doing an approximate/Elasticsearch). However, `grep -ri combinator .` on master finds many usages of the word, but again: no definition.

Related description-like occurrences are in:

- error/option_unwrap/map.md (as mention above)
- error/multiple_error_types/option_result.md
- error/result/enter_question_mark.md.

marioidival:
@peter-kehl 
![Capturar](https://user-images.githubusercontent.com/1129263/67959654-d9ac5100-fbd7-11e9-9742-999be6ab4a3a.PNG)

marioidival:
Can I close this issue?

PR #930, opened 2017-10-03
Fix some first edition links

maccoda:
Partially addresses #918, with some remaining links specified in the issue. Those mentioned don't have any clear mapping in the second book that I could determine.


steveklabnik:
Thank you!

PR #933, opened 2017-10-05
Update cast.md

mbaeten:
Thanks for the great tutorial ;) I think this should be u8?


steveklabnik:
Yup, looks good. Thank you!

Issue #106, opened 2014-06-17
Topics for the "args" chapter


japaric:
- [ ] ~~~`os::args`~~~: return type, first arg is the program name, number of arguments
- [x] ~~~matching on `os::args`~~~
- [x] `getopts`

---

@vks is working on this

steveklabnik:
I'd be happy to mentor anyone who wants to work on this!

vks:
I think `getopts` was moved out of the standard library. Maybe hint at `clap` and friends instead?

steveklabnik:
Or, maybe referring to the Cookbook for "real" argument parsing.

maccoda:
In the initial issue comment, does `os::args` now refer to the `ArgsOs` iterator?

steveklabnik:
Yup! https://doc.rust-lang.org/stable/std/env/fn.args.html and its return type

steveklabnik:
I'm not sure what the second checkbox is, so I just checked it off.

maccoda:
@steveklabnik awesome, thanks will work on the next PR! :smile:

maccoda:
@steveklabnik Having a look at the current [args page](https://rustbyexample.com/std_misc/arg.html) it looks like the issue has already been addressed... It describes how to get the first argument and the number of arguments. Is there something else we should be adding?

steveklabnik:
You're totally right, looks great to me. Thanks for closing this out!

PR #934, opened 2017-10-05
Added rust cookbook reference for args

maccoda:
This is to address the last point in #106. I am more than happy to address the other two points before we merge this pull request, just not sure of what `os::args` refers to. Given the date of the issue I am guessing it may have been an older standard library structure. Let me know what other changes I can add to that :smile:

### DIFF
--- a/src/std_misc/arg.md
+++ b/src/std_misc/arg.md
@@ -1,7 +1,9 @@
 # Program arguments
 
+## Standard Library
+
 The command line arguments can be accessed using `std::env::args`, which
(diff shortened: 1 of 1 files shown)


steveklabnik:
> just not sure of what os::args refers to.

Yeah, that's `std::env::args` nowadays.

steveklabnik:
This looks great on its own, thanks so much! Sorry for the long time on the review here. I'd love PRs for the other two, now that we've straightened out what it is 😄

Issue #563, opened 2015-05-02
Update for loop explanation to explain into iterator and such


mdinger:
In [users.rust-lang.org](https://users.rust-lang.org/t/for-loops-and-ownership/1197) there is confusion over how a for loop works. [for](http://rustbyexample.com/flow_control/for.html) doesn't really explain it at all. This could use being improved.

steveklabnik mentions does some [explanation](http://www.reddit.com/r/rust/comments/324v9q/fearless_concurrency/cq833p4) on reddit over terms:
- `iter()` : yields `&T`s
- `iter_mut()` : yields `&mut T`s
- `into_iter()`: yields `T`s

[`IntoIterator`](http://doc.rust-lang.org/nightly/std/iter/trait.IntoIterator.html) really needs explaining too.

steveklabnik:
https://doc.rust-lang.org/std/iter/#for-loops-and-intoiterator is probably a good basis for this work.

I'm happy to mentor anyone who wants to work on this!

maccoda:
@steveklabnik I would be keen to get onto this one. Do you think this warrants its own page? The current for loop page really only discusses looping over ranges and the iterator page obviously focuses more on the trait itself, with a mention of how for loops interact with it. To address this I think we kind of need to sit in the intersection of these. What do you think?

steveklabnik:
https://github.com/rust-lang/rust-by-example/pull/935 fixed this!

PR #935, opened 2017-10-05
Adding alternate into iter functions

maccoda:
Hopefully this reads well and addresses #563 in discussing the different types of iterators for a collection.


steveklabnik:
Yup, looks great, thank you!

PR #939, opened 2017-10-11
Add newtype idiom section

### LINKED ISSUE #174: Cover the newtype idiom
This idiom allows really nice function signatures with zero cost abstraction. For example, see these [function signatures](https://github.com/japaric/criterion.rs/blob/master/src/statistics/sample.rs#L49) I came up with to do some statistics operations - it reminds me of readable Haskell function signatures like `compute :: Sample -> Statistic -> Float`.

Also the newtype idiom can be used with phantom types for extra type safety (like unit checking). So it may be a good idea to introduce this idiom before phantom types.

cc #145 #151 
cc @mdinger

gurgalex:
Fixes #174

### DIFF
--- a/src/generics/new_types.md
+++ b/src/generics/new_types.md
@@ -1,4 +1,5 @@
 # New Type Idiom
+
 The `newtype` idiom gives compile time guarantees that the right type of value is supplied
 to a program.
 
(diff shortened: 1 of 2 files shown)


src/generics/new_types.md
```diff
@@ -0,0 +1,50 @@
+# New Type Idiom
+The `newtype` idiom gives compile time guarantees that the right type of value is supplied
```
steveklabnik: could you add a blank line above this, please?

PR #947, opened 2017-10-21
Add subchapter about (byte)string literals

### LINKED ISSUE #117: Cover string escaping and raw strings in the "literals" chapter
Add this as a "strings" topic.

Might as well cover the new byte literals.

jansol:
Closes #117

### DIFF
--- a/src/std/str.md
+++ b/src/std/str.md
@@ -57,5 +57,110 @@ More `str`/`String` methods can be found under the
 [std::string][string]
 modules
 
+## Literals and escapes
+
+There are multiple ways to write string literals with special characters in them.
+All result in a similar `&str` so it's best to use the form that is the most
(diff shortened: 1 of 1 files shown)


src/std/str.md
```diff
@@ -57,5 +57,109 @@ More `str`/`String` methods can be found under the
 [std::string][string]
 modules
 
+## Literals and escapes
+
+There are multiple ways to write string literals with special characters in them.
+All result in a similar `&str` so it's best to use the form that is the most
```
steveklabnik: This isn't true, as you note below; byte strings give you arrays

jansol: Added a note about it here as well.

steveklabnik:
Thank you!

Issue #550, opened 2015-04-17
The purpose of macros is unclear


mdinger:
Reading the [macros](http://rustbyexample.com/macros.html) section explains how they work but their purpose when compared to functions [isn't obvious](http://users.rust-lang.org/t/newbie-why-macros-vs-functions/1012). It would be good to more explicitly state what advantages macros offer when compared with things like functions.

steveklabnik:
For sure. I'd be happy to mentor anyone who wants to work on this issue!

sonali0901:
@steveklabnik I'd like to work on this issue if you are available to mentor. I will list down some general reasons/advantages why macros are used over functions. I can use some help in context of macros in rust specifically.

steveklabnik:
@sonali0901 that's great! Let's do it 👍 

>  I can use some help in context of macros in rust specifically.

Macros in Rust tend to be used for a few things:

1. DRYing up your code. If you have similar implementations for many types, for example, writing a macro to do the impl can remove a lot of boilerplate.
2. DSLs. Similar to 1, but generating more complex code from simpler code.
3. APIs that need variadric arguments, that is, some calls have a different number of arguments than other calls. This by itself isn't usually a good reason to write a macro, but it does exist.

That's all I can think of off the top of my head, and only really 3 is rust-specific.

jansol:
See https://github.com/mgattozzi/github-rs/blob/master/docs/design.md for a good example of DSL macros.

mark-i-m:
@sonali0901 Are you still working on this? If not, I could give it a go 😄

sonali0901:
@mark-i-m I really got caught into things and almost forgot it. If you wish to work on it, go on. I did gather information about macros in Ruby but it will take me time to document it, Google code-in coming up. Thanks for taking this up.

mark-i-m:
@sonali0901 

> about macros in Ruby

Why ruby?

mark-i-m:
@steveklabnik  I think this can be closed now that #948 was merged...

marioidival:
@steveklabnik I think this can be closed now that #948 was merged...

steveklabnik:
Thanks @marioidival !

I've noticed you've done a bunch of triage work here; the docs team wants to try to get a little group together to work on RBE; if you're interested, please let me know!

marioidival:
@steveklabnik Yes, of course!

steveklabnik:
[...]

PR #948, opened 2017-10-28
Update macros chapter to address why/when to use

mark-i-m:
This is intended to address #550.

Any feedback is appreciated, ~but keep in mind that it is currently WIP~ :smile: 

EDIT: Not WIP anymore


mark-i-m:
@sonali0901 
@steveklabnik 

I think I have addressed the concerns of #550... Any comments?

mark-i-m:
Also @mdinger

PR #949, opened 2017-10-29
Add a chapter on Cargo

### LINKED ISSUE #215: Suggestion: Add a Cargo chapter
I think it would be interesting to go through the basics of using an external library and generating a package of sorts.

mark-i-m:
Fix #215

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -84,6 +84,7 @@
 
 - [Cargo](cargo.md)
     - [Dependencies](cargo/deps.md)
+    - [Conventions](cargo/conventions.md)
     - [Tests](cargo/test.md)
 
 - [Attributes](attribute.md)
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -82,6 +82,10 @@
     - [Library](crates/lib.md)
     - [`extern crate`](crates/link.md)
 
+- [Cargo](cargo.md)
+    - [Dependencies](cargo/deps.md)
+    - [Tests](cargo/test.md)
+
 - [Attributes](attribute.md)
     - [`dead_code`](attribute/unused.md)
     - [Crates](attribute/crate.md)
--- a/src/cargo.md
+++ b/src/cargo.md
@@ -1,0 +1,12 @@
+# Cargo
+
+`cargo` is the official Rust package management tool. It has lots of really
+useful features to improve code quality and developer velocity! These include
+
+- Dependency management and integration with [crates.io](https://crates.io) (the
+  official Rust package registry)
+- Awareness of unit tests
+- Awareness of benchmarks
+
+This chapter will go through some quick basics, but you can find the
+comprehensive docs [here](http://doc.crates.io/index.html).
--- a/src/cargo/conventions.md
+++ b/src/cargo/conventions.md
@@ -1,0 +1,36 @@
+# Conventions
+
+In the previous chapter, we saw the following directory heirarcy:
+
+```txt
+foo
+├── Cargo.toml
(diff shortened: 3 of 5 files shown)


mark-i-m:
@steveklabnik Thanks :)

I guess the biggest question is what should we include in the cargo chapter?
- Do we want to target consumers of crates only or producers too? This would determine whether we want to add a subsection on `cargo publish` and `crates.io`
- Are we including unstable/nightly features? In this case, maybe we might mention
   - Nightly cargo + unstable cargo features
   - `cargo bench`
   - Integration with other build systems

I'm not sure to what extent anyone learning Rust would need to know about any of these, though. The risk is that we overwhelm beginners with details that aren't that useful to learning the language itself. 

Also, I have almost no experience or knowledge of any of them, so I am not sure how qualified I am write about them.

What do you think?

mark-i-m:
ping @steveklabnik @jbiason @mquandalle @neglectedvalue @arthurprs

Thanks in advance :)

steveklabnik:
I'm not going to be able to give you a good reply to the broader question for a few days; I'm travelling at the moment. What we have here looks great though :)

steveklabnik:
I'm back!

> I guess the biggest question is what should we include in the cargo chapter?

RBE is mostly about learning the language, so 

> Do we want to target consumers of crates only or producers too? 

We should be targeting consumers, and then saying "hey if you want to make a crate, it's easy! go check out the cargo book."

> Are we including unstable/nightly features?

No, we cover stable only.

> I'm not sure to what extent anyone learning Rust would need to know about any of these, though. The risk is that we overwhelm beginners with details that aren't that useful to learning the language itself.

Yup, absolutely! So in my mind, it should cover the basics. I'm just going to list stuff out from memory without re-reading your PR:

* `cargo build`
* basic conventions of the top-level directories
* dependencies
  * crates.io and deps fro it
  * git deps
  * path deps
  * override and `patch` or whatever the latest way to do this is, I forget which one
* workspaces
* profiles
* the `-sys` crate convention from a usage perspective, which leads you into:
* build scripts

thoughts? 

> Also, I have almost no experience or knowledge of any of them, so I am not sure how qualified I am write about them.

I'm happy to take even just the start of things, and we can make issues for the others!

mark-i-m:
@steveklabnik Thanks for the response!

You pointed out a lot of gaps in my chapter so far, which I will start filling in :)

> basic conventions of the top-level directories

Could you elaborate on what you mean by this?

> I'm happy to take even just the start of things, and we can make issues for the others!

Yeah, I think that would be good. It might also be nice to have a tracking issue with some sort of checklist. I think I can get most of the points you listed in this PR, but there are a few I still don't know much about (e.g. workspaces, profiles, -sys crates, override/patch/replace).

Thanks again 👍

steveklabnik:
> Could you elaborate on what you mean by this?

yeah like basically this stuff: http://doc.crates.io/book/guide/project-layout.html (this is the upcoming re-organization of cargo's docs into mdbook format)

mark-i-m:
Oh, got it :)

Just a heads up: Things have gotten rather busy for me lately, so I might be a bit slow on this...

mark-i-m:
Hi @steveklabnik! I'm back. I think I have covered a coherent subset of the things you listed. I also created a tracking issue for the rest in #978 . I think this initial portion is ready for a final review before merge.

Let me know what you think.

Merry Christmas :star:

mark-i-m:
@steveklabnik ping :)

steveklabnik:
Sorry for taking so long, this is great, thank you :)

PR #950, opened 2017-10-30
Add basic error handling strategy guide

### LINKED ISSUE #924: Error strategies section
The error chapter gives the developer a lot of tools but doesn't compare when they might be used.

I'm imagining this at the end of the chapter.  My rough outline
- `unwrap` / `panic` for rapid prototyping or high-locality pre-condition tests (see partition approach to Iter Result)(
- `expect` / `panic` for pre-condition tests
- `Option`
  - optional value
  - obvious why the value doesn't exist (see Path.parent
  - Return value can be ignored
- `Result`
  - Result of an operation that must be handled by the caller
  - `Result<Box>` also for rapid prototyping

Related resources
- https://doc.rust-lang.org/std/macro.panic.html
- https://web.archive.org/web/20140717204300/http://doc.rust-lang.org/std/result/#result-and-option
- https://www.reddit.com/r/rust/comments/2j0k21/rust_option_vs_result/?st=j7wrnocn&sh=7869f07f

jansol:
Quick draft for a rule of thumb for how Option and error situations should be handled.

I'm not entirely sure if this fits here or even contains the most noteworthy facts. Additionally I'm leaning towards replacing this with one or two short flowcharts. Is there a sanctioned format for images like that here? (SVG, graphviz, etc?) Or is that completely out of question?

Closes #924

### DIFF
--- a/src/error.md
+++ b/src/error.md
@@ -5,6 +5,26 @@ example, failing to read a file and then continuing to use that *bad* input
 would clearly be problematic. Noticing and explicitly managing those errors
 saves the rest of the program from various pitfalls.
 
+There are various ways to deal with errors in Rust, which are described in the
+following subchapters. They all have more or less subtle differences and different
+use cases. As a rule of thumb:
+
+An explicit `panic` is mainly useful for tests and dealing with unrecoverable errors.
+For prototyping it can be useful, for example when dealing with functions that
+haven't been implemented yet, but in those cases the more descriptive `unimplemented`
+is better. In tests `panic` is a reasonable way to explicitly fail.
+
+The `Option` type is for when a value is optional or when the lack of a value is
+not an error condition. For example the parent of a directory - `/` and `C:` don't
+have one. When dealing with `Option`s, `unwrap` is fine for prototyping and cases
+where it's absolutely certain that there is guaranteed to be a value. However `expect`
+is more useful since it lets you specify an error message in case something goes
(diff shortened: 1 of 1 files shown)


steveklabnik:
> Is there a sanctioned format for images like that here? (SVG, graphviz, etc?) Or is that completely out of question?

We use graphviz in the book, svg also works.

src/error.md
```diff
@@ -5,6 +5,27 @@ example, failing to read a file and then continuing to use that *bad* input
 would clearly be problematic. Noticing and explicitly managing those errors
 saves the rest of the program from various pitfalls.
 
+There are various ways to deal with errors in Rust, which are described in the
+following subchapters. They all have more or less subtle differences and different
+use cases. As a rule of thumb:
+
+An explicit `panic` is mainly useful for tests and for errors which would prevent
+the program from proceeding at all. For prototyping it can be useful, for example
```
steveklabnik: You may want to mention the "unrecoverable" word here, that is what "prevent from proceeding at all" means, but it's the specific term we tend to use in other docs.

jansol:
Fixed.

I'll make a separate PR for the flowchart if I have time to make one. (unless someone beats me to it, of course)

jansol:
Fixed.

I'll make a separate PR for the flowchart if I have time to make one. (unless someone beats me to it, of course)

PR #951, opened 2017-10-30
Complete sample code block so they can build & run

urashid64:


steveklabnik:
This already should build and run, as they're not marked `ignore`. How are you seeing errors here? Remember that rustdoc inserts a main function for you.

urashid64:
Using the 'play' button on the first sample code block on the "for and range" page worked fine, but the other three resulted in the following error:

---------------
   Compiling playground v0.0.1 (file:///playground)
error: expected item, found `let`
 --> src/main.rs:1:1
  |
1 | let names = vec!["Bob", "Frank", "Ferris"];
  | ^^^

error: aborting due to previous error

error: Could not compile `playground`.

To learn more, run the command again with --verbose.
-------------

After I added the fn main {} block around it (similar to the first sample code block), play button worked without errors.

steveklabnik:
I cannot reproduce this error, maybe it was a temporary bug in the playground. All of the buttons work well for me. Thanks for reporting, but since the tests are passing and it works, and this is the appropriate style, I'm going to leave it as-is.

PR #952, opened 2017-10-30
Fixes #890

### LINKED ISSUE #890: Mention that crate_type attribute is ignored when using Cargo
<https://rustbyexample.com/attribute/crate.html> is the first search hit when I look for information on the `crate_type` attribute. Unfortunately, that site fails to mention that the attribute has *no effect whatsoever* when using Cargo.

Given that this means the attribute is hardly ever useful, I think this is worth mentioning.

outkaj:
How does this look as an improvement to the issue mentioned in #890? Thanks a lot!

### DIFF
--- a/src/attribute/crate.md
+++ b/src/attribute/crate.md
@@ -4,10 +4,10 @@ The `crate_type` attribute can be used to tell the compiler whether a crate is
 a binary or a library (and even which type of library), and the `crate_name`
 attribute can be used to set the name of the crate.
 
-However, it is important to note that the `crate_type` attribute has **no**
-effect whatsoever when using Cargo, the Rust package manager. Since Cargo is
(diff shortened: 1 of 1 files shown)


mark-i-m:
Is the same true for `crate_name`? If so, it might be worth mentioning too.

outkaj:
[It looks to me like the same is not true for `crate_name`](https://dumindu.gitbooks.io/learning-rust/content/04.%20cargo,_crates_and_basic_project_structure.html), but could someone else confirm? Thanks for the question @mark-i-m

steveklabnik:
I believe it's true, yes. Would you mind adding it as well? :)

outkaj:
Not at all!

steveklabnik:
This is great, thanks! After travis is green, I'll merge. Should be fine :)

PR #960, opened 2017-11-21
Add section about diverging functions

### LINKED ISSUE #146: Add examples of diverging functions
Explain why we can do these "weirds" things:

`let a: int = return`

`let y: int = match x { true => 1, false => break };`

msehnout:
Fix #146

### DIFF
--- a/src/fn/diverging.md
+++ b/src/fn/diverging.md
@@ -1,0 +1,69 @@
+# Diverging functions
+
+Diverging functions never return. They are marked using `!`, which is an empty type.
+
+```rust
+fn foo() -> ! {
+    panic!("This call never returns.");
+}
+```
+
+As opposed to all the other types, this one cannot be instantiated, because the
+set of all possible values this type can have is empty. Note, that it is
+different from the `()` type, which has exactly one possible value.
+
+For example, this functions returns as usual, although there is no information
+in the return value.
+
+```rust
+fn some_fn() {
+    ()
+}
+
+fn main() {
+    let a: () = some_fn();
+    println!("This functions returns and you can see this line.")
+}
+```
+
+As opposed to this function, which will never return the control back to the caller.
+
+```rust,ignore
+#![feature(never_type)]
+
+fn main() {
+    let x: ! = panic!("This call never returns.");
+    println!("You will never see this line!");
+}
+```
+
+Although this might seem like an abstract concept, it is in fact very useful and
+often handy. The main advantage of this type is that it can be cast to any other
+one and therefore used at places where an exact type is required, for instance
+in `match` branches. This allows us to write code like this:
+
+```rust
+fn main() {
+    fn sum_odd_numbers(up_to: u32) -> u32 {
+        let mut acc = 0;
+        for i in 0..up_to {
+            // Notice that the return type of this match expression must be u32
+            // because of the type of the "addition" variable.
+            let addition: u32 = match i%2 == 1 {
+                // The "i" variable is of type u32, which is perfectly fine.
+                true => i,
+                // On the other hand, the "continue" expression does not return
+                // u32, but it is still fine, because it never returns and therefore
+                // does not violate the type requirements of the match expression.
+                false => continue,
+            };
+            acc += addition;
+        }
+        acc
+    }
+    println!("Sum of odd numbers up to 9 (excluding): {}", sum_odd_numbers(9));
+}
+```
+
+It is also the return type of functions that loop forever (e.g. `loop {}`) like
+network servers or functions that terminates the process (e.g. `exit()`).
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -70,6 +70,7 @@
             - [Iterator::any](fn/closures/closure_examples/iter_any.md)
             - [Iterator::find](fn/closures/closure_examples/iter_find.md)
     - [Higher Order Functions](fn/hof.md)
+    - [Diverging functions](fn/diverging.md)
 
 - [Modules](mod.md)
     - [Visibility](mod/visibility.md)


msehnout:
Strange, the fail seems to be in this code:
```rust
fn main() {
    fn some_fn() -> () {
        return ()
    }

    let a: () = some_fn();
    println!("This functions returns and you can see this line.")
}
```
but it works for me with rustc 1.21, any ideas?

steveklabnik:
That is...... very odd. Let me restart it, maybe it's flaky or something? No idea.

msehnout:
For some reason works like this:
```rust
fn some_fn() {
    ()
}

fn main() {
    let a: () = some_fn();
    println!("This functions returns and you can see this line.")
}
```

projektir:
This is a really cool example and I love seeing this! Given that I don't actually understand diverging functions I gave it a read to try to, well, understand them, so I have a few comments:

The never type and it's relationship to the unit type strike me as fairly advanced topics. [also, `!` is not yet a type in stable Rust], and it seems the terms used here are not necessarily very precise (and I'm not sure how necessary it is to get into all these details in an RBE example). I.e., `!` is considered "uninhabited" as opposed to "empty", it seems, and also it's not the only "uninhabited" type out there.

> Diverging functions never return. Their return type in Rust is denoted by `!`, which is an empty type.

I feel this is tough to read. You say it never returns, but then it has a return type. I get what this means, but maybe something like this would read better:

> Diverging functions never return. They are specified/marked by using ! in place of the return type.

Maybe provide a `fn foo() -> !` example along with it.

I had trouble understanding the main example used. I imagine this is the relevant piece of code: 

```
let addition: u32 = match i%2 == 1 {
   true => i,
   false => continue,
};
```

Is it the `false => continue` portion? I think it's important to mention that not just _functions_, but also *expressions*, can be divergent, especially since the final example you use uses expressions. I think this should be very clearly specified.  What does that actually do? What gets added to `acc`? I think it would be very helpful to lay out all these small details very explicitly because this may be the first time someone's seeing divergent functions.

msehnout:
Thanks for comments. I will look into them soon.

btw this is the source of information I used: https://github.com/rust-lang/rfcs/blob/master/text/1216-bang-type.md

steveklabnik:
Any update on this PR?

msehnout:
Sorry, I totally forgot about this PR. I submitted a modified version based on the comment above, is it better now?

src/macro/syntax.md
```diff
@@ -0,0 +1 @@
+# Syntax
```
steveklabnik: looks like this was accidentally included

src/fn/diverging.md
```diff
@@ -0,0 +1,63 @@
...
+        acc
+    }
+    println!("Sum of odd numbers up to 9 (excluding): {}", sum_odd_numbers(9));
+}
+```
+
+It is also the return type of functions that loop forever (e.g. `loop {}`) like network servers or functions that terminates the process (e.g. `exit()`).
+
+
+
```
steveklabnik: could you remove these extra blank lines please?

src/fn/diverging.md
```diff
@@ -0,0 +1,63 @@
...
+```rust
+#![feature(never_type)]
+
+fn main() {
+    let x: ! = panic!("This call never returns.");
+    println!("You will never see this line!");
+}
+```
+
+Although this might seem like an abstract concept, it is in fact very useful and often handy. The main advantage of this type is that it can be cast to any other one and therefore used at places where an exact type is required, for instance in `match` branches. This allows us to write code like this:
```
steveklabnik: could you wrap this to 80 columns please?

src/fn/diverging.md
```diff
@@ -0,0 +1,63 @@
...
+}
+
+fn main() {
+    let a: () = some_fn();
+    println!("This functions returns and you can see this line.")
+}
+```
+
+As opposed to this function, which will never return the control back to the caller.
+```rust
```
steveklabnik: could you put whitespace above this line, to separate the paragraph from the code please?

src/fn/diverging.md
```diff
@@ -0,0 +1,63 @@
...
+fn foo() -> ! {
+    panic!("This call never returns.");
+}
+```
+
+As opposed to all the other types, this one cannot be instantiated, because the set of all possible values this type can have is empty. Note, that it is different from the `()` type, which has exactly one possible value.
+
+For example, this functions returns as usual, although there is no information in the return value.
+```rust
```
steveklabnik: could you put whitespace above this line, to separate the paragraph from the code please?

src/fn/diverging.md
```diff
@@ -0,0 +1,63 @@
...
+
+Diverging functions never return. They are marked using `!`, which is an empty type.
+
+```rust
+fn foo() -> ! {
+    panic!("This call never returns.");
+}
+```
+
+As opposed to all the other types, this one cannot be instantiated, because the set of all possible values this type can have is empty. Note, that it is different from the `()` type, which has exactly one possible value.
```
steveklabnik: could you wrap this to 80 columns please?

msehnout:
All requested changes should be fixed now.

msehnout:
I'm starting to dislike Travis. Previous checks went just fine a the last one failed without changing the code. I tried to remove the `#feature[...]` line as it is nightly only, but it is still failing, this time because the feature is not stable yet. Was there any change in the CI?

src/fn/diverging.md
```diff
@@ -0,0 +1,67 @@
...
+
+fn main() {
+    let a: () = some_fn();
+    println!("This functions returns and you can see this line.")
+}
+```
+
+As opposed to this function, which will never return the control back to the caller.
+
+```rust
```
frewsxcv: you can change this to `rust,ignore` so this block doesn't get doc tested, since it uses an unstable feature (and travis ci uses stable rust right now)

frewsxcv: and maybe add that `#[feature(...)]` line back in so readers know this is an unstable featuree

msehnout: Thanks, I didn't know about the `ignore` option. :-)

marioidival:
@msehnout Thank you!

PR #969, opened 2017-12-02
Fix possible typo with 'structs' 

Fantasmos:


Fantasmos:
I see this is other places so i'm assuming it's a design choice

Issue #970, opened 2017-12-03
meta/doc example command throws a dead code warning


sethbattin:
When trying the example at https://rustbyexample.com/meta/doc.html I received the following warning:

```
$ rustc person.rs --crate-type lib
warning: function is never used: `main`
  --> person.rs:38:1
   |
38 | fn main() {
   | ^^^^^^^^^
   |
   = note: #[warn(dead_code)] on by default
```

The second command runs fine (I copied the example code verbatim into `person.rs`):

```
$ rustdoc --test --extern doc="libdoc.rlib" person.rs

running 1 test
test person.rs - Person::new (line 18) ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

Other info:
On Windows 10 Linux subsystem:
```
$ rustc -V
rustc 1.22.1 (05e2e1c41 2017-11-22)
```

I assume there are two easy ways to fix this, either squashing warnings or removing `main`, so I'll try to figure out a patch.  But I've only been writing rust for about 30 minutes so no promises :)

sethbattin:
I found an answer in https://rustbyexample.com/hello/print.html in the form of `#[allow(dead_code)]`

marioidival:
I think it's best to remove the main function, for consistency, because this chapter talk about documentation and the example show how to create a lib. What you think @QuietMisdreavus ?

QuietMisdreavus:
I agree; the best solution would be to remove `fn main` and treat the example as a library to begin with. You can't run tests on binary crates anyway - having a main function there is just confusing.

(On a mostly unrelated note, you could get rid of the `--crate-type lib` on the `rustc` command by adding a `#![crate_type="lib"]` to the sample, right next to the `#![crate_name]` attribute. That would also cement the example as not needing `fn main` in the first place.)

sethbattin:
I read back to see how that fits into the chapter.  Because it's about doc comments (and not compiler arguments), I would agree that it would be an improvement to remove the unrelated cruft related to being a library.  For that matter, I think I recall this was the first page that shows libraries as a thing.  Which was intriguing for my first-time studying, but obviously distracted me from the doc block bits.  :)

Rather than that last paragraph, I would be more interested in how and where I could see the outcome of the docs.  Regardless of libraryhood.

marioidival:
[...]

PR #971, opened 2017-12-03
resolve dead code warning

sethbattin:
fixes  https://github.com/rust-lang/rust-by-example/issues/970
I took my solution from https://rustbyexample.com/hello/print.html, which presumably suffered the same problem.

### DIFF
--- a/src/meta/doc.md
+++ b/src/meta/doc.md
@@ -42,6 +42,7 @@ impl Person {
     }
 }
 
+#[allow(dead_code)]
 fn main() {
     let john = Person::new("John");
 


projektir:
Reading through that issue, it seemed the agreement was to simply remove the main function rather than mark it as dead code?

Maybe it's something else, but I can't even run the example on the page:

```
   Compiling playground v0.0.1 (file:///playground)
error: --crate-name and #[crate_name] are required to match, but `playground` != `doc`
 --> src/main.rs:1:1
  |
1 | #![crate_name = "doc"]
  | ^^^^^^^^^^^^^^^^^^^^^^
```

steveklabnik:
Any updates on this PR?

sethbattin:
The discussion on #970 requested a different solution (removing the main function for the lib example).  I don't know when I would get back to this code; I'll back out and let another patch take my place.

PR #973, opened 2017-12-06
Minor change in activity for fmt::Display

feph:
A generic solution for an output format like `3.3 + 7.2i `is not achievable without a conditional(?) and, IMHO can't be solved by std::fmt alone. Changing the expected output to `3.3 +7.2i` (removing a whitespace) makes this easier, because it's simply the result of using `"{real} {imag:+}i"` as format string.

### DIFF
--- a/src/hello/print/print_display.md
+++ b/src/hello/print/print_display.md
@@ -111,7 +111,7 @@ After checking the output of the above example, use the `Point2D` struct as
 guide to add a Complex struct to the example. When printed in the same
 way, the output should be:
 ```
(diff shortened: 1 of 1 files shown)


marioidival:
@feph I think this change already done in master
https://rustbyexample.com/hello/print/print_display.html
https://github.com/rust-lang/rust-by-example/blob/master/src/hello/print/print_display.md#activity

This can be closed, @steveklabnik

steveklabnik:
It does, but I don't know why github shows a diff then. Let's merge it anyway. Thanks @feph!

PR #979, opened 2017-12-21
Update book.toml

### LINKED ISSUE #963: Live code editor not working
I've tried with chrome, opera and Firefox quantum. It simply works like normal, non-editable text. I can't change anything in the code. This is strange since it used to work wonderfully the last time I tried to use the website.   
I didn't see any other way to ask a question so this seemed like the only way to get a solution.

If it's of use, both the japanese and chinese websites are working fine.

jridgewell:
This specifies the new mdBook syntax (properties under `[book]` table).

It also adds the `editor = "ace"` config, which will fix #963. Once https://github.com/rust-lang-nursery/mdBook/pull/515 is merged, this won't be necessary.

### DIFF
--- a/book.toml
+++ b/book.toml
@@ -1,6 +1,8 @@
+[book]
 title = "Rust By Example"
 description = "A description"
 author = "The Rust Community"
 
 [output.html.playpen]
 editable = true
+editor = "ace"


steveklabnik:
Ah ha! Thank you!

PR #982, opened 2017-12-29
Talk about `unsafe` functions in general instead of just `transmute`

Michael-F-Bryan:
In the `unsafe` chapter it's probably more correct to say that you need `unsafe` when calling `unsafe` functions, instead of just `std::mem::transmute` (which is only one example of an `unsafe` function).

I've also put a bit more emphasis on the fact that `unsafe` means the programmer is taking on the job of ensuring your program is correct.

I also feel like people shouldn't need to know about `transmute` because there is almost always a better way.

### DIFF
--- a/src/unsafe.md
+++ b/src/unsafe.md
@@ -9,7 +9,7 @@ unsafe blocks are used for:
 * dereferencing raw pointers
 * calling a function over FFI (but this is covered in [a previous
   chapter](std_misc/ffi.html) of the book)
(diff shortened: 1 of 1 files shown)


projektir:
> I also feel like people shouldn't need to know about `transmute` because there is almost always a better way.

Agreed, and I'm pretty sure it was removed from TRPL v2 for this reason, as well.

Michael-F-Bryan:
@projektir, should be fixed now.

PR #986, opened 2018-01-10
Made if let page a bit more obvious for beginners

woubuc:
This PR attempts to make it a little bit more obvious that `if let` is used for matching enums, and not just `Option`.

I am currently learning Rust, and I ran into the issue today where in my head I hadn't made the connection yet that `Option` is actually an enum. So I didn't realise that `if let` could be used to match any enum value. This PR simply adds a couple of words to the page that may help someone else in the future get to the point of realisation faster.

### DIFF
--- a/src/flow_control/if_let.md
+++ b/src/flow_control/if_let.md
@@ -59,6 +59,41 @@ fn main() {
 }
 ```
 
+In the same way, `if let` can be used to match any enum value:
+
+```rust,editable
+// Our example enum
+enum Foo {
+    Bar,
+    Baz,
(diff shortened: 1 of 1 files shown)


marioidival:
@woubuc Nice PR, I could add some example using enums?

woubuc:
Good idea @marioidival! I'll add an example tomorrow.

woubuc:
Example added.

marioidival:
@woubuc thank you!
@carols10cents or @steveklabnik you can review this?

PR #987, opened 2018-01-11
Change links from '/foo.html' to 'foo.html'

Havvy:
With mdbook, relative links in any document are relative to the
top level of the book, so there's no need to start them with
'/'s.

There were also incorrect links in macros/syntax.md because of
this - they have been fixed.

There were also incorrect links to the stable Reference. These
have also been updated.

I did not change the doc.rlo links, as even though they could
have been made relative for when RBE is on doc.rlo, we're still
going to keep RBE.com's rendering up until RBE renders on stable
docs, which is going to be awhile. But when RBE.com is just a
redirect, they should be changed to "../std/MOAR" and
"../book/MOAR" and "../reference/MOAR" as appropriate too.

### DIFF
--- a/src/scope/borrow/mut.md
+++ b/src/scope/borrow/mut.md
@@ -56,4 +56,4 @@ fn main() {
 ### See also:
 [`static`][static]
 
-[static]: ../lifetime/static_lifetime.html
+[static]: scope/lifetime/static_lifetime.html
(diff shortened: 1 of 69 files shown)


src/scope/borrow/mut.md
```diff
@@ -56,4 +56,4 @@ fn main() {
 ### See also:
 [`static`][static]
 
-[static]: ../lifetime/static_lifetime.html
+[static]: scrope/lifetime/static_lifetime.html
```
QuietMisdreavus: `scrope`?

Havvy: Fixed.

PR #989, opened 2018-01-13
Fixing links

projektir:
Fixing links for rust-lang/rust#46194.

### DIFF
--- a/src/error/result/enter_question_mark.md
+++ b/src/error/result/enter_question_mark.md
@@ -9,7 +9,7 @@ Upon finding an `Err`, there are two valid actions to take:
 1. `panic!` which we already decided to try to avoid if possible
 2. `return` because an `Err` means it cannot be handled
 
-`?` is *almost*[^1] exactly equivalent to an `unwrap` which `return`s
+`?` is *almost*[^†] exactly equivalent to an `unwrap` which `return`s
 instead of `panic`s on `Err`s. Let's see how we can simplify the earlier
 example that used combinators:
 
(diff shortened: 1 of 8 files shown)


src/error/result/enter_question_mark.md
```diff
@@ -9,7 +9,7 @@ Upon finding an `Err`, there are two valid actions to take:
 1. `panic!` which we already decided to try to avoid if possible
 2. `return` because an `Err` means it cannot be handled
 
-`?` is *almost*[^1] exactly equivalent to an `unwrap` which `return`s
+`?` is *almost*[^†] exactly equivalent to an `unwrap` which `return`s
```
projektir: So... this one is funky. There are two pages in RBE that are attempting to use the `[^1]` annotation. The way [pulldown-cmark](https://github.com/google/pulldown-cmark) does annotations is by creating a `div` with `id` equal to the annotation name (in this case, "1"). This is all well and good, until the `print.html` page is created, in which case this now results in two different `div`s with `id="1"`, which is invalid and is a failure to the linkchecker.

I've replaced `1` here with the dagger symbol `†`, which looks like `1` in the fonts used by RBE, but I am not really sure how to address this issue in general since we basically can't use the 1, 2, 3 footnote style for as long as we intend to have a page that concatenates everything together.

marioidival:
LGTM, @QuietMisdreavus u can review and merge this?

Michael-F-Bryan:
> Looks good. That's a real shame about the footnotes, though. We'll need to figure that one out eventually. :/

It sounds like this isn't a RBE-specific issue, so we may be able to help out upstream. I've created an issue for it (rust-lang-nursery/mdBook#542).

PR #993, opened 2018-01-14
refine wording

bmusin:
"lambda expression" is more precise and technically correct term than simply "lambda"

### DIFF
--- a/src/fn/closures.md
+++ b/src/fn/closures.md
@@ -1,6 +1,6 @@
 # Closures
 
-Closures in Rust, also called lambdas, are functions that can capture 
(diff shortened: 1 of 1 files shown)


marioidival:
@QuietMisdreavus LGTM

QuietMisdreavus:
Hmm. While it's more accurate to call them "lambda expressions", it would still be worth noting that they're often just called "lambdas" informally. Maybe like this?

> Closures in Rust, also called lambda expressions or just lambdas, \[...\]

projektir:
@bmusin 
Hey, are you planning to update this

bmusin:
Thanks, for reminding. Updated.

PR #994, opened 2018-01-14
add example of using `move` keyword

bmusin:
Show how to use `move` keyword to force closure to take ownership
of captured variables.

### DIFF
--- a/src/fn/closures/capture.md
+++ b/src/fn/closures/capture.md
@@ -69,6 +69,30 @@ fn main() {
 }
 ```
 
+Using `move` before vertical pipes forces closure
+to take ownership of captured variables:
+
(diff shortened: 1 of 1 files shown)


QuietMisdreavus:
This doesn't seem like the right page for this? I feel like this would work best in `fn/closures/capture.md`, rather than in the page that talks about `while let`.

bmusin:
@QuietMisdreavus, should I make new PR with this addition to  fn/closures/capture.md?

QuietMisdreavus:
If you want to do that, that would be fine. It may be easier if you've been doing it on GitHub online.

bmusin:
@QuietMisdreavus, moved into capture.md

PR #998, opened 2018-01-14
fix typo

bmusin:


steveklabnik:
This is not a typo; "Cargo" is its proper name, and `cargo` is the command-line tool. The intention was to use the proper name here.

Thank you though!

PR #1004, opened 2018-01-16
replace . with : for consistency

bmusin:
In other places : is used instead of dot.


marioidival:
SGTM, @QuietMisdreavus @steveklabnik

PR #1007, opened 2018-01-16
change println -> eprintln where appropriate

bmusin:
Should I amend this PR by including changes, so
that program prints "Usage..." to stderr as well?

### DIFF
--- a/src/std_misc/arg/matching.md
+++ b/src/std_misc/arg/matching.md
@@ -46,7 +46,7 @@ fn main() {
                     n
                 },
                 Err(_) => {
-                    println!("error: second argument not an integer");
+                    eprintln!("error: second argument not an integer");
                     help();
                     return;
                 },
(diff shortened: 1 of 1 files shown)


src/std_misc/arg/matching.md
```diff
@@ -32,8 +32,8 @@ fn main() {
         // one argument passed
         2 => {
             match args[1].parse() {
-                Ok(42) => println!("This is the answer!"),
-                _ => println!("This is not the answer."),
+                Ok(42) => eprintln!("This is the answer!"),
```
marioidival: I think that use of `eprintln` is just to error. Here can be `println`.

bmusin: That's reasonable. Changed.

marioidival:
@steveklabnik It's good enough to merge?

PR #1008, opened 2018-01-16
collapse two examples into one to simplify reading

bmusin:


marioidival:
LGTM, Thank you @bmusin

PR #1010, opened 2018-01-16
Add example to partial struct's copies

### LINKED ISSUE #207: Struct: Add example about "partial" copies/clones
Something like this, but nicer:

``` rust
#[deriving(Show)]
struct Triplet {
    first: uint,
    second: uint,
    third: uint,
}

fn main() {
    let t = Triplet { first: 0, second: 1, third: 2 };
    let u = Triplet { first: 3, .. t };

    println!("{}", t);
    println!("{}", u);
}
```

Output:

```
Triplet { first: 0, second: 1, third: 2 }
Triplet { first: 3, second: 1, third: 2 }
```

Probably should be in the `struct` chapter.

marioidival:
fixes #207

### DIFF
--- a/src/custom_types/structs.md
+++ b/src/custom_types/structs.md
@@ -38,16 +38,21 @@ fn main() {
     let name = "Peter";
     let age = 27;
     let peter = Person { name, age };
-    
+
     // Print debug struct
     println!("{:?}", peter);
-    
-    
+
+
     // Instantiate a `Point`
     let point: Point = Point { x: 0.3, y: 0.4 };
 
     // Access the fields of the point
     println!("point coordinates: ({}, {})", point.x, point.y);
+
+    // Make a new point by using struct update syntax to use the fields of our other one
+    let new_point = Point { x: 0.1, ..point };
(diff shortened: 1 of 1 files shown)


marioidival:
@QuietMisdreavus you have permissions to rebuild this in travis?

marioidival:
@projektir @QuietMisdreavus 
--

It’s often useful to create a new instance of a struct that uses most of an old instance’s values, but changes some. We do this using struct update syntax.


how we create a new `Triplet` instance in `t` without the update syntax. We set new values for `first` and `second`, but otherwise use the same values from `u` that we created.

QuietMisdreavus:
This is probably better served by adding a few lines to the existing example in this chapter, rather than spotlighting it on its own. Right after the `println!("point coordinates:")` add a few lines like this:

```rust
// Make a new point by using struct update syntax to use the fields of our other one
let new_point = Point { x: 0.1, ..point };

// `new_point.y` will be the same as `point.y` because we used that field from `point`
println!("second point: ({}, {})", new_point.x, new_point.y);
```

marioidival:
@QuietMisdreavus @projektir updated!

PR #1012, opened 2018-01-17
fixed small typos

brieucdesoutter:


src/cargo/deps.md
```diff
@@ -1,8 +1,8 @@
 # Dependencies
 
 Most programs have dependencies on some libraries. If you have ever managed
-dependencies by hand, you know how much of a pain this. Luckily, the Rust
-ecosystem comes standard with `cargo`! `cargo` can manage dependcies for a
+dependencies by hand, you know how much of a pain this can be. Luckily, the Rust
```
frewsxcv: Any reason why 'can be' was removed?

frewsxcv: Oh I read this backwards. I thought it was removed but you added it in. All good!

PR #1015, opened 2018-01-24
Document inclusive ranges.

alercah:
This provides documentation for rust-lang/rust#28237.


steveklabnik:
These just landed in stable. Let's retry travis!

alercah:
Looks like it's failing; is it running stable and/or not up to date with the nightly?

marioidival:
@alercah Rust stable https://github.com/rust-lang/rust-by-example/blob/master/.travis.yml#L5

frewsxcv:
seems like we'll need to do one of three:

* wait until it lands in stable rust, then merge
* ...or update travis to use nightly rust
* ...or update the code block to be `rust,ignore`

frewsxcv:
actually, since rust-by-example now rides the docs release train, _seems_ like it should be safe to do the second option and turn on nightly rust for travis ci

steveklabnik:
Yes, that's what we should do. Run with nightly but check there aren't any feature gates.

PR #1017, opened 2018-01-25
Simplify new type idiom example

michel-zimmer:


marioidival:
@michel-zimmer thank you!

PR #1018, opened 2018-01-26
Fix failing tidy on Windows

petrochenkov:
This happens every time a new doc submodule is added.
cc https://github.com/rust-lang/book/pull/549 https://github.com/rust-lang-nursery/reference/pull/36


marioidival:
@petrochenkov thank you!

PR #1021, opened 2018-01-30
Add more attribute examples

### LINKED ISSUE #1020: More attribute examples
The attribute examples don't cover that

* it's possible to have multiple attributes between the brackets: `#[attribute(value), attribute2(value)]`
* it's possible to split up the attributes within the brackets over multiple lines

I would be happy to add this to the docs if you feel that this is useful, too. Possibly at the bottom of [Attribute](https://rustbyexample.com/attribute.html)?

phansch:
Closes #1020


projektir:
Thanks!

PR #1022, opened 2018-02-01
Unify Rust Code Block Syntax

JordanAdams:
Currently there is a mix of syntaxes for Rust code blocks. This leads to some minor stylistic inconsistencies and play buttons on some non-runnable code.

All Rust code blocks now conform to one of two labels invocation syntaxes

**Editable:**
```
```rust,editable
```

**Non-Editable**
```
```rust,ignore
```

In cases where neither `editable` or `ignore` were specified, I've based the addition on if the code is runnable and produces an output.

🦄

### DIFF
--- a/src/conversion/from_into.md
+++ b/src/conversion/from_into.md
@@ -13,7 +13,7 @@ library for conversion of primitive and common types.
 
 For example we can easily convert a `str` into a `String`
 
-```rust
+```rust,ignore
 let my_str = "hello";
 let my_string = String::from(my_str);
 ```
--- a/src/attribute/cfg/custom.md
+++ b/src/attribute/cfg/custom.md
@@ -3,7 +3,7 @@
 Some conditionals like `target_os` are implicitly provided by `rustc`, but
 custom conditionals must be passed to `rustc` using the `--cfg` flag.
 
-```rust,editable,ignore,mdbook-runnable
+```rust,editable
 #[cfg(some_condition)]
 fn conditional_function() {
     println!("condition met!");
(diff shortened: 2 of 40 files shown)


steveklabnik:
I haven't actually reviewed this PR, but a huge 👍 on the concept

projektir:
I'm not sure if I understand this PR since all those attributes do specific things as far as I recall. `editable` makes something editable, or not. `ignore` makes something not-runnable, iirc. Removing `ignore` may result in users trying to run examples that were never meant to succeed. Modifying these attributes changes how the examples work, which doesn't really sound right to me.

steveklabnik:
@JordanAdams any response to @projektir ? I think when I made my comment, I mis-undestood what the PR was about.

JordanAdams:
So the idea here was to make it more clear to the end-user what code blocks were interactive. There were a number of cases where code which wasn't runnable had a run button or was editable. I don't think it really makes sense for non-runnable code to be editable. This puts all code blocks in one of two forms; editable & runnable **or** not editable and not runnable.

src/conversion/from_into.md
```diff
@@ -13,7 +13,7 @@ library for conversion of primitive and common types.
 
 For example we can easily convert a `str` into a `String`
 
-```rust
+```rust,ignore
```
frewsxcv: we want this to be compiled during our tests to ensure it remains a valid rust program, but by changing it to `ignore`, we no longer keep that guarantee. can you expand on your decision to add `ignore` here?

steveklabnik:
> I was under the impression that it simply controlled wether or not the code editor should be rendered. If that's not the case I'm more than happy to close this :)

Okay, it does seem there was a misunderstanding here. As such, I think I'm gonna give this a close. Sorry for all the confusion, @JordanAdams <3

Issue #1025, opened 2018-02-15
CI is slow


epage:
As one example of a slow down:
```
$ (cargo install mdbook --git https://github.com/azerupi/mdBook.git --force || true)
    Updating git repository `https://github.com/azerupi/mdBook.git`
  Installing mdbook v0.0                              before_script
494.97s
```
From [PR 971](https://travis-ci.org/rust-lang/rust-by-example/builds/310715762#L484)

epage:
The problem is that mdbook is being rebuilt on every PR.
See https://github.com/rust-lang/rust-by-example/blob/master/.travis.yml

One way of speeding it up is to download the prebuilt binaries.

For example: https://github.com/crate-ci/crate-ci.github.io/blob/source/.travis.yml

PR #1033, opened 2018-03-15
Don’t install mdbook if latest version is already installed.

frewsxcv:
Fixes https://github.com/rust-lang/rust-by-example/issues/1025.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,12 @@ branches:
   only:
   - master
 before_script:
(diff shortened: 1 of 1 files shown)


frewsxcv:
at least for this pr, it looks like mdbook isn't ever being retrieved from the cache, so each commit in this pr is triggering an mdbook install

steveklabnik:
This looks fine to me. Soon enough we'll be removing this entirely, so it's not a big deal. (It's now being published on doc.rust-lang.org so no need to use mdbook in CI at all.

PR #1036, opened 2018-03-27
Rename string conversion page name.

frewsxcv:
In my opinion, the previous title makes it sounds like we use the `From` trait (which is mentioned in the previous section), which is not the case


marioidival:
@frewsxcv thank you!

PR #1080, opened 2018-05-14
Update mdBook to 0.1.7

mattico:
cc https://github.com/rust-lang-nursery/mdBook/issues/693 https://github.com/ruRust/rust-by-example-ru/issues/120

I'd like to make these buttons user-configurable so you don't need to have a fork of the index.hbs. See https://github.com/rust-lang-nursery/mdBook/issues/694

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ If you'd like to read it locally, [install Rust], and then:
 ```bash
 $ git clone https://github.com/rust-lang/rust-by-example
 $ cd rust-by-example
-$ cargo install mdbook
+$ cargo install --version 0.1.7 mdbook
 $ mdbook build
 $ mdbook serve
 ```
--- a/theme/index.hbs
+++ b/theme/index.hbs
@@ -1,11 +1,13 @@
 <!DOCTYPE HTML>
-<html lang="{{ language }}">
+<html lang="{{ language }}" class="sidebar-visible no-js">
     <head>
+        <!-- Book generated using mdBook -->
         <meta charset="UTF-8">
         <title>{{ title }}</title>
         <meta content="text/html; charset=utf-8" http-equiv="Content-Type">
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -14,7 +14,7 @@ If you'd like to read it locally, [install Rust], and then:
 ```bash
 $ git clone https://github.com/rust-lang/rust-by-example
 $ cd rust-by-example
-$ cargo install mdbook
+$ cargo install --version 0.1.7 mdbook
```
steveklabnik: since mdbook is already at 0.1.7, this doesn't change things; it actually pins it to that specific version. we already use 0.1.7 https://github.com/rust-lang/rust/blob/master/src/tools/rustbook/Cargo.toml#L11 to render the book.

mattico:
The template used for `index.hbs` is very tightly coupled to the version of `mdbook`. If the version doesn't match you can get [strange issues](https://github.com/rust-lang-nursery/mdBook/issues/693). Because Rust by Example uses a custom `index.hbs`, it may not work properly with past or future versions of `mdbook`, and the readme should recommend users to install the version that works.

So then why doesn't the Rust By Example hosted by rust-lang experience any strange issues? It turns out that the theme override is broken (https://github.com/rust-lang-nursery/mdBook/issues/695) so the custom `index.hbs` isn't being used at all. If you add
```toml
[output.html]
theme = "theme"
```
to `book.toml`, the book does render incorrectly.

Given this new information, this PR doesn't actually do anything. I'll send another similar one once 0.1.8 is out with a fix for the above issue.

steveklabnik:
Whew! Thanks for all that context, and for pushing upstream. Not having the fork will be great!

mattico:
See also the edition guide, which seems to exhibit the above mentioned problems: https://rust-lang-nursery.github.io/edition-guide/

PR #1083, opened 2018-05-17
Redirect to the desired path from the old site

jbradberry:
Whatever path the user tries to access on rustbyexample.com, pass it through when redirecting to the rust-by-example/ pages on doc.rust-lang.org.


jbradberry:
:grinning:

PR #1084, opened 2018-05-28
Synchronize Code of Conduct with rust-www

davidalber:
The Code of Conduct in this repository has fallen out of sync with the [source Code of Conduct in rust-lang/rust-www](https://github.com/rust-lang/rust-www/blob/master/en-US/conduct.md) (and the [clone in rust-lang/rust](https://github.com/rust-lang/rust/blob/master/CODE_OF_CONDUCT.md), for example). This PR resynchronizes the Code of Conduct with the other Rust repositories that contain it.


frewsxcv:
Thanks!

PR #1089, opened 2018-07-12
Clarify the result of running the sample code for the wait section

### LINKED ISSUE #1088: Output provided for the wait section is slightly misleading
For the following section:
https://doc.rust-lang.org/stable/rust-by-example/std_misc/process/wait.html

This code snippet:
```rust
use std::process::Command;

fn main() {
    let mut child = Command::new("sleep").arg("5").spawn().unwrap();
    let _result = child.wait().unwrap();

    println!("reached end of main");
}
```

is followed by this sample output:
```
$ rustc wait.rs && ./wait
reached end of main
# `wait` keeps running for 5 seconds
# `sleep 5` command ends, and then our `wait` program finishes
```

The comment `wait keeps running for 5 seconds` is placed after the `reached end of main` line, which seems to indicate that the program first prints `reached end of main` *before* waiting, which is incorrect.

maljub01:
This changes the order of the sleep comment and print output to make it
match the correct order that those events happen in.

Fixes #1088.

### DIFF
--- a/src/std_misc/process/wait.md
+++ b/src/std_misc/process/wait.md
@@ -16,7 +16,6 @@ fn main() {
 
 ```bash
 $ rustc wait.rs && ./wait
+# `wait` keeps running for 5 seconds until the `sleep 5` command finishes
 reached end of main
-# `wait` keeps running for 5 seconds
-# `sleep 5` command ends, and then our `wait` program finishes
 ```


maljub01:
@marioidival thanks for approving. Unfortunately I don't have write access to merge this though.

marioidival:
@maljub01 thank you

PR #1090, opened 2018-07-13
Fix a typo in the "Filesystem Operations" section

maljub01:
There is no `std::io::fs` module.


marioidival:
@maljub01 thank you

PR #1091, opened 2018-07-13
Fix a broken link to the Rust Cookbook

maljub01:


marioidival:
@maljub01 thank you again!

PR #1096, opened 2018-08-14
Stabilize macro_vis_matcher

jkozlowski:
r? @cramertj


jkozlowski:
I think so, although this is my first time contributing, so not sure what the protocol is :)

PR #1097, opened 2018-08-15
Add how to use brace bracket syntax

lf94:
Self-explanatory.

### DIFF
--- a/src/mod/use.md
+++ b/src/mod/use.md
@@ -4,7 +4,7 @@ The `use` declaration can be used to bind a full path to a new name, for easier
 access. It is often used like this:
 
 ```rust,editable
-extern crate deeply;
+// extern crate deeply; // normally, this would exist and not be commented out!
 
 use deeply::nested::{
     my_first_function,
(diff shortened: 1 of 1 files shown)


src/mod/use.md
```diff
@@ -1,7 +1,23 @@
...
+    my_second_function,
+    AndATraitType
+};
+
+fn main() {
+    my_first_function();
+}
+```
+
+Alternatively you can use `as` syntax too:
```
frewsxcv: It's possible to use `as` within an import group:

```rust
use deeply::nested::{
    my_first_function,
    my_second_function as something_else,
};
```

So since this isn't an either-or scenario, "Alternatively" doesn't seem like the right word here. How about:

> You can use the `as` keyword to bind imports to a different name:

steveklabnik:
This can't ever build, as the crate won't exist, and so should be marked as `ignore`. LGTM otherwise!

lf94:
Sorry @steveklabnik I'm not sure what I'm supposed to do. I can't find any documentation on "ignore crate", only a literal crate called "ignore" keeps coming up... :cry:

steveklabnik:
@lf94 no worries! I mean https://doc.rust-lang.org/rustdoc/documentation-tests.html#attributes

does that make sense?

lf94:
I hope that did it!

steveklabnik:
It looks good to me! Let's see what Travis says..

marioidival:
@steveklabnik travis says SGTM, so I merge this! Thank you @lf94

PR #1103, opened 2018-09-03
Update list of unsafe things

vks:
* Inline assembly is not stable and is thus no longer mentioned.
* Mention static mutable variables and unsafe traits instead.

### DIFF
--- a/src/unsafe.md
+++ b/src/unsafe.md
@@ -2,9 +2,9 @@
 
 As an introduction to this section, to borrow from [the official docs][unsafe],
 "one should try to minimize the amount of unsafe code in a code base." With that
-in mind, let's get started! Unsafe blocks in Rust are used to bypass protections
-put in place by the compiler; specifically, there are four primary things that
-unsafe blocks are used for:
+in mind, let's get started! Unsafe annotations in Rust are used to bypass
+protections put in place by the compiler; specifically, there are four primary
+things that unsafe is used for:
 
 * dereferencing raw pointers
 * calling functions or methods which are `unsafe` (including calling a function
@@ -45,7 +45,7 @@ fn main() {
 
     unsafe {
         let my_slice: &[u32] = slice::from_raw_parts(pointer, length);
-        
+
         assert_eq!(some_vector.as_slice(), my_slice);
     }
 }
--- a/src/unsafe.md
+++ b/src/unsafe.md
@@ -7,10 +7,10 @@ put in place by the compiler; specifically, there are four primary things that
 unsafe blocks are used for:
 
 * dereferencing raw pointers
-* calling a function over FFI (but this is covered in [a previous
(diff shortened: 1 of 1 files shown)


src/unsafe.md
```diff
@@ -7,10 +7,10 @@ put in place by the compiler; specifically, there are four primary things that
...
 
 * dereferencing raw pointers
-* calling a function over FFI (but this is covered in [a previous
-  chapter](std_misc/ffi.html) of the book)
-* calling functions which are `unsafe`
-* inline assembly
+* calling functions or methods which are `unsafe` (including calling a function
+  over FFI, see [a previous chapter](std_misc/ffi.html) of the book) 
+* accessing or modifying static mutable variables
+* implementing unsafe traits
```
frewsxcv: This doesn't require an unsafe _block_, but I think it's still worth calling out in this list 🤔

src/unsafe.md
```diff
@@ -7,10 +7,10 @@ put in place by the compiler; specifically, there are four primary things that
 unsafe blocks are used for:
 
 * dereferencing raw pointers
-* calling a function over FFI (but this is covered in [a previous
-  chapter](std_misc/ffi.html) of the book)
-* calling functions which are `unsafe`
-* inline assembly
+* calling functions or methods which are `unsafe` (including calling a function
+  over FFI, see [a previous chapter](std_misc/ffi.html) of the book) 
+* accessing or modifying static mutable variables
```
frewsxcv: Does accessing require an unsafe block?

vks: TRPL claims it does, but I didn't check. It makes sense to me, because you can still get data races when one thread is modifying and the other is accessing.

src/unsafe.md
vks: Fair enough, maybe the text should change to talk about unsafe annotations. Technically, you don't have to use unsafe blocks at all, you can just declare unsafe functions.

src/unsafe.md
steveklabnik: it is https://play.rust-lang.org/?gist=8ff0c233146c8cc275bd36c0659bb91c&version=stable&mode=debug&edition=2015

src/unsafe.md
```diff
@@ -7,10 +7,10 @@ put in place by the compiler; specifically, there are four primary things that
 unsafe blocks are used for:
```
steveklabnik: I think we can fix [this](https://github.com/rust-lang/rust-by-example/pull/1103/files#r215113805) by modifying this paragraph to say only "unsafe" rather than "unsafe block"

src/unsafe.md
vks: Done.

src/unsafe.md
vks: Done.

Issue #1062, opened 2018-04-04
update for main with non-`()` return types


steveklabnik:
This issue tracks inclusion of information about main with non-() return types into the Book. This will happen in the 2018 edition.

Tracking issue: rust-lang/rust#48453

maccoda:
I wouldn't mind picking this one up. Did we have any ideas for where this would be located in the book?

frewsxcv:
Maybe as a sub-section of `Result`?

https://doc.rust-lang.org/stable/rust-by-example/error/result.html

```
18.3.5 Using `Result` in `main`
```

PR #1108, opened 2018-10-25
Return Result from main function.

maccoda:
Hopefully what we are looking for to cover #1062


src/error/result.md
```diff
@@ -45,6 +45,39 @@ unpleasant error message.
 To improve the quality of our error message, we should be more specific
 about the return type and consider explicitly handling the error.
 
+## Using `Result` in `main`
+
+The `Result` type can also be the return type of of the `main` function if
```
marioidival: `type of of`

marioidival: `type of of`

PR #1113, opened 2018-11-08
Minor fixes for the macros chapter

nkaretnikov:


src/macros/dsl.md
```diff
@@ -39,3 +39,6 @@ Output:
 This was a very simple example, but much more complex interfaces have been
 developed, such as [`lazy_static`](https://crates.io/crates/lazy_static) or
 [`clap`](https://crates.io/crates/clap).
+
+Also, note the two pairs of braces in the macro. The outer ones are
+part of the syntax of `macro_rules!`, in addition to `()` or `[]`.
```
frewsxcv: This is a little confusing to me – which "macro" are you referring to here? what is the "outer one"?

nkaretnikov: The macro that's defined, there's only one. It's almost a copy-paste from the first edition of the [book](https://doc.rust-lang.org/book/first-edition/macros.html#expansion):
> The outer braces are part of the syntax of macro_rules!. In fact, you can use () or [] instead.
> They simply delimit the right-hand side as a whole.

I didn't link to the book since it's the first edition and RBE only references the second one, which doesn't cover this bit.

Would you prefer something like this?
>  Also, note the two pairs of braces in the macro definition [or: in the `calculate` definition]. The outer braces are part of the syntax ...

If not, feel free to suggest a better version. I think it's important to explain this syntax since it's the first time it is used in this book.

nkaretnikov: @frewsxcv @steveklabnik why is this not going forward?

PR #1138, opened 2018-12-30
Added two missing full stops

bartsmykla:
In listing of derivable traits, two of the list items
didn't have full stop at the end, like the other items.

Signed-off-by: Bart Smykla <bsmykla@vmware.com>

PR #1139, opened 2018-12-31
Removed unnecessary spaces before macro designators in macros/dry

bartsmykla:
In other places there is no space before designators so I removed
unnecessary spaces in two places to make it more consistent.

Signed-off-by: Bart Smykla <bsmykla@vmware.com>


bartsmykla:
@QuietMisdreavus can you look at this one too? :-)

bartsmykla:
@QuietMisdreavus I believe it's merged now. Can you fire CI again here and in https://github.com/rust-lang/rust-by-example/pull/1138? :-)

PR #1145, opened 2019-01-17
Recommend implementing Display over ToString

### LINKED ISSUE #1131: "To and from Strings"-section recommends ToString in conflict with the book
The "To and from Strings"-section recommends implementing the `ToString`-trait directly [1](https://doc.rust-lang.org/rust-by-example/conversion/string.html).  According to the [docs](https://doc.rust-lang.org/std/string/trait.ToString.html), best practice is to not implement this trait directly, but rather to implement the `fmt::Display` trait.

Related, the section on the `fmt::Display` trait does not mention that implementing `fmt::Display` gives `ToString` for free. 

I suggest to replace the "To String" section in "To and from Strings" with an example of this approach, and to include a link to this section from the existing section on `fmt::Display`.

Should I submit a pull-request to this effect, or have I missed something?

Japanuspus:
Resolves #1131.

### DIFF
--- a/src/conversion/string.md
+++ b/src/conversion/string.md
@@ -1,20 +1,22 @@
 # To and from Strings
 
-## `ToString`
+## Converting to String
 
-To convert any type to a `String` it is as simple as implementing the [`ToString`]
-trait for the type.
+To convert any type to a `String` is as simple as implementing the [`ToString`]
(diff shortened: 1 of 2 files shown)


marioidival:
@Japanuspus thank you

PR #1146, opened 2019-01-17
Hello: Fix hint link in `fmt` chapter.

bperson:
It looks like one of the hints is pointing to an irrelevant paragraph about argument types even though the text suggests it should point to the named parameters to DRY the code a bit.


marioidival:
@bperson thank you

PR #1148, opened 2019-01-20
Update new_types.md to clarify conversion to base type

cwaldren:
As someone new to the language, I struggled for a little while trying to figure out how to make use of the `newtype` idiom. I needed to convert my custom type back to the base type, and the documentation did not make it obvious. 

I tried using `as`, I tried using `From`, and neither worked. 

I wasn't aware of the differences/similarities between structs, struct tuples, and tuples, and I think this deserves more documentation as well.


marioidival:
@cwaldren thank you

PR #1151, opened 2019-02-04
Do not use deprecated functions from `std::error::Error` trait

yerke:
According to https://doc.rust-lang.org/std/error/trait.Error.html, `fn description` is deprecated in favor of `Display` trait and `fn cause` will be deprecated starting from 1.33 in favor of `fn source` (which is already available on stable from 1.30).

I updated the examples accordingly.


marioidival:
@yerke thank you

PR #1152, opened 2019-02-05
Fix borrow so it fails in 2018 edition Fixes #1141

### LINKED ISSUE #1141: 2018 borrow checker
https://doc.rust-lang.org/rust-by-example/scope/borrow.html shows an example that supposedly does not work, but actually does work with Rust 2018:
```rust
{
        // Take a reference to the data contained inside the box
        let _ref_to_i32: &i32 = &boxed_i32;

        // Error!
        // Can't destroy `boxed_i32` while the inner value is borrowed.
        eat_box_i32(boxed_i32);
        // FIXME ^ Comment out this line

        // `_ref_to_i32` goes out of scope and is no longer borrowed.
    }
```

gurgalex:
Fix #1152

### DIFF
--- a/src/scope/borrow.md
+++ b/src/scope/borrow.md
@@ -29,18 +29,14 @@ fn main() {
     borrow_i32(&boxed_i32);
     borrow_i32(&stacked_i32);
 
-    {
-        // Error!
-        // Can't destroy `boxed_i32` while the inner value is borrowed later in scope.
-        eat_box_i32(boxed_i32);
-        // FIXME ^ Comment out this line
+    // Take a reference to the data contained inside the box
+    let _ref_to_i32: &i32 = &boxed_i32;
 
-        // Take a reference to the data contained inside the box
-        let _ref_to_i32: &i32 = &boxed_i32;
-        // `_ref_to_i32` goes out of scope and is no longer borrowed.
-    }
-
-    // `boxed_i32` can now give up ownership to `eat_box` and be destroyed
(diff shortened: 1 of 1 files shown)


frewsxcv:
this causes a different error than the one we want

with rust 2015, the error is:

```
error[E0505]: cannot move out of `boxed_i32` because it is borrowed
  --> src/main.rs:27:21
   |
23 |         let _ref_to_i32: &i32 = &boxed_i32;
   |                                  --------- borrow of `boxed_i32` occurs here
...
27 |         eat_box_i32(boxed_i32);
   |                     ^^^^^^^^^ move out of `boxed_i32` occurs here
```

with these changes, the error is now:

```
error[E0382]: use of moved value: `boxed_i32`
  --> src/main.rs:30:34
   |
27 |         eat_box_i32(boxed_i32);
   |                     --------- value moved here
...
30 |         let _ref_to_i32: &i32 = &boxed_i32;
   |                                  ^^^^^^^^^ value used here after move
   |
   = note: move occurs because `boxed_i32` has type `std::boxed::Box<i32>`, which does not implement the `Copy` trait
```

marioidival:
@gurgalex ^

gurgalex:
Ah, whoops.

I'll think about this again today.

gurgalex:
@frewsxcv 
The error messages are now the same as before.

marioidival:
@gurgalex thank you!

PR #1153, opened 2019-02-14
Add "literal" to list of macro designators

anowlcalledjosh:
[`macro_literal_matcher`](https://github.com/rust-lang/rfcs/blob/master/text/1576-macros-literal-matcher.md) is stable, and is in Rust 1.32, but isn't currently documented in Rust By Example.


marioidival:
@anowlcalledjosh thank you!

PR #1159, opened 2019-02-25
Add link to Cargo chapter in the index page

### LINKED ISSUE #1158: Cargo is missing in the Introduction/Homepage
In the list of chapters, Cargo is between Crates (chapter 11) and Attributes (chapter 13). However, the homepage is missing a link to the Cargo chapter.

I'd like to add a link to the Cargo chapter to the list along with a basic description:

> * Cargo - Go through some basic features of the official Rust package management tool.

I'm happy to take suggestions on a better description.

Also, do we need to update #978?

I've noticed several problems with the homepage, but I'd like to keep this issue focused on just the missing Cargo chapter item. I'll open a separate issue to list the other problems. However, one problem related to this missing chapter link is the use of an unordered list. It's easier to notice a missing chapter if we have an ordered list instead. Should we make the switch? I can include a change to an ordered list when I submit a PR for this depending on feedback.

mikong:
Fixes #1158

The change simply updates the Introduction/Homepage/index page to add a link to the Cargo chapter (between Crates and Attributes in the list) along with a basic description.

Please advise if you have a better description that I should use for Cargo so I can update the PR.

### DIFF
--- a/src/index.md
+++ b/src/index.md
@@ -33,6 +33,8 @@ Now let's begin!
 
 - [Crates](crates.html) - A crate is a compilation unit in Rust. Learn to create a library.
 
+- [Cargo](cargo.html) - Go through some basic features of the official Rust package management tool.
+
 - [Attributes](attribute.html) - An attribute is metadata applied to some module, crate or item.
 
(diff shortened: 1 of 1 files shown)


marioidival:
@mikong thank you!

PR #1160, opened 2019-02-25
Replace lvalue and rvalue with place and value

### LINKED ISSUE #988: Rvalue and Lvalue have different names now
Lvalues are called place expressions and Rvalues are called value expressions. This has been updated at least partially in the compiler and fully in the Reference. It needs to be updated here as well.

mikong:
Fixes #988

I didn't perform a simple find and replace. For rvalue, in one instance I simply replaced it with _"value"_, but in another, I used the phrase _"value expression"_. For lvalue, I replaced it with _"place expression such as a local variable"_ to make it easier to understand in the given context.

I'm open to suggestions to improve the changes.

### DIFF
--- a/src/expression.md
+++ b/src/expression.md
@@ -25,9 +25,9 @@ fn main() {
 }
 ```
 
-Blocks are expressions too, so they can be used as [r-values][rvalue] in
+Blocks are expressions too, so they can be used as values in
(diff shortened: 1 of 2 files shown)


marioidival:
@mikong thank you!

Issue #1149, opened 2019-01-26
From/Into section 6.1 introduction confusing


mitchblevins:
The introductory text of section 6.1 suggests that the reciprocal Into trait for a From trait does the conversion in the opposite direction.  (type A to B, type B to A).

This is confusing because the conversion is always type A to B, but the From/Into difference is which type it is initiated by.

jonmccune:
I'm a random Rust n00b going through rust-by-example, and I also noticed this. After staring at it for a while, I actually think it would be an improvement if the first two sentences were simply deleted. I.e., just drop:

"The From and Into traits are inherently linked, and this is actually part of its implementation. If you are able to convert type A from type B, then it should be easy to believe that we should be able to convert type B to type A."

jakehemmerle:
Hi, I thought this was confusing too and I came to open an issue.

The example in the book only shows how to convert A to B (using both `B::from(a)` and a.into()`), but it very well suggests we can convert A to B as well as B to A.

These two points should be emphasized:
1. The `From` implementation is preferred to Into
2. `Into` is **automatically** implemented on the destination type,

There was a third claim on the above, closed, PR:
> 3.`Into` must be implemented instead of `From` when both types are not defined by the current crate.

3 appears to be not true, according to this error I got today:

```rust
error[E0117]: only traits defined in the current crate can be implemented for types defined outside of the crate
  --> crypto/shared/src/constraints.rs:27:1
   |
27 | impl Into<sp_core::H160> for H160 {
   | ^^^^^-------------------^^^^^----
   | |    |                       |
   | |    |                       `primitive_types::H160` is not defined in the current crate
   | |    `sp_core::H160` is not defined in the current crate
   | impl doesn't use only types from inside the current crate
   |
   = note: define and implement a trait or new type instead

```

If someone elaborate on the idiomatic way to allow for type conversion between types in different crates, I will open a new PR with these changes. Using a new arbitrary trait seems dirty.

jakehemmerle:
cc @marioidival for visibility

PR #1163, opened 2019-03-06
Clarify `Into` description

crajcan:
https://github.com/rust-lang/rust-by-example/issues/1149

### DIFF
--- a/src/conversion/from_into.md
+++ b/src/conversion/from_into.md
@@ -42,13 +42,8 @@ fn main() {
 
 ## `Into`
 
-The [`Into`] trait is simply the reciprocal of the `From` trait. That is, if you
-have implemented the `From` trait for your type you get the `Into`
-implementation for free.
-
-Using the `Into` trait will typically require specification of the type to
-convert into as the compiler is unable to determine this most of the time.
-However this is a small trade-off considering we get the functionality for free.
+The [`Into`] trait is derived from the `From` trait and is a syntactically
+different way to perform the same conversion.
 
 ```rust,editable
 use std::convert::From;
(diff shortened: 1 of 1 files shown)


marioidival:
@crajcan Oops, I'm sorry for the long delay.

I thought your change was good but to complement what we already had.

You can take your paragraph and replace it only with the first paragraph of the code snippet, keeping the second paragraph?

crajcan:
> @crajcan Oops, I'm sorry for the long delay.
> 
> I thought your change was good but to complement what we already had.
> 
> You can take your paragraph and replace it only with the first paragraph of the code snippet, keeping the second paragraph?

@marioidival, thanks for taking the time to review.

Now that I look at this closer, it is apparent you are right about keeping the second paragraph. I have made another commit for you to review.

There is another sentence at the top of this page that I think has mislead some users (see issue  #1149):

```If you are able to convert type A from type B, then it
should be easy to believe that we should be able to convert type B to type A.```

And perhaps a sentence in the API documentation that has done the same:

https://github.com/rust-lang/rust/blob/d19a359444295bab01de7ff44a9d72302e573bc9/src/libcore/convert.rs#L203

Since both from and into have the same type signatures, I think what is important to highlight in Rust By Example is that:

1. The `From` implementation should be preferred
2. `Into` is implemented on the destination type, while `From` is implemented on the origin type.
3. `Into` must be implemented instead of `From` when both types are not defined by the current crate.

Perhaps these differences are be too subtle to warrant description in RBE. I'll try to make another small PR or two, and maybe you can review?

marioidival:
@crajcan of course!

marioidival:
@crajcan you can make it on this PR

marioidival:
@crajcan ?

marioidival:
@crajcan i can close this PR?

PR #1172, opened 2019-04-13
Fix a typo ("half" -> "halve")

schutzekatze:
Half mistakenly used as a verb; should be halve.


marioidival:
@schutzekatze thank you!

PR #1173, opened 2019-04-18
Small improvements to various files

alexander-clarke:
Various small fixes to:
11.1: This code will not run as there is no main function, so removed the ability to run it
24.1: This code will not run due to the crate_name being set to something other than playground. Could be changed in a different way, but running the code is not the purpose of this code example so I removed the ability to run it. This could be changed in a different way, open to discussion if this isn't the best.

3.2: Removed an unneeded attribute of allow(dead_code) as there isn't any dead code

8.5.1.4: Changed example to actually use match as all the other destructuring examples do. This is the biggest change so can be moved into its own pull request, but didn't want to split these issues so finely as they are all quite small.

Can split all of these into separate pull requests but wasn't sure if spamming the pull requests would be best.
Alex

### DIFF
--- a/src/crates/lib.md
+++ b/src/crates/lib.md
@@ -2,7 +2,7 @@
 
 Let's create a library, and then see how to link it to another crate.
 
-```rust,editable
+```rust,ignore
 pub fn public_function() {
     println!("called rary's `public_function()`");
 }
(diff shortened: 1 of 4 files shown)


marioidival:
@alexander-clarke thank you!

PR #1176, opened 2019-04-19
For https://github.com/rust-lang/rust-by-example/issues/1175

### LINKED ISSUE #1175: flow_control/while.html is confusing
Hi,

Thank you for Rust By Example.
https://doc.rust-lang.org/stable/rust-by-example/flow_control/while.html reads "The while keyword can be used to loop until a condition is met." That is confusing - it sounds more like do...until() or until()...end loop from other languages.

Suggest changing to "The while keyword can be used to run a loop while a condition is true." (Based on https://doc.rust-lang.org/stable/book/ch03-05-control-flow.html#conditional-loops-with-while.)

peter-kehl:
closes #1175

### DIFF
--- a/src/flow_control/while.md
+++ b/src/flow_control/while.md
@@ -1,6 +1,6 @@
 # while
 
-The `while` keyword can be used to loop until a condition is met.
+The `while` keyword can be used to run a loop while a condition is true.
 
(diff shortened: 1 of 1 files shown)


marioidival:
@peter-kehl nice!

PR #1178, opened 2019-04-19
19.2 Vectors Error in Code Example

sebadoh:
https://doc.rust-lang.org/1.29.2/rust-by-example/std/vec.html

Code contain an error in the following places:

Line number 3: let collected_iterator: Vec = (0..10).collect();
Should be: let mut collected_iterator: Vec = (0..10).collect();
Or else Line number16 can't borrow mutably


marioidival:
@sebadoh thank you!

PR #1180, opened 2019-04-20
For https://github.com/rust-lang/rust-by-example/issues/1179

### LINKED ISSUE #1179: scope/lifetime/lifetime_bounds.md to clearly say it uses "where" keyword
The documentation part of https://doc.rust-lang.org/rust-by-example/scope/lifetime/lifetime_bounds.html (before the code example) doesn't mention `where` keyword at all. If you follow Rust by Example table of content, and you never saw `where` and `T: ...` before, then you may expect the bounds to be a part of the main generic specification, for example `fn print<T>(t: T)` or `fn print_ref<'a, T>(t: &'a T)`.

Hence, could you make the documentation (the top part) mention `where`.

peter-kehl:
closes #1179


marioidival:
@peter-kehl thank you

PR #1182, opened 2019-04-23
Link to Reference for macro_rules designators

felixrabe:


marioidival:
@felixrabe thank you again!

Issue #1184, opened 2019-04-24
flow_control/if_let.md misses an important benefit


peter-kehl:
`if let` docs and example doesn't mention a benefit: It allows to match enum non-parameterized variants, even if the enum doesn't `#[derive(PartialEq)]`, neither we implement `PartialEq` for it. In such case, classic `if Foo::Bar==a` fails, but `if let` works.

PR #1185, opened 2019-04-24
For https://github.com/rust-lang/rust-by-example/issues/1184

peter-kehl:


peter-kehl:
Hi @marioidival,
this is intended to fail. The example shows an error and asks the user to fix it. Is there a way to annotate the example so as to keep Travis accept the compile error (for that example)?

PR #1186, opened 2019-04-26
File read lines

AndyGauge:
Thought this was a good implementation (https://stackoverflow.com/a/30801708/5586783) but could use a usage example.


marioidival:
@AndyGauge thank you!

PR #1190, opened 2019-05-21
Reword "Flow Control" to "Flow of Control"

russelldavis:
It's a bit of a nit, but flow control generally refers to something different. The way it's being used here, it's usually worded as flow *of* control or control flow.

References:
https://en.wikipedia.org/wiki/Flow_control
https://en.wikipedia.org/wiki/Control_flow


marioidival:
@russelldavis thank you, 💃

PR #1195, opened 2019-05-29
Fixed wording in the `Seconds` struct description.

Rosto75:
I'm willing to be wrong on this one. The original author's intent could have been:
```rust
// `Seconds`, a tuple struct, no additional attributes
```
or
```rust
// `Seconds`, a tuple struct; no additional attributes
```
or maybe
```rust
// `Seconds`, a tuple struct - no additional attributes
```


marioidival:
@Rosto75 thank you

PR #1203, opened 2019-06-11
Add Cargo comment to the `try!` macro example.

Rosto75:
In Rust "2018", `try!` macro is deprecated, and `try` is a reserved keyword. If you're using Cargo to compile the code using this syntax, you'll get an error message. To fix it, you need to change the value of the `edition` field in the `Cargo.toml` file back to "2015".


marioidival:
@Rosto75 thank you

PR #1208, opened 2019-06-22
Update enter_question_mark.md

ralvessa:
A minor thing, but I think that, grammatically, this sounds better.


marioidival:
@ralvessa thank you!

PR #1210, opened 2019-06-25
[typo] fix unit_testing wrong output

adispring:


marioidival:
@adispring thank you!

PR #1218, opened 2019-07-12
panic.md: Fix highlighting

felixrabe:
A "```bash" section is only meant for Bash code, not for transcripts of interactive sessions including command output.

### DIFF
--- a/src/std/panic.md
+++ b/src/std/panic.md
@@ -34,7 +34,7 @@ fn main() {
 
 Let's check that `panic!` doesn't leak memory.
 
-```
+```text
 $ rustc panic.rs && valgrind ./panic
 ==4401== Memcheck, a memory error detector
 ==4401== Copyright (C) 2002-2013, and GNU GPL'd, by Julian Seward et al.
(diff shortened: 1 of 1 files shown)


marioidival:
@felixrabe it's broken.

felixrabe:
My bad, forgot that `` ``` `` equals `` ```rust `` in rustdoc. Looking into the proper way to do this correctly.

felixrabe:
Quoting https://www.google.com/search?q=rust+doc+plain+text+code:

> If you are only showing plain text, put text instead of rust after the triple ...

marioidival:
@felixrabe has conflict :(

marioidival:
@felixrabe ? =D

Rosto75:
Fixed in #1222. I used `shell` attribute instead of `text` because it works better with the console output.

marioidival:
oh, thanks @Rosto75 💃

Rosto75:
> oh, thanks @Rosto75 💃

Cheers!

PR #1219, opened 2019-07-13
Changed mutable collected_iterator into immutable

HomoEfficio:
`collected_iterator` should be immutable in order for `collected_iterator.push(0)` to produce error.


Rosto75:
Already in #1216

marioidival:
@HomoEfficio Thank you!!

PR #1220, opened 2019-07-14
Fix minor formatting inconsistencies

austinwk:
Fixes minor formatting inconsistencies.

Example:
`println!("called my_mod::nested::public_function_in_super_mod");`
deviates from the precedent
``println!("called `my_mod::nested::public_function_in_super_mod()`");``


marioidival:
@austinwk thank you

PR #1225, opened 2019-07-19
Reorder links in destructuring.md

funkill:
Move "Destructuring Tuples" to start of list, like in Summary


marioidival:
@funkill thank you

Issue #1217, opened 2019-07-11
freezing doesn't work in website


HaoXuan40404:
In https://doc.rust-lang.org/stable/rust-by-example/scope/borrow/freeze.html，example of freezing like：
```
fn main() {
    let mut _mutable_integer = 7i32;

    {
        // Borrow `_mutable_integer`
        let _large_integer = &_mutable_integer;

        // Error! `_mutable_integer` is frozen in this scope
        _mutable_integer = 50;
        println!("_mutable_integer = {}",_mutable_integer);
        // println!("_large_integer = {}",_large_integer);
        // FIXME ^ Comment out this line

        // `_large_integer` goes out of scope
    }
    println!("_mutable_integer = {}",_mutable_integer);

    // Ok! `_mutable_integer` is not frozen in this scope
    _mutable_integer = 3;
}
```

 In website, "freezing" doesn't work. but it's right in local compilation : )

dhl:
The example is definitely incorrect.

For easy reference, here's the original example from the [freezing section](https://doc.rust-lang.org/stable/rust-by-example/scope/borrow/freeze.html):
```rust
fn main() {
    let mut _mutable_integer = 7i32;

    {
        // Borrow `_mutable_integer`
        let _large_integer = &_mutable_integer;

        // Error! `_mutable_integer` is frozen in this scope
        _mutable_integer = 50;
        // FIXME ^ Comment out this line

        // `_large_integer` goes out of scope
    }

    // Ok! `_mutable_integer` is not frozen in this scope
    _mutable_integer = 3;
}
```

I think the reason why the code works (i.e. there's no error) is simply because of observablility.

As `_large_integer` is not used after the mutation to `_mutable_integer`, the change cannot be affect other code later in the scope. It is for the same reason that the following code:

```rust
fn main() {
    let mut _immutable_integer = 7;

    {
        let _mutable_borrow_1 = &mut _immutable_integer;
        let _mutable_borrow_2 = &mut _immutable_integer;
        
        // println!("{}", _mutable_borrow_1);
        // ^ uncomment to trigger compilation error
    }
}
```

also works until the println line is uncommented.

adamchalmers:
I think this issue can be closed now that [this PR](https://github.com/rust-lang/rust-by-example/pull/1226)

PR #1226, opened 2019-07-23
freeze.md: Incorrect example

dhl:
The freeze example is meant to show that a value cannot be immutably borrowed
while the original value is modified within the same scope.

However, as the immutably borrowed value is not used after the mutation, the
borrow checker allows the example to be compiled successfully, contrary to what
we intend to demonstrate.

See #1217 for discussion.

Solution: Use the immutably borrowed value after original value is modified.

### DIFF
--- a/src/scope/borrow/freeze.md
+++ b/src/scope/borrow/freeze.md
@@ -17,7 +17,7 @@ fn main() {
 
         println!("Immutably borrowed {}", large_integer);
 
-        // `_large_integer` goes out of scope
+        // `large_integer` goes out of scope
     }
 
     // Ok! `_mutable_integer` is not frozen in this scope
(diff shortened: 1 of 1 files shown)


src/scope/borrow/freeze.md
```diff
@@ -9,12 +9,14 @@ fn main() {
...
-        let _large_integer = &_mutable_integer;
+        let large_integer = &_mutable_integer;
 
         // Error! `_mutable_integer` is frozen in this scope
         _mutable_integer = 50;
         // FIXME ^ Comment out this line
 
+        println!("Immutably borrowed {}", large_integer);
+
         // `_large_integer` goes out of scope
```
marioidival: change the name of the variable.

dhl: @marioidival thanks for catching this. Just pushed a fix for this.

marioidival:
@dhl thank you!

PR #1228, opened 2019-07-25
Snake didn't deserve to die 🐍

josephrocca:


marioidival:
@josephrocca thank you

PR #1229, opened 2019-07-28
Make `point` consistent

funkill:
In changes lines "point" makes like in 38 line (``// Error! Can't borrow `point` as immutable because it's currently``)


marioidival:
@funkill thank you

PR #1233, opened 2019-08-05
flow_control/match/binding.md: `...' -> `..='

rusty-snake:
`...` range patterns are deprecated and only kept for backward compatibility (https://doc.rust-lang.org/reference/patterns.html#range-patterns) so they souldn't be used in new code or when you learn rust.

To get compiler warnings compile local with:
`rustc --edition=2018 -W ellipsis-inclusive-range-patterns match.rs` or
`cargo rustc -- --warn  ellipsis-inclusive-range-patterns `


marioidival:
@rusty-snake thank you

PR #1242, opened 2019-08-10
Add colon after "See Also"

funkill:
In many files after "See Also" placed colon, but some files don't have this colon.


marioidival:
@funkill thank you!

PR #1243, opened 2019-08-18
macros.md: Changing Sometime to more appropriate adverb form Sometimes

KingAkeem:
Previously using adjective instead of the appropriate adverb.


marioidival:
@KingAkeem thank you!

PR #1247, opened 2019-09-06
fix 1037: add the TryFrom chapter back in

adamchalmers:
Commit 2a808e6a3d25a26702a4a4a08568904764ae1bea reverted the chapter on TryFrom, because the TryFrom trait was unstabilized. But now it's stabilized again, so we should put the chapter back in. This fixes issue 1037.


marioidival:
@adamchalmers thank you!

PR #1251, opened 2019-09-10
Fix #1060: add page on Impl Trait

### LINKED ISSUE #1060: update for `impl Trait`
This issue tracks inclusion of information about impl Trait into the Book. This will happen in the 2018 edition.

Tracking issue: rust-lang/rust#34511

adamchalmers:
I added a page on `impl Trait`, and also updated the page on "how to return closures". There's no need to return boxed functions anymore, so I changed those examples to use `impl Trait`.


marioidival:
@adamchalmers Thank you!

PR #1252, opened 2019-09-11
fix #1067: explain that unit tests can return Result<()>

### LINKED ISSUE #1067: update for unit tests with a non-`()` return type
This issue tracks inclusion of information about unit tests with a non-() return type into the Book. This will happen in the 2018 edition.

Tracking issue: rust-lang/rust#48854

adamchalmers:
I didn't go into the details of Termination because I don't think that's relevant to people who just wanna use ? in their unit tests. I did link to the Edition Guide where it explains Termination, if people want that detail they can read it there.


marioidival:
@adamchalmers Thank you very much!

PR #1255, opened 2019-09-13
Fix #1253: Document enum type aliases

### LINKED ISSUE #1253: Document Self::Variant for enums
In Rust 1.37 you can use [type aliases for enums](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1370-2019-08-15). I think it's worth showing people how that works in the page on Enums.

In my code, I only ever use this feature when I want to refer to an enum by `Self` instead of by its name (usually if the name is very long). I'm planning to use that as my example and link to the release notes/RFC if readers want more info.

adamchalmers:


marioidival:
@adamchalmers thank you!

PR #1259, opened 2019-09-18
Fix #90: Add supertraits and Fully Qualified syntax

### LINKED ISSUE #90: Extend traits chapter
- [ ] supertraits `FooBar: Foo + Bar`
- [ ] trait objects `Box<Foo>`, `&Foo`
- [ ] "function overloading" using double dispatch: `Vec` \* `Vec` and `Vec` \* `int`

adamchalmers:
The issue also mentions 
> "function overloading" using double dispatch: Vec * Vec and Vec * int

But to be honest, I have no idea what that means! So I'm opening up this request for now. Happy to document it with an example if someone can link me to what language feature they're talking about.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -139,6 +139,8 @@
     - [Iterators](trait/iter.md)
     - [`impl Trait`](trait/impl_trait.md)
     - [Clone](trait/clone.md)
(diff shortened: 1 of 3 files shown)


marioidival:
@adamchalmers Thank you!

PR #1263, opened 2019-09-19
Move chapters from folder "traits" to "trait"

funkill:
Chapters "Supertrait" and "Disambiguating overlapping traits" belongs to "Traits" section. Currently all chapters of section "Traits" saved in folder `trait`, not in `traits`.


marioidival:
@funkill thank you!

PR #1267, opened 2019-09-20
Added type alias enum variant rfc

funkill:
Empty brackets in "See Also:" changed to link to type alias enum variant rfc. Another way - drop this brackets.


marioidival:
@funkill thank you

PR #1269, opened 2019-09-23
Fix match range pattern usage

agaskell:
I was working through the [match example](https://doc.rust-lang.org/rust-by-example/flow_control/match.html) and received this compiler warning:

```
warning: `...` range patterns are deprecated
  --> src/main.rs:12:11
   |
12 |         13...19 => println!("A teen"),
   |           ^^^ help: use `..=` for an inclusive range
   |
   = note: #[warn(ellipsis_inclusive_range_patterns)] on by default
```


marioidival:
@agaskell Thank you

PR #1270, opened 2019-09-24
Rectangle ambiguity

### LINKED ISSUE #1170: After reading the book, RBE's definition of Rectangle in 3.1 is un-intuitive
Disclaimer: This is just feedback on the content from the perspective of a learner, not a "real" bug or issue.

I found quite a bit of advice online recommending "the book", followed by Rust By Example to get your feet wet with understanding rust. The book defines a rectangle in Chapter 5 as a height and width. RBE defines a rectangle as 2 points in space. Then, similar to the book, RBE asks the reader to implement a function that computes the area of a rectangle. It actually took me a bit of googling to figure out how a rectangle could be defined in such a way.

It felt un-intuitive to me to have these two official resources diverge in this way, and came across as confusing to me. (I found myself asking, "Why am I googling the different ways that a Rectangle can be represented in code on a day where I sat down to practice some rust concepts I learned from the book"

### LINKED ISSUE #1194: Activity seems misleading on Custom Types -Structures
> Activity
> Add a function rect_area which calculates the area of a rectangle (try using nested destructuring).

https://github.com/rust-lang/rust-by-example/blob/master/src/custom_types/structs.md

Given the example, I'm not too sure how we should be calculating the area or the rectangle.  

I've made an abridged version of the question/answer:

```
use std::fmt::{self, Formatter, Display};

struct Point {
    x: f32,
    y: f32,
}

struct Rectangle {
    p1: Point,
    p2: Point,
}

impl Display for Point {
    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
        write!(f, "x: {}, y:{}", self.x, self.y)
    }
    
}
impl Display for Rectangle {
    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
        write!(f, "(p1: {}, p2: {})", self.p1, self.p2)
    }
    
}

 fn rect_area(rect: Rectangle) -> f32 {
    // Is this what you were looking for? seems misleading if so
    // (rect.p1.x - rect.p2.x)*(rect.p1.y - rect.p2.y)
 }
fn main() {
    let point: Point = Point { x: 0.3, y: 0.4 };
    let Point { x: my_x, y: my_y } = point;

    let _rectangle = Rectangle {
        p1: Point { x: my_y, y: my_x },
        p2: point,
    };

    println!("{}", _rectangle);
   //println!("{}", rect_area(_rectangle));
}
```

andymac-2:
Multiple users (#1170 and #1194) have identified that the `Rectangle` struct composed of two points is confusing. This particular structure is referenced in two places: the page on boxes, and the page on structs. The fields are renamed to make the structure of the rectangle more obvious.

Fixes #1170 
Fixes #1194


marioidival:
@andymac-2 thank you!

PR #1274, opened 2019-09-30
Add reference to lifetime in structs

illiteratewriter:
Lifetimes were not introduced before structures and therefore 
caused confusion as to the origin of `<'a>`


marioidival:
@illiteratewriter thank you

PR #1277, opened 2019-10-05
Explicit mention of slice range meaning

kishaningithub:
In `&ys[1 .. 4]` i did not understand whether 4 meant the length or the end index. This is to prevent other readers from having the same confusion.


marioidival:
@kishaningithub thank you

PR #1283, opened 2019-10-18
refactor: simplify extracting Result from Option

kaiserd:
Using the ?, the code basically extracted an Option from the Result we just built.
 Just to again wrap it in a Result.
 We can just return the Result.


kaiserd:
It should be added that the current version would be necessary if the
the function signature was

`fn double_first(vec: Vec<&str>) -> Result<Option<i32>, Box<std::error::Error>>`

to coerce the error error type.

marioidival:
@kaiserd

PR #1286, opened 2019-10-28
Use plural in section heading

stomar:
Alternatively, "Create a Library".


marioidival:
Sorry but this chapter is correct using just `Library`.

PR #1301, opened 2019-12-27
eliminate the warnings

bud-adamas:
if there is no `dyn`, there are warnings like this:

>         warning: trait objects without an explicit `dyn` are deprecated
>         --> xxx.rs:5:45
>         |
>         5 | type Result<T> = std::result::Result<T, Box<error::Error>>;
>         |                                             ^^^^^^^^^^^^ help: use `dyn`: `dyn error::Error`
>         |
>         = note: `#[warn(bare_trait_objects)]` on by default
> 
>         warning: trait objects without an explicit `dyn` are deprecated
>         --> xxx.rs:21:32
>         |
>         21 |     fn cause(&self) -> Option<&error::Error> {
>         |                                ^^^^^^^^^^^^ help: use `dyn`: `dyn error::Error`

### DIFF
--- a/src/error/multiple_error_types/boxing_errors.md
+++ b/src/error/multiple_error_types/boxing_errors.md
@@ -13,7 +13,7 @@ use std::error;
 use std::fmt;
 
 // Change the alias to `Box<error::Error>`.
-type Result<T> = std::result::Result<T, Box<error::Error>>;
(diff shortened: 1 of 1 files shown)


marioidival:
@bud-adamas thank you!

PR #1302, opened 2020-01-09
CamelCase -> UpperCamelCase

Cawibo:
The compiler will raise a warning unless it is UpperCamelCase. Since the warning is explicit about mentioning "upper", maybe the guide should be aswell?


marioidival:
@Cawibo thank you

PR #1316, opened 2020-03-14
Fight the confusion over 'static

najamelan:
Many people learning Rust are confused by the fact that owned values are `'static`. I feel the description here as well as in _the book_ are in part responsible for that. 

I hope that this change makes things a lot clearer. Note that I have intersected the explanation before the existing example because if It's underneath, it will 
be offscreen without scrolling. I kind of feel this is to important to be hidden offscreen, even if I tried to reformulate the original text to be more accurate.

I would propose we put this explanation in _the book_ as well, in the section about 'static lifetimes.

For an example of what can happen if people are confused about lifetime bounds (and use unsafe): https://github.com/libra/libra/pull/1949

ps: edited on Github, so I didn't run this through mdbook and doctests... but the example was verified on the playground.

### DIFF
--- a/src/scope/lifetime/static_lifetime.md
+++ b/src/scope/lifetime/static_lifetime.md
@@ -1,15 +1,61 @@
 # Static
 
-A `'static` lifetime is the longest possible lifetime, and lasts for 
-the lifetime of the running program. A `'static` lifetime may also be 
-coerced to a shorter lifetime. There are two ways to make a variable 
-with `'static` lifetime, and both are stored in the read-only memory
-of the binary:
+`'static` is one of the reserved lifetime names. It has a special meaning
+in Rust. A __reference__ that has a `'static` lifetime is guaranteed to be 
+valid for the entire lifetime of the running program. 
+
+There are two ways to create data that lives for the entire lifetime of
+a program and both are stored in the read-only memory of the binary:
 
 * Make a constant with the `static` declaration.
 * Make a `string` literal which has type: `&'static str`.
 
-See the following example for a display of each method:
+__Warning__: It's easy to get confused about `'static` because all variable
+bindings that own their data, eg. do not contain references, fulfill a 
+bound on `'static`:
+
+```rust,editable,compile_fail
+use std::fmt::Debug;
+
+fn print_it( input: impl Debug + 'static )
+{
+    println!( "'static value passed in is: {:?}", input );
+}
+
+fn use_it()
+{
+    let i = 5;     // i is owned and contains no references, thus it's 'static
+    print_it( i ); // valid call
+    
+    // oops, &i only has the lifetime defined by the scope of 
+    // use_it(), so it's not 'static
+    print_it( &i ); 
+}
+```
+
+
+The compiler will tell you:
+```ignore
+error[E0597]: `i` does not live long enough
+  --> src/lib.rs:15:15
+   |
+15 |     print_it( &i ); 
+   |     ----------^^--
+   |     |         |
+   |     |         borrowed value does not live long enough
+   |     argument requires that `i` is borrowed for `'static`
+16 | }
+   | - `i` dropped here while still borrowed
+```
+
+A bound on `'static` means that the callee does not have to worry
+about their variable binding becoming invalid after some time.
+They can use it as long as they want. In contrast to a `'static` 
+__reference__, it is unrelated to the runtime of the entire program.
+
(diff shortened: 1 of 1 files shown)


najamelan:
bors r? @steveklabnik 

It seems I modified your writing, so I suppose you probably want to have a look at whether you're ok with the change.

src/scope/lifetime/static_lifetime.md
```diff
@@ -1,15 +1,61 @@
 # Static
 
-A `'static` lifetime is the longest possible lifetime, and lasts for 
-the lifetime of the running program. A `'static` lifetime may also be 
-coerced to a shorter lifetime. There are two ways to make a variable 
-with `'static` lifetime, and both are stored in the read-only memory
-of the binary:
+`'static` is one of the reserved lifetime specifiers. It has a special meaning
+in Rust. A __reference__ that has a `'static` lifetime is guaranteed to be 
+valid for the entire lifetime of the running program.
```
steveklabnik: This is not quite correct. It *may* live for the entire lifetime, but does not have to.

src/scope/lifetime/static_lifetime.md
```diff
@@ -1,15 +1,61 @@
...
-the lifetime of the running program. A `'static` lifetime may also be 
-coerced to a shorter lifetime. There are two ways to make a variable 
-with `'static` lifetime, and both are stored in the read-only memory
-of the binary:
+`'static` is one of the reserved lifetime specifiers. It has a special meaning
+in Rust. A __reference__ that has a `'static` lifetime is guaranteed to be 
+valid for the entire lifetime of the running program. 
+
+There are two ways to create data that lives for the entire lifetime of
+a program and both are stored in the read-only memory of the binary:
```
steveklabnik: This is also not the only two ways: owned values are `'static`.

src/scope/lifetime/static_lifetime.md
najamelan: Do you mean that if I had an API that takes in a `&'static Something` it would possibly stop being valid at some point before the end of the program? That seems problematic, no? How do you know when?

src/scope/lifetime/static_lifetime.md
najamelan: The two points where there in the original text. The PR tries to make the distinction between `'static` as a bound, which passes for owned data, and `'static` on a reference. That's the whole point of a common confusion. Owned data AFAICT won't live for the entire lifetime of the program unless it is in a global. Otherwise it's always scoped, maybe to `fn main()` but still scoped.

najamelan:
Looking over it. Maybe the mistake is to try to fix the original text with minimal changes. The ways to make data that lives for the entire lifetime isn't very relevant unless you are taking a reference to them. So that might still be confusing. Maybe it's worth considering rewriting it from scratch with a little list of what we want to convey. 

All in all, because of the different meaning of `'static` as a bound, plus the existence of `static` as a keyword, makes the concept more complicated in rust that it seems at first sight. Maybe that warrants a bit more in depth explanation, but that probably is more suited for the book than for rust by example.

src/scope/lifetime/static_lifetime.md
steveklabnik: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c7241593e258fdc2ba77c6b603c15bb5

steveklabnik:
>  Maybe it's worth considering rewriting it from scratch with a little list of what we want to convey.

Yeah, that may be true! Like I said, appreciate you taking a crack at it here. Hm.

src/scope/lifetime/static_lifetime.md
najamelan: Ok, but now we are confusing with the lifetime of the variable binding of the reference. I generally assumed that when we speak about reference lifetimes, we mean the lifetime of the pointed to data and the one we name as in `&'a SomeValue` -> `'a` in this case. The text says "the **reference** is guaranteed to be valid", as opposed to "The reference lives for ...". But if you feel that that is still not clear enough, I will propose a complete rewrite. I am generally in favor of being very precise and thorough, but often my documentation PR's get downplayed, so I didn't necessarily proposed the way I like it best but more as what I estimate the general opinion might be. It will probably be a fair bit longer if trying to be complete. I suppose also that in RBE we prefer showing code rather than long descriptions?

najamelan:
See follow up in #1320

PR #1320, opened 2020-03-23
Rewrite section on 'static lifetime.

najamelan:
This tries to be more precise and in specific help avoid confusion about `'static` trait bounds. This is a follow up from #1316 .

How do you feel about this?

bors r? @steveklabnik


najamelan:
Just thinking, it might be worth mentioning here as well that Box defaults to `'static`. Another thing most people discover the hard way.

Issue #108, opened 2014-06-17
Topics for the "rustdoc" chapter


japaric:
- [ ] Which comment styles `///` get translated to docs
- [ ] How to use the `rustdoc` command
- [ ] `#[doc]` attributes: `inline`, `no_inline`, `hidden`
- [ ] [Playpen](http://play.rust-lang.org) integration
- [ ] Do we have "[doctests(https://docs.python.org/2/library/doctest.html)]"? i.e. examples in the docs that get tested to match some expected output
- [ ] The [std docs](http://static.rust-lang.org/doc/master/std/index.html) have been generated using `rustdoc`!

steveklabnik:
I'd be happy to mentor anyone who wants to work on this!

frewsxcv:
somewhat relevant: we have a rustdoc resource we can link to now: https://doc.rust-lang.org/stable/rustdoc/

jsjoeio:
@steveklabnik it's been a few years 😂 Are you still willing to mentor someone on this? 

I'd be up for giving this a shot.

steveklabnik:
Hi @jsjoeio !

So, I am in the sense that I can help you with the mechanics of how to contribute, etc. But I haven't worked on RBE for a long time, so I am not entirely sure that the list above makes sense. That is, I don't know what stuff has landed and what stuff hasn't, if the list above is still relevant, etc.

Is there something specific that made you pick this issue? If so, we can dive into that and I can help you out for sure!

jsjoeio:
Thanks for the speedy reply @steveklabnik! 

> That is, I don't know what stuff has landed and what stuff hasn't, if the list above is still relevant, etc.

Good to know. That might be a good first step then - identifying what's landed and what hasn't. 

> Is there something specific that made you pick this issue? 

Nope! I chose it because
- it had the label "E-mentor"
- no one else had grabbed it
- docs seemed like a good entry point into Rust OSS

## Next Steps

I think what I'll do then is get started on this. Here's what I'm thinking:
- [ ] familiarize myself with RBE
- [ ] identify what's landed and what hasn't 

Afterwards, I'll share thoughts and questions here and we can go from there. How does that sound?

jsjoeio:
## Audit

Here's a summary of what's landed and what hasn't:

### ✅ Which comment styles `///` get translated to docs

[...]

PR #1321, opened 2020-03-25
fix: make spelling/style of rustdoc consistent

jsjoeio:
On the "[Documentation](https://doc.rust-lang.org/rust-by-example/meta/doc.html?highlight=rustdoc#doc-comments)" page, there is inconsistency for how "rustdoc" is spelled. 

There are three versions being used:
- Rustdoc
- rustdoc
- `rustdoc`

This PR fixes that by making them all `rustdoc` (where it makes sense) as part of #108

### DIFF
--- a/src/meta/doc.md
+++ b/src/meta/doc.md
@@ -9,7 +9,7 @@ These commands will appropriately invoke `rustdoc` (and `rustc`) as required.
 ### Doc comments
 
 Doc comments are very useful for big projects that require documentation. When
-running Rustdoc, these are the comments that get compiled into
+running `rustdoc`, these are the comments that get compiled into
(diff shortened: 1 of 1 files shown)


src/meta/doc.md
```diff
@@ -68,7 +68,7 @@ $ rustdoc --test --extern doc="libdoc.rlib" doc.rs
 ### See also:
 
 * [The Rust Book: Making Useful Documentation Comments][book]
-* [The Rustdoc Book][rustdoc-book]
+* [The rustdoc Book][rustdoc-book]
```
jsjoeio: Should this one be an inline code block too? I wasn't sure

PR #1329, opened 2020-04-09
Improve the conversion example

juntao:
I added a sentence to the `conversion` chapter. It points the readers to the `types/cast` chapter if they want to convert between primitive types.

I also added an error example for `types/cast` to illustrate the limitations of primitive type casting.


marioidival:
@juntao Thank you

PR #1331, opened 2020-04-09
Capture example should use String

ScriptDevil:
Using a `&'static str` like "green" would not prevent us from using the closure `print` post the move. 

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=411cb0b9776ece079c97d820ea6733ee


marioidival:
Thank you!

PR #1337, opened 2020-04-22
Clarifies how to customize lib crate name

### LINKED ISSUE #1201: Attribute name error in the Crates section
Libraries get prefixed with "lib", and by default they get named after their crate file, but this default name can be overridden using the `crate_name` attribute.

`crate_name` should be `crate-name`

newca12:
Fix  #1201


marioidival:
@newca12 Thank you!

PR #1338, opened 2020-04-23
Clarify conditional compilation

gokhanettin:
Introduction of `cfg` attribute and `cfg!` macro operators can be confusing. `cfg!` macro and `cfg` attribute behave differently in terms of conditional compilation. While `cfg` attribute enables conditional compilation of some code, `cfg!` macro conditionally evaluates to `true` or `false` literals for run-time checks.

### DIFF
--- a/src/attribute/cfg.md
+++ b/src/attribute/cfg.md
@@ -1,11 +1,13 @@
 # `cfg`
 
-Conditional compilation is possible through two different operators:
+Configuration conditional checks are possible through two different operators:
 
 * the `cfg` attribute: `#[cfg(...)]` in attribute position
 * the `cfg!` macro: `cfg!(...)` in boolean expressions
 
-Both utilize identical argument syntax.
+While the former enables conditional compilation, the latter conditionally
+evaluates to `true` or `false` literals allowing for checks at run-time. Both
+utilize identical argument syntax.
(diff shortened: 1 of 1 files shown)


gokhanettin:
[The Result Reference](https://doc.rust-lang.org/reference/conditional-compilation.html) also describes `cfg!` macro as one of the ways of conditional compilation. But I still find it a little bit confusing.

For the attribute case, compiler decides what to compile at compile time, which I would call conditional compile. For the macro case, the macro evaluates to either true or false literals at compile time, which can then be used at run-time. This is not something special about `cfg!` macro I guess. All macros evaluate to something during compile time.

Obviously following `cfg.rs` won't compile. Neither `rustc cfg.rs` nor `rustc cfg.rs --cfg "some_condition"` works as one of the functions is not (conditionally) compiled.

```rust
#[cfg(some_condition)]
fn needs_some_condition() {
}

#[cfg(not(some_condition))]
fn not_needs_some_condition() {
}

fn main() {
    if cfg!(some_condition) {
        needs_some_condition();
    }

    if cfg!(not(some_condition)) {
        not_needs_some_condition();
    }
}
```

marioidival:
@gokhanettin Thank you for you clarification

PR #1340, opened 2020-04-23
Fix Example -> Examples

gokhanettin:
As suggested by RFC505, use the plural form: "Examples" rather than
"Example" even for one example.


marioidival:
@gokhanettin u find more occurrences?

gokhanettin:
There are no other occurrences. Checked with

```
grep -RE "/// # Example" src
grep -RE "/// # Panic" src
grep -RE "/// # Error" src
grep -RE "/// # Argument" src
```

PR #1343, opened 2020-04-26
Fix empty bound examples

### LINKED ISSUE #1341: Ord as an empty bound example
From Testcase: empty bounds

> A consequence of how bounds work is that even if a trait doesn't include any functionality, you can still use it as a bound. `Eq` and `Ord` are examples of such traits from the std library.

Unlike `std::cmp::Eq`, `std::cmp::Ord` doesn't look like an empty bound in the [doc](https://doc.rust-lang.org/std/cmp/trait.Ord.html). Is it a correct example for empty bounds?

If it is not, we can change with `std::marker::Sync` and `std::marker::Send` maybe?

gokhanettin:
Unlike `std::cmp::Eq`, `std::cmp::Ord` is not an empty bound. Therefore,
replace it with an empty bound, `std::marker::Copy`.

Closes #1341

### DIFF
--- a/src/generics/bounds/testcase_empty.md
+++ b/src/generics/bounds/testcase_empty.md
@@ -2,7 +2,7 @@
 
 A consequence of how bounds work is that even if a `trait` doesn't
 include any functionality, you can still use it as a bound. `Eq` and
-`Ord` are examples of such `trait`s from the `std` library.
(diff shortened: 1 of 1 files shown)


marioidival:
@gokhanettin Thank you!

PR #1346, opened 2020-05-09
Fix link of formatting traits

hsi:
According to the actual context (print formatting), the author probably wanted to reference the "formatting traits" instead of traits as a "collection of methods".


marioidival:
@hsi thank you!

PR #1347, opened 2020-05-16
Gramatical fix in std/rc.md

romanlevin:


marioidival:
@romanlevin thank you

PR #1348, opened 2020-05-25
Person of age 0 is alive

alygin:
A person is of age 0 right after he or she was born, and before the first birthday.


marioidival:
@alygin thank you!

PR #1349, opened 2020-06-08
This explanation incorrectly inverts the meaning of SuperTrait

zancas:
@AloeareV and I noted this discrepancy with the [Rust lang reference definition of Supetraits](https://doc.rust-lang.org/stable/reference/items/traits.html?highlight=SuperTrait#supertraits).

They can't both be correct, (they're inverted!) we prefer the language reference definition.

### DIFF
--- a/src/trait/supertraits.md
+++ b/src/trait/supertraits.md
@@ -8,7 +8,7 @@ trait Person {
     fn name(&self) -> String;
 }
 
-// Student is a supertrait of Person.
+// Person is a supertrait of Student.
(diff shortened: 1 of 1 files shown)


zancas:
Also, this:  https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#using-supertraits-to-require-one-traits-functionality-within-another-trait

Matches the explanation in the reference, which conflicts with the explanation in Rust by example.

marioidival:
@zancas thank you!

PR #1357, opened 2020-06-23
Remove mention of `try!` in `Display` example

rossmacarthur:
There is already a detailed explanation of what the `try!` macro is in `src/error/result/enter_question_mark.md` and it is confusing to introduce it here.

### DIFF
--- a/src/hello/print/print_display/testcase_list.md
+++ b/src/hello/print/print_display/testcase_list.md
@@ -13,14 +13,6 @@ Using `?` on `write!` looks like this:
 write!(f, "{}", value)?;
 ```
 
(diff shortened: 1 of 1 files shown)


marioidival:
It's okay for me to remove the explanation, but not to remove it in the comments. It is always good to remember all the possibilities when possible. @rossmacarthur

rossmacarthur:
@marioidival The `try!` macro is deprecated. It seemed strange to me that it was mentioned here since the `try!` macro was not even being used in this code.

This is the only case of the `try!` macro being mentioned or used outside of the error handling section even though the question mark operator is obviously used everywhere.

PR #1367, opened 2020-08-08
Fix "`0.2`, is not a valid semver version"

### LINKED ISSUE #1317: Build instructions in README cause error
The build instructions in README for using RBE locally specify a version, causing the following error:
```
cargo install mdbook --version 0.2 --force 
    Updating crates.io index
error: the `--vers` provided, `0.2`, is not a valid semver version: cannot parse '0.2' as a semver

if you want to specify semver range, add an explicit qualifier, like ^0.2

```

In the Contribution Guidelines, the build instructions do not specify version. This procedure seems to work fine.

xiaochuanyu:
Fixes #1317.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ If you'd like to read it locally, [install Rust], and then:
 ```bash
 $ git clone https://github.com/rust-lang/rust-by-example
 $ cd rust-by-example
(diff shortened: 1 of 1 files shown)


xiaochuanyu:
Already fixed.

Issue #1358, opened 2020-06-24
Drop extern crate


butlermatt:
Hi Folk,

I'm still new to Rust, and when looking at Rust by Example section 11.2 it indicates that:

> To link a crate to this new library, the extern crate declaration must be used. This will not only link the library, but also import all its items under a module named the same as the library. The visibility rules that apply to modules also apply to libraries.

However I was under the impression that as of Rust 2018, the extern crate functionality was no longer _required_ (eg: the "must" key word here is no longer valid, and perhaps "may" is more applicable). Since 2015 and older syntax still use the extern crate, it is probably important to mention this functionality but perhaps demonstrate how it has now been superseded by the ability to import with use statements. In fact "The Book" now specifically states that it is "unidiomatic" to use extern crate. (https://doc.rust-lang.org/reference/items/extern-crates.html in the "Edition Differences" aside).

I feel these two areas should be brought in alignment with each other.

steveklabnik:
Yes, this should be done. I'd happily accept a PR to do it.

xiaochuanyu:
@steveklabnik I opened a PR to address this issue. I'm somewhat new to Rust so I would really appreciate your feedback on the PR if anything is wrong.

PR #1369, opened 2020-08-09
Update extern crate related sections

xiaochuanyu:
As discussed in https://github.com/rust-lang/rust-by-example/issues/1358 `extern crate` is no longer required when using an external crate in edition 2018. 
This PR adjusts sections that reference to `extern crate` accordingly.


src/crates/using_lib.md
```diff
@@ -0,0 +1,38 @@
...
+    //rary::private_function();
+
+    rary::indirect_access();
+}
+```
+
+```txt
+# Where library.rlib is the path to the compiled library, assumed that it's
+# in the same directory here:
+$ rustc executable.rs --extern rary=library.rlib --edition=2018 && ./executable
```
xiaochuanyu, line 23: Note that I added `--edition=2018`. `rustc` currently still defaults to `2015` (`cargo new` defaults to `2018`).

marioidival:
@xiaochuanyu thank you

PR #1372, opened 2020-08-18
prefer `length` over `size` when talking about number of elements vs. bytesize

michaelkirk:
These terms can be synonyms, so what's there isn't wrong, but `length` is a less loaded term.

e.g. the example:

    [T; size]

_might_ be misconstrued as:

``` 
// two float32's has a size of 8 bytes
let my_array: [float32; 8] = [1.0, 2.0]; // wrong
```


marioidival:
@michaelkirk Thank you!

PR #1375, opened 2020-08-27
Prevent duplicated error information in "wrapping errors" example

d-e-s-o:
The "Wrapping errors" example, which illustrates how to wrap errors, is
using the wrapped error in the text that is displayed as well as the
implementation of source(). In so doing it effectively duplicates the
error information. Such behavior does not play nice with error reporting
crates such as anyhow, eyre, and others that make use of both this data.
It is also questionable at best from a logical point of view, because,
as the name suggests, the source should be a lower level error that is
the cause of this one.
With this change we update the documentation, suggesting the usage of a
custom error string describing at a higher level what went wrong. To
keep the example simple, we did not include the actual string in the
error (a comment already suggests that this could be done), although
that probably should be done in real life to provide additional context.
An alternative approach would be to not return a source, but I figured
that it's sort of desired for the illustration of wrapping an error.

### DIFF
--- a/src/error/multiple_error_types/wrap_error.md
+++ b/src/error/multiple_error_types/wrap_error.md
@@ -4,6 +4,7 @@ An alternative to boxing errors is to wrap them in your own error type.
 
 ```rust,editable
 use std::error;
+use std::error::Error as _;
 use std::num::ParseIntError;
 use std::fmt;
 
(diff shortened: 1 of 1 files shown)


d-e-s-o:
See for instance [this pull request](https://github.com/bytecodealliance/wasmtime/pull/532) that fixes exactly such a duplication of error information.

d-e-s-o:
Any chance this change could get a review?

d-e-s-o:
r?

marioidival (approved):

PR #1393, opened 2020-12-01
Drop extern crate

falc0nsec:
Hello,

I'm pretty new to Rust so apologize in advaced if Im wrong. Since version 2018, it is not longer required to use extern crate. This is similar to 
[Issue 1358](https://github.com/rust-lang/rust-by-example/issues/1358)


marioidival (approved):

marioidival:
@falc0nsec thank you!

PR #1396, opened 2020-12-15
Simplify the call of filter_map

prehistoric-penguin:


marioidival:
@prehistoric-penguin thank you!

PR #1411, opened 2021-02-11
Fix for.md to use mdbook-runnable on ignored example

jeckersb:
Without this the user cannot run the example even after addressing the FIXME


jeckersb:
Looks like this was implemented and merged in duplicate #1415

PR #1412, opened 2021-02-12
iter_result: add suggestion using `itertools::process_results`

xu-cheng:


xu-cheng:
Can I use third party crates in rust by example?

PR #1413, opened 2021-02-16
Update function name in comment

abdnh:
A minor nitpick: a function in the static lifetime example code was renamed from `use_it` to `main` in https://github.com/rust-lang/rust-by-example/pull/1383 but a comment mentioning the function just a couple of lines below it was not updated.


marioidival:
@abdnh thank you

PR #1415, opened 2021-02-24
Make flow_control/for/.into_iter() example run

EmilLuta:
For: `https://doc.rust-lang.org/rust-by-example/flow_control/for.html`,
the example containing `.into_iter()` the suggestion is to comment the
line trying to print the vector, after move. This code throws compilation
error.
In order to fix this, commit `4cc85655b97be83acbca5d036cf679f39ea66eb8`
has been introduced. This indeed makes the example pass the tests, but
disables the run option. Now, if I want to learn, I need to copy paste
the example in my own editor and compile it, rather than being able to
click run, comment as suggested and click run again.

This commit fixes this issue.
Alongside, it makes the style of playpen attributes consistent with the
rest of the book (majority, seems `rust,editable` is preferred over
`rust, editable`).

### DIFF
--- a/src/flow_control/for.md
+++ b/src/flow_control/for.md
@@ -60,7 +60,7 @@ within.
 * `iter` - This borrows each element of the collection through each iteration.
   Thus leaving the collection untouched and available for reuse after the loop.
 
-```rust, editable
+```rust,editable
 fn main() {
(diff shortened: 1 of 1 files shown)


marioidival:
@EmilLuta make sense, thank you!

PR #1419, opened 2021-02-28
update struct.md

4geru:
* struct types seems three types at first time.

```
* Tuple structs, which are, basically, named tuples
* The classic [C structs][c_struct]
* Unit structs, which are field-less, are useful for generics.
```

but struct type is only two types.


sumade123 (commented):
The second presumably refers to just "struct Regular {..." from here https://doc.rust-lang.org/std/keyword.struct.html
so there are three flavors of struct

PR #1420, opened 2021-02-28
fix name from rectangle to Rectangle

4geru:
rectangle is not defined in sample code.

it is seems `Rectangle` or `_rectangle`. but `_rectangle` is variable name. Struct name better than variable name.


marioidival:
@4geru thank you

PR #1432, opened 2021-04-08
fix: HashSet::insert() an already present value example code

Jchao:
example code: use  if to judge insert result, no panic!


marioidival (commented):
The purpose of this code is for the reader to see that if you remove this, it will work. Your PR is removing the student's ability to see this problem in action.

marioidival:
@Jchao thank you my friend

PR #1440, opened 2021-05-20
Update structs.md

tylerturk:
It seems like the variables here are flipped. When implementing the rectangle area method, I noticed that my area should be 0 and discovered this.


marioidival (approved):

marioidival:
@tylerturk thank you!

PR #1453, opened 2021-07-23
Grammatical mistake: Comparison as ... as the

J-Arun-Mani:
The sentence: This will convert the string into the type specified **so long as the** [`FromStr`]
trait is implemented for that type.
Misses an _as_: so **as long as the**
Thank you!


marioidival:
@J-Arun-Mani thank you

Abdullah8006 (commented):
The final statement should be 

> This will convert the string into the type specified as long as the [`FromStr`]

PR #1458, opened 2021-09-01
Fix link to "integration testing" page

jrodalo:
The old link was pointing to `unit_testing.md` instead of `integration_testing.md`

https://doc.rust-lang.org/rust-by-example/cargo/test.html


marioidival:
@jrodalo thank you

PR #1462, opened 2021-09-19
str: The number of #s delimiting a raw string is limited to 65535

andersk:
https://github.com/rust-lang/rust/blob/308fc2322bf00b5dc12454489679de1420320f56/compiler/rustc_lexer/src/lib.rs#L690


marioidival:
@andersk thank you!

PR #1465, opened 2021-09-30
Close #1464

### LINKED ISSUE #1464: impl Trait playground example bug
### What needs to be fixed?
Running the last playground example produces an error.

### Page(s) Affected
https://doc.rust-lang.org/rust-by-example/trait/impl_trait.html
(end of page)

### Suggested Improvement
Unsure.

mautamu:
Adds a `main` function to the last impl Trait example that demonstrates the expected behavior. Closes #1464.

Best,
Mautamu


marioidival:
@mautamu thank you

PR #1466, opened 2021-09-30
Fix incorrect formatted print hint; closes #1459.

### LINKED ISSUE #1459: Rust by Example seems to show incorrect hint for Formatted Print
### What needs to be fixed?
Disclaimer: I am a complete newbie to Rust, so I might be missing something.
I am following the Rust by Example Guide and got stuck on `1.2.3 - Hello World/Formatted print/Formatting`, because the HEX code I was supposed to print out was padded using spaces instead of 0s, when following the second hint at the bottom of the page: `You can pad with zeros to a with of 2 with :02`.
After some googling, I found out that it actually is `:0>2` instead of `:02`.
As said before, I might be missing something, but in my code (which I'll include below), `:02` does not work.
```rust
#[derive(Debug)]
struct Color {
    red: u8,
    green: u8,
    blue: u8,
}

impl Display for Color {
    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
        let red_hex = format!("{:X}", self.red);
        let green_hex = format!("{:X}", self.green);
        let blue_hex = format!("{:X}", self.blue);

        write!(f, "RGB ({red}, {green}, {blue}) 0x{redHex:0>2}{greenHex:0>2}{blueHex:0>2}",
                                red = self.red,
                                green = self.green,
                                blue = self.blue,
                                redHex = red_hex,
                                greenHex = green_hex,
                                blueHex = blue_hex)
    }
```

### Page(s) Affected

Rust by Example, 1.2.3: Formatting (https://doc.rust-lang.org/rust-by-example/hello/print/fmt.html)

mautamu:
Fix incorrectly-formatted print hint: :02 to :0>2, as per @cod-eric in #1459. Closes #1459.

Best,
Mautamu

### DIFF
--- a/src/hello/print/fmt.md
+++ b/src/hello/print/fmt.md
@@ -80,7 +80,7 @@ RGB (0, 0, 0) 0x000000
 
 Two hints if you get stuck:
  * You [may need to list each color more than once][named_parameters],
- * You can [pad with zeros to a width of 2][fmt_width] with `:02`.
+ * You can [pad with zeros to a width of 2][fmt_width] with `:0>2`.
 
 ### See also:
 


marioidival (approved):

marioidival:
@mautamu thank you

PR #1468, opened 2021-10-12
Added example of `impl Trait` as an argument

### LINKED ISSUE #1442: `impl Trait` example with argument type
The [`impl Trait`](https://doc.rust-lang.org/rust-by-example/trait/impl_trait.html) page only contains example where that pattern was used as a return type of a method. It would be cool if there also was at least one example that used it as argument type. Ideally, there'd be again two methods, one showing the pattern and one showing what it de-sugars into.

sammhicks:
closes #1442

### DIFF
--- a/src/trait/impl_trait.md
+++ b/src/trait/impl_trait.md
@@ -1,5 +1,56 @@
 # `impl Trait`
 
+`impl Trait` can be used in two locations:
+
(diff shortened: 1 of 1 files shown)


marioidival:
@sammhicks thank you

PR #1474, opened 2021-11-10
Add note about distinction between attribute and macro cfg from std docs into rust-by-example docs

cspanda:
The docs for `std::cfg` located at https://doc.rust-lang.org/std/macro.cfg.html contain this distinction between the attribute vs the macro.

> `cfg!`, unlike `#[cfg]`, does not remove any code and only evaluates to true or false. For example, all blocks in an if/else expression need to be valid when `cfg!` is used for the condition, regardless of what `cfg!` is evaluating.

I think we can copy this note over the the docs in rust-by-example - it's important enough that I ran into this exact problem but the rust-by-example docs didn't contain the disclaimer. 

Test plan:
- `cargo install mdbook`
- `mdbook serve`
- navigate to http://localhost:3000/attribute/cfg.html
- screenshot below of new docs!

Screen Shot 2021-11-10 at 2 15 28 PM

### DIFF
--- a/src/attribute/cfg.md
+++ b/src/attribute/cfg.md
@@ -9,6 +9,8 @@ While the former enables conditional compilation, the latter conditionally
 evaluates to `true` or `false` literals allowing for checks at run-time. Both
 utilize identical argument syntax.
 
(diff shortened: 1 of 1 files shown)


marioidival:
@cspanda thank you!

PR #1476, opened 2021-11-19
fix into_iter for arrays

adsick:
https://doc.rust-lang.org/stable/rust-by-example/fn/closures/closure_examples/iter_find.html#searching-through-iterators
Present code is out of date due to 2021 edition and gives error on compile. Replacing extra & fixes it.

### DIFF
--- a/book.toml
+++ b/book.toml
@@ -9,3 +9,6 @@ editor = "ace"
 
 [output.html.fold]
 enable = true
+
+[output.html]
+git-repository-url = "https://github.com/rust-lang/rust-by-example"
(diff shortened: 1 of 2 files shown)


adsick:
it fails because the integrated playpen uses old version of Rust. see this for ensuring: 
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ef3c2e1229ea41d49377b2b4df35ca60

marioidival (approved):

marioidival:
@adsick your PR is broken

adsick:
what exactly do you mean?

marioidival:
@adsick could you take a look at github actions log?

adsick:
I've seen the error and the exact reason of it is outdated Rust version. So it is not my PR broken but the checker is using 2018 edition Rust compiler where into_iter for arrays is not implemented the way it is in 2021.

PR #1482, opened 2021-12-05
add missing 'return' in std/result/question_mark.html

schrej:
A previous PR changed this code snippet to `Err(From::from(err))` and dropped the `return`, but according to the [docs](https://doc.rust-lang.org/std/result/index.html#the-question-mark-operator-) the `?` operator _does_ return.


marioidival:
@schrej thank you

PR #1483, opened 2021-12-09
Add a page for inline assembly

Amanieu:
This PR adds inline assembly to Rust by example, which was recently stabilized in rust-lang/rust#72016.

This is based on the "guide-level explanation" from the unstable book which was originally from the RFC.


Amanieu:
Tests are failing because the actual feature in rust-lang/rust hasn't been stabilized yet, but the FCP is done.

PR #1485, opened 2022-01-02
adding missing blank before opening brace in text

ahus1:
Thanks for providing these excellent docs!

This fixes the missing blank on this page:  https://doc.rust-lang.org/rust-by-example/meta.html

image


marioidival:
@ahus1 thank you

PR #1488, opened 2022-01-05
Add the edition requirement and fix into_array for arrays

stupidodie:
Add the edition requirement in the book.toml file and update the code in Iterator::any part for it can compile in the 2021 edition.


stupidodie:
Sorry, seems there are some conflicts that the pr is broken. I will fix it.

PR #1489, opened 2022-01-05
Migrated editions to 2021 and fixed conflicts

stupidodie:


stupidodie:
The check is failed because the file rust-by-example/src/unsafe/asm.md compiled failed. The fail I think is reasonable due to the  [Add a page for inline assembly #1483](https://github.com/rust-lang/rust-by-example/pull/1483) shown that  "Tests are failing because the actual feature in rust-lang/rust hasn't been stabilized yet, but the FCP is done.".

Amanieu:
Inline assembly is fixed by #1490.

marioidival:
@stupidodie and @Amanieu thank you for this 🙏

marioidival:
@stupidodie and @Amanieu thank you so much for you effort, but I'll close this PR because this was fixed in #1499 1499

PR #1500, opened 2022-02-03
documentation for cfg_panic

cchiw:
I'm working on stabilizing the cfg_panic feature [issue](https://github.com/rust-lang/rust/issues/77443).

### DIFF
--- a/src/error/abort_unwind.md
+++ b/src/error/abort_unwind.md
@@ -6,7 +6,6 @@ The previous section illustrates the error handling mechanism `panic`.  The `cfg
 Building on the prior lemonade example, we explicitly use the panic strategy to execise different lines of code.  
 
 ```rust,editable,ignore,mdbook-runnable
-#![feature(cfg_panic)]
 
 fn drink(beverage: &str) {
    // You shouldn't drink too much sugary beverages.
@@ -26,7 +25,6 @@ fn main() {
 Here is another example focusing on rewriting `drink()` and explicitly use the `unwind` keyword.
 
 ```rust,editable,ignore
-#![feature(cfg_panic)]
 
 #[cfg(panic = "unwind")]
 fn ah(){ println!("Spit it out!!!!");}
--- a/src/error/abort_unwind.md
+++ b/src/error/abort_unwind.md
@@ -1,0 +1,53 @@
+# `abort` and `unwind`
+
+The previous section illustrates the error handling mechanism `panic`.  The `cfg_panic` feature makes it possible to execute different code depending on the panic strategy. The current values available are `unwind` and `abort`. 
+
+
+Building on the prior lemonade example, we explicitly use the panic strategy to execise different lines of code.  
+
+```rust,editable,ignore,mdbook-runnable
+#![feature(cfg_panic)]
+
+fn drink(beverage: &str) {
+   // You shouldn't drink too much sugary beverages.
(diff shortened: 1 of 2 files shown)


marioidival:
@cchiw thank you

src/error/abort_unwind.md
```diff
@@ -0,0 +1,51 @@
+# `abort` and `unwind`
+
+The previous section illustrates the error handling mechanism `panic`.  The `cfg_panic` feature makes it possible to execute different code depending on the panic strategy. The current values available are `unwind` and `abort`.
```
ehuss, line 3: I would not use internal code-names for communicating with rust users.  Nothing in the language or other documentation should reference `cfg_panic`.  Instead I would word it something like "Different code paths can be conditionally compiled based on the `panic` setting." or something like that.

src/error/abort_unwind.md
```diff
@@ -0,0 +1,51 @@
+# `abort` and `unwind`
+
+The previous section illustrates the error handling mechanism `panic`.  The `cfg_panic` feature makes it possible to execute different code depending on the panic strategy. The current values available are `unwind` and `abort`. 
+
+
+Building on the prior lemonade example, we explicitly use the panic strategy to execise different lines of code.  
+
+```rust,editable,ignore,mdbook-runnable
```
ehuss, line 8: I'm not sure why this is `ignore`, but it means that the code won't be validated.

src/error/abort_unwind.md
```diff
@@ -0,0 +1,51 @@
...
+
+fn main() {
+    drink("water");
+    drink("lemonade");
+}
+```
+
+Here is another example focusing on rewriting `drink()` and explicitly use the `unwind` keyword.
+
+```rust,editable,ignore
```
ehuss, line 27: Same ignore here.

src/error/abort_unwind.md
```diff
@@ -0,0 +1,51 @@
...
+
+fn main() {
+    drink("water");
+    drink("lemonade");
+}
+```
+
+The panic strategy can be set from the command line by using `abort` or `unwind`.
+
+```rust,editable,ignore
```
ehuss, line 48: This isn't `rust` syntax.  It should be something like `shell` or `console`.  Also, it should not be `editable` or `ignore`.

src/error/abort_unwind.md
```diff
@@ -0,0 +1,51 @@
...
+rustc  lemonade.rc -C panic=abort
```
ehuss, line 49: I'm not sure what the `.rc` extension is. Perhaps this is meant to be `.rs`?

ehuss (commented):
This should not have been merged before the feature was accepted and stabilized on master. This means the documentation will get out-of-sync with what is actually implemented.  I may need to skip updating rustc-by-example to avoid pulling this in.

src/error/abort_unwind.md
cchiw: Thanks for this feedback

src/error/abort_unwind.md
cchiw: Fixed in PR #1505

src/error/abort_unwind.md
cchiw: Fixed in PR #1505

PR #1503, opened 2022-02-09
Add +1 to next_age

FlorentCLMichel:
While the example as it is correctly demonstrates the use of the `?` operator, I feel adding 1 to `next_age` makes it more consistent with the variable name (unless I am missing something?)


marioidival:
Nice catch, @FlorentCLMichel

PR #1506, opened 2022-02-21
Remove duplicate text in asm.md

yusefkarim:


marioidival:
@yusefkarim thank you!

PR #1511, opened 2022-02-27
fix a syntax bug in example assembly

oconnor663:
The `call *` syntax is used in AT&T syntax but not in Intel syntax.
(Source: https://stackoverflow.com/a/19088205/823869.) To make this
example compile, I think we need to remove the `*`.

Playground example broken: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=585dff3c1b43c7a493082ea66c7cdaad
Playground example working: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c4517f46219140ba4345a1656fe7e106

### DIFF
--- a/src/unsafe/asm.md
+++ b/src/unsafe/asm.md
@@ -309,7 +309,7 @@ fn call_foo(arg: i32) -> i32 {
     unsafe {
         let result;
         asm!(
-            "call *{}",
+            "call {}",
             // Function pointer to call
             in(reg) foo,
             // 1st argument in rdi


marioidival (commented):
Cool @oconnor663 ! What you think: we mentioned the two possibilities?

oconnor663:
This PR just removes the `*`, which matches the defaults and is more consistent with the other examples I think?

marioidival:
yes @oconnor663 you right, but I'm asking to you add these examples, with and without *, what you think?

PR #1513, opened 2022-02-28
Update destructure_slice.md

IngridMorstrad:
Fix text. "In like manner to" -> "Like"


marioidival:
@IngridMorstrad thank you!

PR #1518, opened 2022-03-20
Move allow dead code attribute and add comment

khandrym:
Dear @marioidival,

The attribute #![allow(dead_code)] is first time mentioned in section [3.1. Structures]. So my proposition for this section is to move #![allow(dead_code)] attribute to the top of the code and add a comment for it similarly like it is done in other sections (for example in [3.2.1. use] and [3.2.2. C-like]). I think this will help better to understand for the first time what this attribute is for.


marioidival (approved):

marioidival:
@khandrym cool! Thank you for that!

PR #1519, opened 2022-03-21
PathBuf details and example

### LINKED ISSUE #1260: Examples of PathBuf
RBE doesn't mention [PathBuf](https://doc.rust-lang.org/std/path/struct.PathBuf.html) anywhere. A programmer at my work asked:

"Why PathBuf? Just why generally. I started out using it where I thought it made sense, now I just pass &str. Things like .extension() returning an OsStr mean I just use .ends_with() instead which sort of does the same thing (but not always)"

Spoonbender:
fixes #1260 
- added details about `PathBuf` compared to `Path`
- added some usage of `PathBuf` to the example

### DIFF
--- a/src/std_misc/path.md
+++ b/src/std_misc/path.md
@@ -8,6 +8,10 @@ platform-specific `Path` variant.
 A `Path` can be created from an `OsStr`, and provides several methods to get
 information from the file/directory the path points to.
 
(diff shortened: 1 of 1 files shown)


marioidival:
@Spoonbender thank you!

PR #1522, opened 2022-03-27
Update iter_any.md

rafidude:
Line#38 used iter() method instead of into_iter as specified in the comment on line#37.
Modified code from:
println!("2 in array2: {}", array2.iter().any(|&x| x == 2));
to:
println!("2 in array2: {}", array2.into_iter().any(|x| x == 2));

PR #1526, opened 2022-04-03
fix: spacing in iter_any example

fredmaggiowski:
Hi there, while reading the doc (which is great btw!) I noticed that you use to space code between examples to match `.` and function names that are being explained.

I think that this line should be changed in order to be properly aligned with the example below.

Thanks

### DIFF
--- a/src/fn/closures/closure_examples/iter_any.md
+++ b/src/fn/closures/closure_examples/iter_any.md
@@ -33,7 +33,7 @@ fn main() {
     let array2 = [4, 5, 6];
 
     // `iter()` for arrays yields `&i32`.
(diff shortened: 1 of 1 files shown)


gokhanettin:
I think this is more than a simple spacing problem. I believe #1510 is the way to go. It seems we introduced this problem when updating to 2021 edition. `into_iter` on arrays now can iterate by values just like it does on vectors. You may also want to see [this](https://github.com/rust-lang/rust/pull/65819) and [this](https://stackoverflow.com/questions/34733811/what-is-the-difference-between-iter-and-into-iter) for the whole story.

marioidival:
supress by #1522

Issue #1328, opened 2020-04-08
Show warnings in output of live editor


Riezebos:
Thanks for the great book! I was just getting started, trying to learn some Rust.

In `/hello/print/print_display/testcase_list` I was a bit confused about the `?` and was wondering why it is needed. I removed all the `?` after the `write` statements, expecting some error to pop up that would make it more clear why these question marks are added. The output was exactly the same as before.

After years of seeing people say all kinds of (mostly nice) things about about rust, I had a hunch that this was not intended. So I installed rust on my laptop and reran the program. Now i got two very clear warnings:

```
warning: unused `std::result::Result` that must be used
  --> main.rs:12:9
   |
12 |         write!(f, "[");
   |         ^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_must_use)]` on by default
   = note: this `Result` may be an `Err` variant, which should be handled
   = note: this warning originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)
```
My way of trial and error learning might not be the same as others, but it would definitely have helped me here to see this warning in the output. If others feel the same way I would propose to show warnings. (Maybe below the output and slightly lighter or darker)

xiaochuanyu:
This book uses `mdbook` to generate the website and there is a limitation within that tool which ignores warnings when you run Rust code blocks. I've made a PR for `mdbook` that tries to address this issue https://github.com/rust-lang/mdBook/pull/1315.

wooorm:
Bumping because I just ran into the same! (and you closed that issue 😅)

Issue #1296, opened 2019-12-01
"TODO" suggestion doesn't seem to illustrate the point in types/alias


robstarling:
In https://doc.rust-lang.org/rust-by-example/types/alias.html , it says

```
// Use an attribute to silence warning.
#[allow(non_camel_case_types)]
type u64_t = u64;
// TODO ^ Try removing the attribute
```
but removing the attribute does not result in the expected warning about a non-CamelCase type.

What should this page say and/or what would be a better example of this?

ErikSwan:
I believe this is a duplicate of #1328.

robstarling:
Agreed. Thanks!

PR #1528, opened 2022-04-06
Make all new types have UpperCamelCase names in code example in the Aliasing section

khandrym:
Dear Rustaceans,

As was mentioned by @xiaochuanyu in his comment https://github.com/rust-lang/rust-by-example/issues/1328#issuecomment-687883454 currently there is no way to see warnings while running Rust code blocks in `mdbook`.

Thus, my proposition for the code example in the Aliasing section is to use only correct UpperCamelCase names for the new types. These changes do not influence the explanation of how type aliasing works, but can prevent readers from the confusion of not being able to see the warning by removing `#[allow(non_camel_case_types)]` attribute as TODO suggests.

### DIFF
--- a/src/types/alias.md
+++ b/src/types/alias.md
@@ -5,19 +5,15 @@ must have `UpperCamelCase` names, or the compiler will raise a warning. The
 exception to this rule are the primitive types: `usize`, `f32`, etc.
 
 ```rust,editable
-// `NanoSecond` is a new name for `u64`.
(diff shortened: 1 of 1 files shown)


khandrym:
This problem in the Aliasing section was also mentioned by @robstarling in #1296.

marioidival:
cool @khandrym thank you for that!

PR #1534, opened 2022-04-22
i8 u8 typo correction.

liamHowatt:


marioidival:
@liamHowatt this change does not make sense. 
Can you add the unsigned values on page?

marioidival:
fixed by #1521

PR #1542, opened 2022-05-18
Rework cpuid asm example.

### LINKED ISSUE #1540: 22.1 Unsafe Asm: Clobbered registers cpuid example can write to an invalid pointer
In [https://doc.rust-lang.org/nightly/rust-by-example/unsafe/asm.html#clobbered-registers](https://doc.rust-lang.org/nightly/rust-by-example/unsafe/asm.html#clobbered-registers) there is no protection for rbx being used for an input register.

According to the [reference](https://doc.rust-lang.org/nightly/reference/inline-assembly.html#register-names), reserved registers such as rbx may be allocated into the register class "reg".
If the pointer to the cpu maker name gets put into rbx, the following asm is emitted.

```asm
push rbx
cpuid ;rbx is replaced with part of the cpu maker name
mov [rbx], ebx ;rbx is not a valid pointer UNDEFINED BEHAVIOR
;...
```

This bug should be fixed and a warning should be put in to prevent people from making similar mistakes. An example fix would be using explicit input operands and warning against using automatic integer/pointer register assignment when using rbx(64-bit), esi(32-bit) or rbp in asm.

alex34567:
The current cpuid example can write to an invalid register based on register allocation.
Also, the comments and surrounding documentation imply a previous version of the code that did not use an array.

The example was reworked to use rdi to store the array's pointer.
The documentation was updated to reflect the array version of the cpuid code.
The paragraph explaining using reserved registers was split and information saying that general register classes do not work well with reserved registers was added.

Fixes #1540

### DIFF
--- a/src/unsafe/asm.md
+++ b/src/unsafe/asm.md
@@ -232,25 +232,24 @@ fn main() {
     // three entries of four bytes each
     let mut name_buf = [0_u8; 12];
     // String is stored as ascii in ebx, edx, ecx in order
-    // Because ebx is reserved, we get a scratch register and move from
-    // ebx into it in the asm.  The asm needs to preserve the value of
-    // that register though, so it is pushed and popped around the main asm
+    // Because ebx is reserved, the asm needs to preserve the value of it.
+    // So we push and pop it around the main asm.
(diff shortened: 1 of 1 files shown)


marioidival:
@alex34567 thank you

PR #1552, opened 2022-06-15
Example improvements

Rqnsom:
Few more small updates


src/std/arc.md
```diff
@@ -21,6 +26,8 @@ fn main() {
             println!("{:?}", apple);
```
Rqnsom, line 26: Sometimes 5 apples are printed, sometimes 3 apples... 
Let's rather have deterministic behavior with all 10 apples getting printed on the screen.

Joining threads could be used, but that concept is later introduced in `20.1. Threads`, so I prefer simple sleep here.

PR #1571, opened 2022-07-08
Fix inconsistency between comment and code in hello/print.md

gkaply532:


gkaply532:
I was reading and then saw an inconsistency, the comment says 5 spaces but code had 6 instead

gkaply532:
@marioidival whoops I don't know what I was thinking at the time but I am completely wrong about this. this must be reverted.

gkaply532:
huh the more I read now the more I became confused.

gkaply532:
ok, so the `>` modifier allows me to specify the total width after padding (which makes sense) so this patch is wrong, BUT the other example in this same file at the time was also wrong. (I think...)

Issue #183, opened 2014-07-18
A lack of semicolon - is it a bug?


justanotheranonymoususer:
https://github.com/japaric/rust-by-example/blob/master/examples/tuples/tuples.rs#L33-L34

Two of the last `println`s aren't followed by a semicolon. Is it intended?

I'm just learning rust, but why does it even compile? In fact, removing a semicolon from every `println` call works. Is it because `println` is a macro?

PR #1631, opened 2022-10-29
diverging: fix issues with example code

ericswpark:
- Fixes code not compiling due to missing semicolon
- Fixes warning about the unused variable `a`

### DIFF
--- a/src/fn/diverging.md
+++ b/src/fn/diverging.md
@@ -21,8 +21,8 @@ fn some_fn() {
 }
 
 fn main() {
-    let a: () = some_fn();
-    println!("This function returns and you can see this line.")
+    let _a: () = some_fn();
+    println!("This function returns and you can see this line.");
 }
 ```
 


ericswpark:
Apparently I filed this PR too quick without actually checking if the example code compiles. It does so without the semicolon. I didn't realize `println!` macros didn't require a semicolon in the end. Sorry!

I'll leave this PR open as every other example I've looked at includes a semicolon at the end of all statements, but please let me know if omitting it is accepted in Rust syntax (as I don't know enough to comment on it).

ericswpark:
@marioidival may I ask why this was closed without merging?

marioidival:
@ericswpark your PR is invalid

ericswpark:
@marioidival I see, thank you for the reply. Can you please explain why it is invalid? The macro one I'm not sure about, but the unused variable does emit a warning during compilation and I believe this PR fixes it.

ericswpark:
@marioidival I got curious and decided to look it up. In #183 there is an explanation for why semicolons are not required for `println()` macros, although there is a comment about how it is "usual" for `println()` macros to be ended with a semicolon.

In addition, issue #183 links to PR #191, which seems to be the exact same thing as this PR, except that PR #191 adds two semicolons.

I'd appreciate it if you could reconsider, or at the very least let me know why you think this PR is invalid. Thanks!

jdahlstrom:
@marioidival That's both false and rude. Rustaceans should do better than that.

marioidival:
@ericswpark Hello, I'm sorry about the delay.

I skimmed through your description too quickly and didn't notice the addition of the `_` to avoid the warning and only noticed the addition of the semicolon and that's why I responded as invalid.

You know that putting a semicolon in that situation is optional, right?

Anyway, I'll be happy to accept the warning fix if you want to resubmit it.

marioidival:
https://github.com/rust-lang/rust-by-example/pull/1631#issuecomment-1364709523
Yes.

ericswpark:
@marioidival thank you for reconsidering!

> 

Yes, I noticed right after submission: https://github.com/rust-lang/rust-by-example/pull/1631#issuecomment-1295735466 While I do apologize for not checking if the code would compile without one I decided to leave the PR as is because other examples in RBE do include semicolons after `println()` macros and it's easier for beginners if the convention stays consistent throughout.

marioidival:
@ericswpark yes, nice perspective.. I do agree with you!

PR #1632, opened 2022-10-29
fix erroneous white spaces in types/cast.md

n0toose:
Sorry, I know that the change is super super minor, just happened to notice this as I was trying to relearn the language again and got a *super* tiny *bit* annoyed. I promise I'm not doing it for Hacktoberfest or something like that :D


marioidival:
@n0toose no worries, thank you for that

PR #1634, opened 2022-10-30
Clarify Fn in fn/closures/input_parameters.md

n0toose:
Being explicit may be a bit redundant, but I believe that words like "However" grab the reader's attention, which is kind of important here.


n0toose:
I used a new sentence because I did not want to make the previous sentence extremely long, and I also refrained from using any formatting for additional emphasis.

If necessary, I can explain my intentions and as to why I decided to clarify something that should be pretty obvious (from the perspective of a reader like myself).

n0toose:
I also understood the following table from [huonw's entry](https://huonw.github.io/blog/2015/05/finding-closure-in-rust/) way better, but I did not want to risk plagiarizing his work.

image

marioidival (approved):

Issue #1480, opened 2021-12-01
Add a simpler version of read_lines


samoylovfp:
Hello! First of all thanks for the amazing resource!

The [read_lines](https://doc.rust-lang.org/rust-by-example/std_misc/file/read_lines.html) example features an optimized way of reading lines from a file with a `BufReader`.

The code is efficient but is pretty involved and makes a newcomer stumble, forcing them to read about `Result`s, `BufReader`s, `AsRef` and  generics with a `where` clause.

My new-to-rust friend ran into this when trying to solve challenges from Advent of Code.

Most challenges from AoC require reading a sequence of lines from an input file and I feel it would be good to also have a less efficient example that exposes less new concepts to a newcomer, because this is the first thing you run into when trying to solve an AoC challenge.

Example

```rust
fn read_lines(filename: &str) -> Vec<String> {
    read_to_string(filename) 
        .unwrap()  // panic on possible file-reading errors
        .lines()  // split the string into an iterator of string slices
        .map(String::from)  // make each slice into a string
        .collect()  // gather them together into a vector
}
``` 

or

```rust
fn read_lines(filename: &str) -> Vec<String> {
    let mut result = Vec::new();

    for line in read_to_string(filename).unwrap().lines() {
        result.push(line.to_string())
    }

    result
}
```

I am still not happy about having to mention `&str` to `String` conversion but I feel like even though these examples are longer they apply less cognitive pressure to someone not familiar with rust

samoylovfp:
Case in point https://www.reddit.com/r/rust/comments/r6c1hn/how_to_readin_a_file_linebyline_as_integer/

cafce25:
While a good step in the direction of a simpler version FritteX14's solution doesn't solve all the concerns with this issue, it is still using `BufReader`. I'd also argue it's barely less efficient because it still avoids reading the contents of the file into memory. Further because it explicitly states it's more efficient it [confuses some new users on Stack Overflow](https://stackoverflow.com/questions/75359557/why-is-the-efficient-example-of-bufread-in-the-rust-book-efficient) possible more.

vincentdephily:
[...]

PR #1641, opened 2022-11-10
Simpler version of the read_lines script.

FritteX14:
I added a simpler version of the read_lines script, fixing the issue https://github.com/rust-lang/rust-by-example/issues/1480


marioidival (approved):

marioidival:
@FritteX14 thank you!

PR #1658, opened 2022-12-23
fix: add missing space

tomasmiguez:
This way it's consistent with the previous example.


src/fn/closures/closure_examples/iter_any.md
```diff
@@ -43,7 +43,7 @@ fn main() {
     // `iter()` for arrays yields `&i32`.
     println!("2 in array1: {}", array1.iter()     .any(|&x| x == 2));
     // `into_iter()` for arrays yields `i32`.
-    println!("2 in array2: {}", array2.into_iter().any(|x| x == 2));
+    println!("2 in array2: {}", array2.into_iter().any(| x| x == 2));
```
marioidival, line 46: Actually, the previous example is wrong because its better to read without this whitespace :D Can you fix that?

tomasmiguez: Sure, fixed!

PR #1659, opened 2022-12-30
Fix sample code mismatching its comments

chuanqisun:
I assume the intention was to demonstrate how `get_or_insert_with()` can modify value in place. If that's the case, the code should print `my_fruit`, instead of printing `first_available_fruit` twice. 

However, I need to swap the order of the two prints to avoid borrow checker issue


marioidival (approved):

marioidival:
Thank you @chuanqisun

PR #537, opened 2015-04-08
file: fix some comments

woxtu:
Hi,

I fixed some comments to use `io::Result` instead of `IoResult`.

PR #1660, opened 2023-01-01
types/alias: use io::{Result,Error} instead of Io{Result,Error}

SilverRainZ:
seealso: https://github.com/rust-lang/rust-by-example/pull/537


marioidival:
@SilverRainZ thank you

PR #1675, opened 2023-02-05
remove redudant code

woofyzhao:
Even as the inferior counterpart to ```and_then``` the code can be made more concise.


marioidival (approved):

marioidival:
@woofyzhao thank you for that

PR #1679, opened 2023-02-10
Improve the content for `read_lines`

pringshia:
#1480 raised a valid concern around the accessibility of the content for the `read_lines` page for novices.

#1641 tried to address that issue, but I think in the process it caused a lot more confusion as seen [here](https://github.com/rust-lang/rust-by-example/issues/1480#issuecomment-1418822861) and [here](https://stackoverflow.com/questions/75359557/why-is-the-efficient-example-of-bufread-in-the-rust-book-efficient) by making efficiency claims with dubious basis.

This PR attempts to truly address the concerns of #1480 and also revert the confusion introduced by #1641.

### DIFF
--- a/src/std_misc/file/read_lines.md
+++ b/src/std_misc/file/read_lines.md
@@ -42,9 +42,7 @@ and `String::from` respectively.
 ## A more efficient approach
 
 Here we use the `BufRead` class to read the file. `BufRead` uses an internal
-readahead buffer to performantly reduce the number of times the underlying
-file storage layer must be queried per line read. The intermediary buffer
-is much faster to access.
+buffer to reduce intermediate allocations.
 
 We also update `read_lines` to return an iterator instead of allocating new
 `String` objects in memory for each line.
--- a/src/std_misc/file/read_lines.md
+++ b/src/std_misc/file/read_lines.md
@@ -1,44 +1,50 @@
 # `read_lines`
 
-## Beginner friendly method
-This method is NOT efficient. It's here for beginners
-who can't understand the efficient method yet.
+## A naive approach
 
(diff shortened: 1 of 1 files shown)


src/std_misc/file/read_lines.md
```diff
@@ -1,53 +1,62 @@
...
-of a file.
+Note that in both examples above, we must convert the `&str` reference
+returned from `lines()` to the owned type `String`, using `.to_string()`
+and `String::from` respectively.
 
-`File::open` expects a generic, `AsRef<Path>`.  That's what
-`read_lines()` expects as input.
+## A more efficient approach
+
+Here we use the `BufRead` class to read the file. `BufRead` uses an internal
```
sgeisenh, line 44:
```suggestion
Here we pass ownership of the open `File` to a `BufReader` struct. `BufReader` uses an internal
```

src/std_misc/file/read_lines.md
```diff
@@ -1,53 +1,62 @@
...
+readahead buffer to performantly reduce the number of times the underlying
+file storage layer must be queried per line read. The intermediary buffer
+is much faster to access.
```
sgeisenh, line 47:
```suggestion
buffer to reduce intermediate allocations.
```

sgeisenh (commented):
Looks great!

theredpea:
I prefer the change here in #1679 over #1681

Both pull requests address the confusion about the beginner-friendly being no-less-efficient; both pull requests use `read_to_string` to show a less-efficient/naive approach. 

However, #1679 keeps the beginner friendly method more beginner-friendly, (in my opinion -- and I'm a beginner)

PR #1681, opened 2023-02-12
Replaced "non-efficient" implementation

arthurmilliken:
I was a bit confused when I read this example, where the "non-efficient" and "efficient" implementations were identical, except for error-checking, and using a `&str` for the filename.

When reading the following message at the bottom of the document:
```
This process is more efficient than creating a `String` in memory
especially working with larger files.
```

...it seems to me that the original intention was for the "non-efficient" implementation to read the file contents entirely into memory first, then perform parsing on the in-memory String. This PR provides that implementation.

(However, it might be noted that this "non-efficient" implementation actually performs ~7x faster than the "efficient" one, when dealing with a small file!)

### DIFF
--- a/src/std_misc/file/read_lines.md
+++ b/src/std_misc/file/read_lines.md
@@ -12,9 +12,11 @@ use std::io::Read;
 fn read_lines(filename: String) -> Vec<String> {
     // Open the file in read-only mode.
     let mut file = File::open(filename).unwrap();
+
     // Read file contents into a String buffer.
     let mut buffer = String::new();
     file.read_to_string(&mut buffer).unwrap();
+
     // Parse the buffer line by line, and collect into a `Vec<String>`
     buffer.lines().map(|line| line.to_owned()).collect()
 }
--- a/src/std_misc/file/read_lines.md
+++ b/src/std_misc/file/read_lines.md
@@ -1,31 +1,36 @@
 # `read_lines`
 
 ## Beginner friendly method
+
 This method is NOT efficient. It's here for beginners
 who can't understand the efficient method yet.
 
 ```rust,no_run
 use std::fs::File;
-use std::io::{ self, BufRead, BufReader };
+use std::io::Read;
 
-fn read_lines(filename: String) -> io::Lines<BufReader<File>> {
+fn read_lines(filename: String) -> Vec<String> {
     // Open the file in read-only mode.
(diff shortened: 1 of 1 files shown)


src/std_misc/file/read_lines.md
```diff
@@ -75,4 +84,4 @@ $ rustc read_lines.rs && ./read_lines
 ```
 
 This process is more efficient than creating a `String` in memory
-especially working with larger files.
\ No newline at end of file
+especially working with larger files.
```
uschi2000, line 87: consider

```
This process is more efficient than creating a `String` in memory,
especially when working with larger files.
```

arthurmilliken: right. Also consider this comment at the top of the file:
```
This method is NOT efficient. It's here for beginners
who can't understand the efficient method yet.
```
Shouldn't the implementation given actually be an "inefficient" in-memory approach, given this comment at the top of the file?

As it currently stands, both the "inefficient" and "efficient' approaches use the exact same implementation, calling `io::BufReader::new(file).lines()` to parse the file. Why go through all the trouble of warning the user that the first implementation is "NOT efficient", when it uses the exact same approach as the "efficient" solution?

uschi2000: To clarify, I was merely commenting on the punctuation and the `when` in this sentence.

uschi2000: I'm 100% in sync with the change you propose here, although you'll have to duke it out with https://github.com/rust-lang/rust-by-example/pull/1679

arthurmilliken: Ah, understood, thanks for the clarification!

arthurmilliken: I pushed the proposed update, but looks like #1679 goes through a little more trouble to clean up the prose 😄

theredpea: I appreciate this pull request #1681, thanks @arthurmilliken , but I guess I prefer #1679 because its beginner friendly method(s) are more accessible to my beginner brain! (and more consistent with the beginner-friendly methods proposed in #1480)
#1679 's beginner friendly methods:
 - avoids `&mut` 
 - avoids closure syntax `| line | line.to_owned()` 
 - uses line breaks help readability/fluency
 
 Thanks again for everyone's help, it's great to reinforce my learning Rust :)

PR #1688, opened 2023-03-11
Allow dead code to avoid warning

Bi0T1N:
Without it the compiler generates following warning:
```
warning: variant `Fahrenheit` is never constructed
```

### DIFF
--- a/src/flow_control/match/guard.md
+++ b/src/flow_control/match/guard.md
@@ -3,6 +3,7 @@
 A `match` *guard* can be added to filter the arm.
 
 ```rust,editable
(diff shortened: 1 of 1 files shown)


marioidival (changes requested):
Nice, but can I suggest something? Can you add some example to fill the gap?

Bi0T1N:
As far as I understood the user should change the line `let temperature = Temperature::Celsius(35);` and test different numbers as well as using `let temperature = Temperature::Fahrenheit(35);`. For the latter one the compiler will bark at `Temperature::Celsius`  because it is dead code then.

PR #1690, opened 2023-03-13
Change `runtime` error to `compile time` error

jakthom:
The final comment in section `2.3` indicates out of bound indexing causes a runtime error:

```
// Out of bound indexing causes runtime error.
//println!("{}", xs[5]);
```

But this code:

```
fn main() {
    let xs: [i32; 4] = [1,2,3,4];
    println!("{}", xs[4]);
}
```

actually won't compile:
```
rustc src/main.rs
error: this operation will panic at runtime
 --> src/main.rs:5:20
  |
5 |     println!("{}", xs[4]);
  |                    ^^^^^ index out of bounds: the length is 4 but the index is 4
  |
  = note: `#[deny(unconditional_panic)]` on by default

error: aborting due to previous error
```


jakthom:
Woot 🙌 .

@marioidival it looks like rustbot failed to assign to you. Plz let me know if there's anything more I should do here.

marioidival:
Hello @jakthom , we have a conflict here :D Sorry the delay

PR #1710, opened 2023-05-27
Expand `Option::and_then` example to contrast with `map`

vlazzle:


marioidival (approved):

marioidival:
Thank you @vlazzle

vlazzle:
Thanks for the review @marioidival

PR #1717, opened 2023-06-05
Improved wording in asm.md

PhilVoel:
The old version did not make clear that and why rbx was used in the code instead of ebx (this only happens much later at the end of the explanation after the code block), and the difference between which register is used in which cpu mode wasn't clear.


marioidival:
@PhilVoel thank you!

PR #1720, opened 2023-06-15
Fix example in from_into.md

snpkx:
The code using `From` trait is being misused as an example for `Into` trait. I fixed it properly.


marioidival (approved):

marioidival:
@snpkx thank you

PR #1725, opened 2023-07-05
Remove awkward match in if_let

gepbird:
I tried removing the braces from that code example and it compiled, so I assume this feature was introduced to rust after RBE was written thus the whole example became obsolete.
```rust
let optional = Some(7);
match optional {
  Some(i) => println!("This is a really long string and `{:?}`", i),
  _ => {}
};
```

### DIFF
--- a/src/flow_control/if_let.md
+++ b/src/flow_control/if_let.md
@@ -7,11 +7,7 @@ For some use cases, when matching enums, `match` is awkward. For example:
 let optional = Some(7);
 
 match optional {
-    Some(i) => {
-        println!("This is a really long string and `{:?}`", i);
-        // ^ Needed 2 indentations just so we could destructure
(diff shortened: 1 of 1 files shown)


gepbird:
The exhaustive `_` shouldn't be removed so I'm reverting the previous change and only removing the the block after `Some(i)`.

marioidival:
So, essentially you just remove the comments that could be useful for someone. I'll close this PR because do anything, ok @gepbird ?

gepbird:
Sorry, I didn't include the comment in the PR description, please check the changes.
I removed the brackets around `println!`. One comment said those brackets are necessary when they are not.

marioidival (changes requested):
this snippet looks awkward because he's has a comment. I think should be fine!

PR #1726, opened 2023-07-07
Hint for RGB color calculation

kberov:
The reader should concentrate on learning Rust - not waste time on figuring out how to calculate the color which in turn has to display as a hexadecimal value.


marioidival (approved):

marioidival:
Thank you @kberov

PR #1732, opened 2023-07-28
Pedantic `'static` lifetime corrections

k12ish:
I just thought the `Box::leak` was example was too good to keep to myself - i think it further emphazises how memory leaks are safe rust.

Thank you for maintaining this project!

### DIFF
--- a/src/scope/lifetime/static_lifetime.md
+++ b/src/scope/lifetime/static_lifetime.md
@@ -69,7 +69,7 @@ demonstrate, the below example uses
 to dynamically create `'static` references. In that case it definitely doesn't
 live for the entire duration, but only for the leaking point onward.
 
-```rust,editable,norun
+```rust,editable,compile_fail
 extern crate rand;
 use rand::Fill;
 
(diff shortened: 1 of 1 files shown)


src/scope/lifetime/static_lifetime.md
```diff
@@ -69,7 +69,7 @@ demonstrate, the below example uses
 to dynamically create `'static` references. In that case it definitely doesn't
 live for the entire duration, but only for the leaking point onward.
 
-```rust,editable,norun
+```rust,editable,compile_fail
 extern crate rand;
 use rand::Fill;
```
k12ish, line 74: I'd like to note that the rust playground allows use of the top 100 most downloaded crates, allowing `rand` to be used here

src/scope/lifetime/static_lifetime.md
```diff
@@ -62,6 +62,31 @@ fn main() {
 }
 ```
 
+Since `'static` references only need to be valid for the _remainder_ of
+a program's life, they can created while the program is executed. Just to
```
kberov, line 66: they can created
they can be created

k12ish:
Good spot @kberov

marioidival (approved):

marioidival:
thank you @k12ish and @kberov

PR #1748, opened 2023-09-26
Fix and extend the explanation of outer vs inner attributes.

vitvakatu:
While reading the chapter on attributes, I noticed that it contains a misleading phrasing:

> When attributes apply to a whole crate, their syntax is #![crate_attribute], and when they apply to a module or item, the syntax is #[item_attribute] (notice the missing bang !).

If the attribute looks like `#![crate_attribute]`, it doesn't mean it is applied to the whole crate. It does mean it will be applied to the enclosing item.

I fixed it and extended the paragraph with examples.

### DIFF
--- a/src/attribute.md
+++ b/src/attribute.md
@@ -14,9 +14,34 @@ can be used to/for:
 * mark functions that will be part of a benchmark
 * [attribute like macros][macros]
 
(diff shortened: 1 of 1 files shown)


marioidival:
@vitvakatu thank you!

Issue #1520, opened 2022-03-24
Feature request: add i18n support


ghost:
The current [rust by example ] only has english document.
But why not add more i18n translate document.

user can select their own language, like : France, De, Es, Jp, Zh, Kr, It

I would like to join and translate document into Zh [Simple Chinese]

ericswpark:
@cnmade all translations for RBE is available here:

https://github.com/rust-lang/rust-by-example#translations-to-other-languages

The section also lists Chinese which might be what you're looking for?

dalance:
Now [mdbook-i18n-helpers](https://github.com/google/mdbook-i18n-helpers) works fine with RBE because https://github.com/google/mdbook-i18n-helpers/issues/66 was fixed.
It is an internationalization plugin for mdbook which is used at [Comprehensive Rust 🦀](https://google.github.io/comprehensive-rust)

I think that translations can be merged into this repository by the plugin and provide language selector.

PR #1760, opened 2023-10-26
Add i18n support

dalance:
This PR adds internationalization support based on https://github.com/google/mdbook-i18n-helpers.
This doesn't contain actual translations.
If this is acceptable, I want to port the Japanese translation from https://github.com/rust-lang-ja/rust-by-example-ja.
How about it?

#1520

### DIFF
--- a/.github/workflows/rbe.yml
+++ b/.github/workflows/rbe.yml
@@ -1,12 +1,19 @@
 name: CI
 on: [push, pull_request]
 
+env:
+  # Update the language picker in index.hbs to link new languages.
+  LANGUAGES:
+
 jobs:
   test:
     name: Run tests
     runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


marioidival:
@dalance I have a question. Do we not need to anything about currently code (Makefile or set up) to run correctly?

dalance:
I think there is no need to modify the existing codes.
mdbook-i18n-helpers will be called by mdbook internally if `MDBOOK_BOOK__LANGUAGE` is specified.

If mdbook-i18n-helpers is not installed, a warning will be shown only.
So the existing users will not be affected by this PR.

I think an instruction for translators and the note about the above warning for the existing users may be required.

marioidival:
Nice, can you add this information around? @dalance

PR #1768, opened 2023-11-09
fix test name in cargo/test.md

aviad-dev:
I think the _name_ of the test being ran is incorrect.

### DIFF
--- a/src/cargo/test.md
+++ b/src/cargo/test.md
@@ -135,7 +135,7 @@ Corro
 ```
 What actually gets put into `ferris.txt` is this:
 ```shell
-$ cargo test test_file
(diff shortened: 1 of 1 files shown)


marioidival (changes requested):
I would like to suggest to you take a look into whole file, because we had more references to change for test_file. Can you do that?

aviad-dev:
@marioidival - thank you for taking the time to review this.
I looked at the whole file (`src/cargo/test.md`) and found no other places where `test_file` is appropriate. I did add 2 more minor fixes.
If I misunderstood, or just missed anything - please let me know 🙏.

marioidival (approved):

PR #1780, opened 2023-12-11
Make example in pipe.md compatible with both Windows and Unix-type system with less code

HosseinAssaran:


marioidival:
Hello @HosseinAssaran thank you for all previous PR... Can you take a look into this conflict please?

PR #1786, opened 2023-12-14
Mention a source that gives a solution for the assignment in testcase_mapreduce.md

HosseinAssaran:


hkBst:
I'm not sure about the policy about linking to extern solutions, but looking at your solution I see missing indentation and comment lines with wildly different lengths, and then a typo like `THEARD_LIMIT`. I have the impression that otherwise you put a lot of effort in and I did not notice any serious issues, but it definitely needs a bit more polishing.

I personally also think that if an example solution is required, then it should not be external.

PR #1793, opened 2024-01-04
Fix translation workflow

dalance:
I found a missing command in translation workflow.


dalance:
Sorry. This change is not required.

PR #1794, opened 2024-01-05
Add Japanese translation

dalance:
This PR adds

* Japanese translation resource
  + This is based on https://github.com/rust-lang-ja/rust-by-example-ja/commit/baf7bf07780af48842e27e55ff9218ba43ec3512.
  + The original translators are below.

* Additional translation guide for Japanese
  + To avoid review about Japanese at this repo, rust-lang-ja members will review it at our organization's repo.
  + After the review, we open PR to this repo, so you can merge without Japanese review.
  + This guide describes the step for the review.

* Update mdbook-i18n-helpers to v0.3.1
  + This repo is affected by the issue: https://github.com/google/mdbook-i18n-helpers/pull/129.

The original translators list:

@joemphilips @kdnakt @tatsuya6502 @dalance @a2not @h1g0 @swnakamura @etoal83 @TAKOYAKING8 @sei0o @todays-mitsui @furon-kuina @KANIYONIKA @baseballyama @eldesh @kotet @ohno418 @hnakai0909 @4geru @botamochi0x12 @sadnessOjisan @kakubin @takahiro-itazuri @isudzumi @naoty @kishiguro @KisaragiEffective @kazu728 @tomoyuki-nakabayashi @yoichi @H1Gdev @akiradeveloper @akiomik @uenojp @KeenS @stny @shiomiyan @r6eve @tos-kamiya

### DIFF
--- a/.github/workflows/rbe.yml
+++ b/.github/workflows/rbe.yml
@@ -3,7 +3,7 @@ on: [push, pull_request]
 
 env:
   # Update the language picker in index.hbs to link new languages.
-  LANGUAGES:
+  LANGUAGES: ja
 
 jobs:
   test:
@@ -32,7 +32,7 @@ jobs:
 
     - name: Install mdbook-i18n-helpers
       run: |
-        cargo install mdbook-i18n-helpers --locked --version 0.3.0
+        cargo install mdbook-i18n-helpers --locked --version 0.3.1
 
     - name: Report versions
       run: |
--- a/README.md
+++ b/README.md
@@ -58,6 +58,12 @@ Please see the [TRANSLATING.md] file for more details.
 
 [TRANSLATING.md]: https://github.com/rust-lang/rust-by-example/blob/master/TRANSLATING.md
 
+### Translating guide for each languages
+
+* Japanese/日本語: [TRANSLATING_JA.md]
+
+[TRANSLATING_JA.md]: https://github.com/rust-lang/rust-by-example/blob/master/TRANSLATING_JA.md
+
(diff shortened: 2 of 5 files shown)


.github/workflows/rbe.yml
```diff
@@ -3,7 +3,7 @@ on: [push, pull_request]
 
 env:
   # Update the language picker in index.hbs to link new languages.
-  LANGUAGES:
+  LANGUAGES: ja
```
marioidival, line 6: Hello @dalance I have doubt about this line... That change could break the build, couldn't?

dalance: The build artifact seems to work fine.
What point is your consideration?

marioidival: My point is about the other languages. Unfortunately, I'm not fully aware of how the build process for other languages works and it seemed "strange" to add only the Japanese language to the variable, thus being able to influence it and there is also this comment that says that the change could only happen in the index.hbs file

dalance: `LANGUAGES` represents which language should be built.
So the current value is `ja` because there is Japanese resource only, and it will be like `LANGUAGES: ja de fr po` when other language resources are added.

The comment means "When a new language is added to `LANGUAGES`, `index.hbs` should be updated too".
(It may be better that the comment is changed to the above sentence)
In `index.hbs`, the items of language selector are defined.
So If a new language is added, it should be added to the language selector.

KisaragiEffective:
🚀

PR #1798, opened 2024-01-08
Update primitives.md

daniel-web-developer:
Since tuples can hold values of different types, I think it would be important to make it obvious they can do so.

Right now, the book is saying "Tuples like `(1, true)`.". This might make the impression that tuples are limited to holding values of the same type, so that's why I'm proposing to change `1` to a `2`, making it "Tuples like `(2, true)`.".

### DIFF
--- a/src/primitives.md
+++ b/src/primitives.md
@@ -18,7 +18,7 @@ type because it does not contain multiple values.
 ### Compound Types
 
 * Arrays like `[1, 2, 3]`
-* Tuples like `(1, true)`
+* Tuples like `(2, true)`
 
 Variables can always be *type annotated*. Numbers may additionally be annotated
 via a *suffix* or *by default*. Integers default to `i32` and floats to `f64`.


AndreasHuber:
`1` and `true` are of different types,  `i32 `and `bool` respectively, but `2` is an `i32` just like `1`. Thus, the proposal presents no meaningful change to the example and I would suggest closing this PR.

marioidival:
I think the best decision here is closing the PR because in the same page we are introducing to the users to the `scalar types` in Rust:

```
# Primitives
Rust provides access to a wide variety of `primitives`. A sample includes:
### Scalar Types
* Signed integers: `i8`, `i16`, `i32`, `i64`, `i128` and `isize` (pointer size)
* Unsigned integers: `u8`, `u16`, `u32`, `u64`, `u128` and `usize` (pointer
  size)
* Floating point: `f32`, `f64`
* `char` Unicode scalar values like `'a'`, `'α'` and `'∞'` (4 bytes each)
* `bool` either `true` or `false`
* The unit type `()`, whose only possible value is an empty tuple: `()`
```

Issue #1457, opened 2021-08-27
Documentation example contains non-standard `Arguments` section


Skepfyr:
The documentation example is `src/meta/doc.md` on `Person::new` looks like:
```rust
    /// Returns a person with the name given them
    ///
    /// # Arguments
    ///
    /// * `name` - A string slice that holds the name of the person
    ///
    /// # Examples
    ///
    /// ```
    /// // You can have rust code between fences inside the comments
    /// // If you pass --test to `rustdoc`, it will even test it for you!
    /// use doc::Person;
    /// let person = Person::new("name");
    /// ```
```
As far as I am aware the `Arguments` section is non-standard (and recommended against?), I've definitely never seen it in the standard library or any of the most popular crates. I think it could just be removed, maybe even replaced by an example with the headings suggested by [RFC 1574](https://rust-lang.github.io/rfcs/1574-more-api-documentation-conventions.html#using-markdown-1): Examples, Panics, Errors, Safety, Aborts, Undefined Behavior. (Maybe not all at once!)

marioidival:
Hi, I'm closing this issue just to clean up the items that already exist. If you think this issue makes sense to stay open, please create a new issue with updated information and mention this one.

thanks!

PR #1804, opened 2024-01-21
Update the rustdoc example to match best practice

Skepfyr:
#1457 was closed with prejudice ;) so I've made a PR instead!

This change updates the meta/doc example to match the best practices outlined in RFC 1574 (already referenced by the doc). Specifically, it removes the "Arguments" section from the `Person` doc example.

I've also added a mention of `cargo doc --open` as it's pretty useful and people may not think to see if it exists.

I'm also suspicious about all the mentions of invoking `rustdoc` directly as I suspect ~everyone uses `cargo doc` but I haven't touched them in this change.

### DIFF
--- a/src/meta/doc.md
+++ b/src/meta/doc.md
@@ -1,6 +1,7 @@
 # Documentation
 
-Use `cargo doc` to build documentation in `target/doc`.
+Use `cargo doc` to build documentation in `target/doc`, `cargo doc --open`
+will automatically open it in your web browser.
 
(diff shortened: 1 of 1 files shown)


Skepfyr:
For some flavour as to why I want this change - I attempt to advise someone that Rust docs generally avoid enumerating all the arguments like javadoc and they pointed me to this example.

marioidival:
@Skepfyr thanks!

PR #1805, opened 2024-01-22
Updated first sentence of "parsing a string" in string.md

eje211:
Currently, the version of this sentence that's online makes no sense at all. The version in the source does literally make sense, but it is very difficult to read. This version is less terse, but I think it's much clearer.


marioidival (approved):

marioidival:
@eje211 thank you!

PR #1812, opened 2024-02-03
fix(vec): Correct typo in Vec macro

RyoshiKayo:
Just a little type I noticed


marioidival:
Hello @RyoshiKayo I appreciate you opening the PR. Your change is invalid because it includes a compiler error.

```
   Compiling playground v0.0.1 (/playground)
error[E0423]: expected value, found builtin type `i32`
 --> src/main.rs:7:23
  |
7 |     let mut xs = vec![i32, 2, 3];
  |                       ^^^ not a value
  |
```

This is the reason why I will be closing this PR

PR #1819, opened 2024-02-19
While-Let Unable to compile code example on page

stephenlb:
Added `fn main()` as rustdoc didn't catch omitted main() function. There may be a simpler change by adding an annotations.


stephenlb:
Hi! Checking in to see if we can merge? Thank you! 😊

stephenlb:
🎉

Issue #1817, opened 2024-02-17
If and how to PR a great deal of changes


QuineDot:
I ended up reviewing most of this book, and have come to the conclusion that the book needs a non-trivial number of changes to bring it more up-to-date and to correct some inaccuracies.

[You can read my notes here.](https://github.com/QuineDot/rbe-notes/blob/main/README.md)

If you (the maintainers) wish, I can see about turning my notes into PRs.  But as you can see, there are a lot of notes.  My questions are
1. Would this be productive -- can you entertain a large number of changes?
2. If so, how best should the PRs be made -- per page, per chapter?
3. Is it okay restructure sections/pages or are you trying to maintain the page names, etc?

Alternatively, feel free to make use of the notes without my further involvement.

marioidival:
Hi @QuineDot thank you! I just read your notes and I agree with everything from there!

About your questions, It's ok with that for all.

simonsan:
Disclaimer: Not a maintainer, but talking from my experience with Rust Patterns book.

> * If so, how best should the PRs be made -- per page, per chapter?

Per chapter changes work the best IMO, like easy to review (without a lot of context switches) pieces of text for one topic.
 
> * Is it okay restructure sections/pages or are you trying to maintain the page names, etc?

If you do that, and rename a page, don't forget to set up a redirect so old links, e.g. in search engines, stay alive. Here is an example from Rust patterns `book.toml`. Page names are determined by their markdown name.

```toml
[output.html.redirect]
# Redirects in the form of "old-path" = "new-path", where the new path
# is relative to the old path.
"functional/lenses.html" = "optics.html"
```

If you open it happens that there will be a lot of PRs, I would be able to also help to review them, in case they become overwhelming @marioidival 🤗

simonsan:
Another important thing, actually. We use [dprint](https://dprint.dev) in Rust patterns. I think it makes a lot of things easier, because the diffs are essentially always easy to read. I think there is also some setting, that formats the code blocks, I need to look into it. This would essentially make it much easier to merge the above changes.

So I opened a PR to let you have a look how it looks for this codebase.

PR #1829, opened 2024-03-18
style: format the markdown and lint against it in ci

simonsan:
To make future diffing easier and adopt a common style over the whole codebase.

This is essentially our `dprint.json` from https://github.com/rust-unofficial/patterns/blob/main/dprint.json

I think it would make it easier to merge these changes from #1817 

TODO:
- Does it affect the translations as line numbers change, or does parse the markdown directly?


simonsan:
@mgeisler Does this reformatting affect the `mdbook-i18n-helper` translations as line numbers change, or does parse the markdown directly and this reformatting would be ok to do?

mgeisler:
> @mgeisler Does this reformatting affect the `mdbook-i18n-helper` translations as line numbers change, or does parse it the markdown directly and this reformatting would be OK to do?

Reformatting is fine! [Version 0.2 of mdbook-i18n-helpers](https://github.com/google/mdbook-i18n-helpers/blob/main/i18n-helpers/CHANGELOG.md) started normalizing the whitespace when extracting messages, so you won't see any change to the translations.

With version 0.3, we started to only extract string literals and comments from Markdown code blocks, so reformatting them will now also be invisible to the translations. You should run the `mdbook-i18n-normalize` tool on your existing translation files if you have files from version 0.1 that you need to migrate.

PR #1830, opened 2024-03-24
Edit grammatical mistake

progalgo:


progalgo:
It would be easier to read.

PR #1842, opened 2024-05-07
Fix an external link to absolute

dalance:
This PR fixes an  external link from relative to absolute.


marioidival:
Did you have a problem with that? @dalance

dalance:
The link of Japanese translation is broken because it is in `ja` sub-directory, and the relative path points an unexpected place.
I can adjust `ja.po` directly, but I think changing to absolute path is better solution.

PR #1845, opened 2024-05-15
Update ja.po based on the latest master

dalance:


dalance:
I'll update ja.po after merging #1847.

PR #1849, opened 2024-05-17
Adjust translation build on CI

dalance:
This PR adjusts translation build on CI:

* Remove `git restore`
The `git restore` causes differences between CI of this repo and the release build of rust-lang/rust.
* Add link check for all translations
* Remove unused envs: `MDBOOK_OUTPUT__HTML__SITE_URL`, `MDBOOK_OUTPUT__HTML__REDIRECTS`
This is because `site_url` and `redirects` is not specified in `book.toml`.


dalance:
The current broken links will be fixed after merging #1848  and #1847.

PR #1888, opened 2024-09-29
Minor improvements

ShapelessCat:
- Fix typos

- Follow markdown linter to improve style:
  * Remove redundant empty lines
  * Insert empty lines for readability
  * remove trailing spaces


ShapelessCat:
The headings in one file don't increase one level each time. Is this on purpose? Do we want to fix this?

ShapelessCat:
I also think section 8.4 name should be `for`, instead of current "`for` and range", because this is one of 8.4's subsection, which can't cover all contents of 8.4. Can I change this?

marioidival (approved):

PR #1899, opened 2024-11-26
Minor Update to vec.md to fix compiler error

torressam333:
Without `mut` before the `collected_iterator` variable the compiler throws an error when the program is ran.


marioidival:
This is intentional to follow up the excersise on RBE

torressam333:
Oh, I see. Thank you.

Issue #1908, opened 2025-01-10
Improve src/trait/drop.md documentation


wowinter13:
Hello, guys!
What do you think about adding a more thorough description with some custom drop logic, rather than just println!("> Dropping {}", self.name);?
There was a related issue: https://github.com/rust-lang/book/issues/4181#user-content-fn-length-a8074d918a8b10a69d9454365210a841

> TLDR: my main idea is to provide something more obvious than just println!("Dropping CustomSmartPointer with data {}!", self.data);. For instance, when we are talking about files, the current book chapter states that "...lets you customize what happens when a value is about to go out of scope.", but it doesn't really show any customization behavior. So I thought it might be helpful for newcomers to see that drop() could be useful when working with fixtures, so you don't need to write a function fn clear_data() and call it manually after every test - you can just extend drop behavior.

wowinter13:
@marioidival 
Hello there, could you please provide your opinion on this?
If you find it valuable for the community, I’ll submit a PR with the related change

marioidival:
@wowinter13 hello! for sure you can work on that!

PR #1912, opened 2025-02-05
Add more examples @ drop.md

wowinter13:
Fixes https://github.com/rust-lang/rust-by-example/issues/1908

Related: https://github.com/rust-lang/book/issues/4181


wowinter13:
@marioidival I decided to add a separate example as I am not sure about the format & structure you/we want to keep

marioidival (approved):

PR #1934, opened 2025-05-30
fix(#1904): update doc tests to use `playground` as the crate name

### LINKED ISSUE #1904: src/flow_control/for.md into_iter() example invalid
The code doesn't compile (not even online):

```
error[E0382]: borrow of moved value: `names`
   --> src/main.rs:11:29
    |
2   |     let names = vec!["Bob", "Frank", "Ferris"];
    |         ----- move occurs because `names` has type `Vec<&str>`, which does not implement the `Copy` trait
3   |
4   |     for name in names.into_iter() {
    |                       ----------- `names` moved due to this method call
...
11  |     println!("names: {:?}", names);
    |                             ^^^^^ value borrowed here after move
    |
note: `into_iter` takes ownership of the receiver `self`, which moves `names`
   --> /playground/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/iter/traits/collect.rs:346:18
    |
346 |     fn into_iter(self) -> Self::IntoIter;
    |                  ^^^^
    = note: this error originates in the macro `$crate::format_args_nl` which comes from the expansion of the macro `println` (in Nightly builds, run with -Z macro-backtrace for more info)
help: you can `clone` the value and consume it, but this might not be your desired behavior
    |
4   |     for name in names.clone().into_iter() {
    |                      ++++++++
```

mautamu:
This updates the Documentation Testing (21.2) chapter to assume a crate name of `playground` instead of `doccomments`. This allows for the code contained within to be copy-pasted directly into play.rust-lang.org to run; anyone who would want to reproduce this locally would want to create a `playground` library.

Fixes #1904.


mautamu:
Captured the wrong issue number originally, this is for 1656, not 1904.

PR #1937, opened 2025-06-11
introduce new @media query to set a higher content width on ultra wide screens

### LINKED ISSUE #1936: Code examples too narrow on ultrawide monitors when using large font
Hello everybody!

I've noticed that, on my ultrawide monitor (3440 x 1440, 21:9), the width of the code examples is consistently too narrow to fit the actual code, resulting in horizontal scrollbars, which are inconvenient. Considering all the free real estate to the left and right of the content (see screenshot), I was wondering if it would maybe be feasible to insert another CSS breakpoint (or something like that, I'm only half-literate in webdesign, don't mock me) to more efficiently make use of the screen real estate of ultra wide monitors.

Image

Christoph-Harms:
Fixes #1936 

Adds a `@media` CSS query that queries for `min-width: 1380px` (value copied from existing previous break point) and `min-aspect-ratio: 2` (a value I looked up on [Wikipedia](https://en.wikipedia.org/wiki/Ultrawide_formats)). If these conditions are met, it sets the global CSS variable `--content-max-width` to `1200px` (from `750px`).

**Note**: Since modifying the mdbook standard template requires copying over the whole file that one wants to make changes to, the diff isn't very helpful in locating the changes I made. My actual change consists of the *addition of lines 239-244 in `theme/css/chrome.css`*. The rest is just the standard CSS file you get with `mdbook init --theme`.

**Note 2**: I am not a web designer and I more or less tailored this to my personal preference. I'd be more than happy to have someone who actually does web design for a living tear this apart and lecture me on established standards and the likes. :blush: 

## Screenshots
For non-ultra-wide aspect ratios, nothing changes:
image

If the conditions for the new break point are met, the main content area is widened:
image

### DIFF
--- a/theme/css/chrome.css
+++ b/theme/css/chrome.css
@@ -236,6 +236,7 @@ html:not(.js) .left-buttons button {
     }
 }
 
+/* wider main content on ultra wide screens */
 @media only screen and (min-width: 1380px) and (min-aspect-ratio: 2) {
     :root {
         --content-max-width: 1200px;
--- a/theme/css/chrome.css
+++ b/theme/css/chrome.css
@@ -1,0 +1,792 @@
+/* CSS for UI elements (a.k.a. chrome) */
+
+html {
+    scrollbar-color: var(--scrollbar) var(--bg);
+}
+
+#searchresults a,
+.content a:link,
+a:visited,
+a>.hljs {
+    color: var(--links);
+}
+
+/*
+    body-container is necessary because mobile browsers don't seem to like
+    overflow-x on the body tag when there is a <meta name="viewport"> tag.
+*/
+#body-container {
+    /*
+        This is used when the sidebar pushes the body content off the side of
+        the screen on small screens. Without it, dragging on mobile Safari
+        will want to reposition the viewport in a weird way.
+    */
+    overflow-x: clip;
+}
+
+/* Menu Bar */
+
+#menu-bar,
+#menu-bar-hover-placeholder {
+    z-index: 101;
+    margin: auto calc(0px - var(--page-padding));
+}
+
+#menu-bar {
+    position: relative;
+    display: flex;
+    flex-wrap: wrap;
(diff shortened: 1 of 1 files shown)


ehuss:
Can we please not duplicate the chrome.css file? This causes problems when updating mdbook, particularly as this file changes. Can the changes to just the one variable instead be included as an `additional-css` file?

In general, I would also prefer to not special case the styling for individual books, since it causes an inconsistency across our documentation. I'm fine with a little experimentation, but please try to keep it to a minimum.

Christoph-Harms:
> Can we please not duplicate the chrome.css file? This causes problems when updating mdbook, particularly as this file changes. Can the changes to just the one variable instead be included as an `additional-css` file?
> 
I'm sorry, I didn't find anything in the [mdbook documentation for themeing](https://rust-lang.github.io/mdBook/format/theme/index.html) about including new files. I also thought this was cumbersome, but I genuinely thought this was how it had to be done. Although, looking at it now, I probably could have guessed it was possible, given the existence of `language-picker.css`... my bad, sorry. I'd be happy to make a new PR with just the new CSS in a new file. 

> In general, I would also prefer to not special case the styling for individual books, since it causes an inconsistency across our documentation. I'm fine with a little experimentation, but please try to keep it to a minimum.

Hmm, in that case, would it maybe make more sense to do all this directly over at `../mdbook`? Because you're right of course, this is just _one_ book, I hadn't thought about it that way.

Anyway, thanks for the feedback! Not going to lie, I was a little taken aback when this got merged without comment, I was prepared for a discussion. :sweat_smile:

ehuss:
Thanks!

We can probably do it directly in mdbook. I haven't looked closely at this, but I understand that on wider screens it doesn't make good use of space. There are certain concerns around readability and accessibility, so we just need to be a little cognizant of that.

Christoph-Harms:
In that case this can just simply be reverted I suppose.

PR #1964, opened 2025-10-05
Update hello.md

codesurf42:
fixed to a correct output message


codesurf42:
sorry, too quickly reading through, not needed

PR #1977, opened 2025-11-22
Update to mdbook 0.5

ehuss:
Changelog: https://github.com/rust-lang/mdBook/blob/master/CHANGELOG.md#mdbook-051

The language picker builder had to move to head.hbs because of the way the font awesome icons are loaded. The `<i>` handling is no longer dynamic, and not available on the client side.


ehuss:
This should be ready to merge now.

PR #1978, opened 2025-11-27
Add examples for default field values

agraven:
Tracking issue: https://github.com/rust-lang/rust/issues/132162


agraven:
@marioidival Why was this closed?

PR #1987, opened 2026-01-19
Update static_lifetime.md

LucaPaterlini:
update the deprecated example of trait bound,


marioidival:
Please create a new one with the correct code, after the merge conflict.

PR #1991, opened 2026-01-20
Update map.md

LucaPaterlini:
use process on potato show that its not due to the change that the response is not edible, I found confusing the previous version


marioidival:
Ill close because the intend of the example is show the hard way and a simple one, your editing do not change anything - cognitively

Issue #2031, opened 2026-09-13
Test Functions Naming Scheme


LukasBauza:
The naming scheme in (The Rust Programming Language)[https://doc.rust-lang.org/book/ch11-01-writing-tests.html] uses the follwoing naming scheme for tests:
```rust
#[cfg(test)]
mod tests {
    #[test]
    fn foo() {
        assert!(true);
    }
}
```

But within (Rust By Example)[https://doc.rust-lang.org/rust-by-example/testing/unit_testing.html] it uses the following naming scheme for tests:
```rust
#[cfg(test)]
mod tests {
    #[test]
    fn test_foo() {
        assert!(true);
    }
}
```

Rust By Example should use the same naming scheme for tests, as The Rust Programming Language website.

PR #2032, opened 2026-09-13
Update Unit Testing Naming Scheme

LukasBauza:
From:
#[cfg(test)]
mod tests {
    #[test]
    fn test_foo() {
        assert!(true);
    }
}

To:
rust
#[cfg(test)]
mod tests {
    #[test]
    fn foo() {
        assert!(true);
    }
}


LukasBauza:
Fixed #2031

